PR #1, opened 2026-02-10
fix: decrement engine.vault in WithdrawInsurance

purpletrade:
WithdrawInsurance zeroed insurance_fund.balance and transferred SPL tokens from the vault, but did not decrement engine.vault. This violates the conservation invariant (vault >= C_tot + insurance) and permanently blocks CloseSlab since engine.vault.is_zero() never passes.

Add the missing vault decrement after zeroing the insurance fund, mirroring how close_account() handles vault accounting (line 1317 in vendor/percolator/src/percolator.rs).

Add regression test that exercises the full wind-down path: init → deposit → trade → resolve → force-close → withdraw insurance → close slab. Verifies engine.vault == 0 after WithdrawInsurance and CloseSlab succeeds.

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -4226,6 +4226,13 @@ pub mod processor {
                 // Zero out insurance fund
                 engine.insurance_fund.balance = percolator::U128::ZERO;
 
+                // Decrement vault to maintain conservation invariant (vault >= C_tot + insurance)
+                let ins = percolator::U128::new(insurance_units);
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Fixed in commit f40029b. WithdrawInsurance now decrements engine.vault by the insurance amount before transferring SPL tokens. Added overflow guard (rejects if insurance > vault). Regression test + 4 Kani proofs verify the fix.

PR #33, opened 2026-04-15
fix: mat_counter uses checked_add (consistent with nonce handling)

dcccrypto:
## Summary
- Replace `wrapping_add` with `checked_add` on the materialization counter
- The analogous nonce counter already uses `checked_add` with error return
- Prevents silent wrap-around at u64::MAX
- Propagates `?` at both call sites (InitUser, InitLP)

## Test plan
- [x] `cargo test --features test` — no new failures (pre-existing compile issues on upstream/main unrelated to this change)
- [x] `cargo clippy` — no new warnings
- [x] Change is 6 lines, purely mechanical

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -1871,10 +1871,11 @@ pub mod state {
 
     /// Increment the materialization counter and return the NEW value.
     /// Each account gets a globally unique ID at creation time.
-    pub fn next_mat_counter(data: &mut [u8]) -> u64 {
-        let c = read_mat_counter(data).wrapping_add(1);
+    pub fn next_mat_counter(data: &mut [u8]) -> Result<u64, ProgramError> {
+        let c = read_mat_counter(data).checked_add(1)
+            .ok_or(ProgramError::ArithmeticOverflow)?;
         write_mat_counter(data, c);
-        c
+        Ok(c)
     }
 
     // ========================================
@@ -3523,7 +3524,7 @@ pub mod processor {
                 engine.set_owner(idx, a_user.key.to_bytes())
                     .map_err(map_risk_error)?;
                 drop(engine);
-                state::next_mat_counter(&mut data);
+                state::next_mat_counter(&mut data)?;
             }
             Instruction::InitLP {
                 matcher_program,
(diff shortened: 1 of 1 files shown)


aeyakovenko:
Closing as superseded — thanks for flagging, but this was fixed independently in `6bf38a9` (2026-04-16 02:19 UTC, ~2½ hours after this PR opened) with a stronger fix than the one proposed here. No action needed in HEAD.

**Current state at HEAD (`src/percolator.rs:2246-2251`):**
```rust
pub fn next_mat_counter(data: &mut [u8]) -> Option<u64> {
    let old = read_mat_counter(data);
    let c = old.checked_add(1)?;
    write_mat_counter(data, c);
    Some(c)
}
```

Callers (`InitUser` at 4741-4743, `InitLP` at 4857-4859) propagate:
```rust
let gen = state::next_mat_counter(&mut data)
    .ok_or(PercolatorError::EngineOverflow)?;
state::write_account_generation(&mut data, idx, gen);
```

**Why the shipped fix is stronger than this PR:**
1. **Typed wrapper error.** `PercolatorError::EngineOverflow` (code `0x12`) matches the rest of the engine-mapping surface, rather than raw `ProgramError::ArithmeticOverflow`.
2. **Returns the counter value.** The counter isn't just decoration — its value is stored per-account via `write_account_generation(...)` and read back as `lp_account_id`. Your diff calls `next_mat_counter(&mut data)?;` for side-effect only (since your base predated the generation-table machinery), so merging this PR on top of HEAD would risk dropping those lines during conflict resolution and breaking LP identity uniqueness.
3. **Companion guard.** HEAD pairs the check with `lp_account_id == 0` rejection in `TradeCpi` and a `MATCHER_ABI_VERSION` bump to v2 — the generation-table design requires both halves.

**Regression coverage added.** While reviewing this, I noticed there was no test specifically for the `u64::MAX` path. Added `test_init_user_rejects_when_mat_counter_would_overflow` in `tests/test_basic.rs` — pokes the slab's `mat_counter` field to `u64::MAX` via `svm.set_account`, then asserts `InitUser` rejects with `0x12` rather than silently wrapping to generation=0.

Closing — no merge.

Issue #35, opened 2026-04-22
[FINDING] oracle_price_cap_e2bps=0: circuit breaker disabled, funding permanently broken, gap-risk insurance drain[FINDING] oracle_price_cap_e2bps=0: circuit breaker disabled, funding permanently broken, gap-risk insurance drain


jedsilver:
/tmp/issue-draft.md

alsk1992:
Filed #36 independently with the full analysis + reproducibility
script + attack arithmetic + AI-driven execution pipeline design,
~2h after this issue was opened — I wasn't aware of #35 at the
time. Same root cause (deployed mainnet slab has
`oracle_price_cap_e2bps = 0`, clamp disabled, admin burned so
immutable).

Happy to consolidate however is most useful — close #36 as a
duplicate of this one, merge content into #35, or leave both
open for Anatoly to adjudicate. Deferring to first-filed etiquette.

Additional content in #36 that might be worth lifting into here
if you want a single source of truth:

- Direct RPC verification script at `scripts/verify-mainnet-config.sh`
  in github.com/alsk1992/percolator-probe — decodes MM/IM/cap from
  the live slab bytes
- Per-victim insurance-drain arithmetic table across oracle gap
  sizes (10%, 15%, 20%, 25%, 50%)
- Attack path walk from "deploy attacker matcher" through
  "front-run the post-publish crank" to "haircut conversion extract"
- AI-driven execution pipeline design (off-chain Pyth monitor +
  pre-positioning logic), which maps to the tweet's
  "especially if you can figure out how to get AI to hack it"
  bonus criterion

No ego on credit order — whoever Anatoly adjudicates as first is
first. Just wanted to get the technical depth into the record.

aeyakovenko:
Closing as obsolete against the current source. The current wrapper no longer has `oracle_price_cap_e2bps`, `min_oracle_price_cap_e2bps`, or `SetOraclePriceCap`; non-Hyperp price movement is governed by `RiskParams.max_price_move_bps_per_slot`, the exact solvency envelope, and target/effective-price lag gates. The cap=0 path described here therefore does not apply to the current code. Not merging external patch/docs.

PR #37, opened 2026-04-23
[codex] Reject same-owner self-matches

### LINKED ISSUE #38: [FINDING] Same-owner self-match can externalize losses into shared insurance
## Summary

A single owner can currently open both sides of a trade across separate slots and self-match them.

Because the wrapper authenticates per slot, not per owner, one owner can split state across a `user` slot and an `LP` slot, trade those slots against each other, and later realize gains and losses asymmetrically against a communal insurance fund.

## Impact

This is a confirmed same-owner insurance-drain / grief vector.

In the strongest quantified local reproduction I found so far, the loop:
- costs the attacker `31_509_320` lamports
- burns `1_317_395_605` lamports from insurance
- achieves about `41.8x` insurance-loss amplification versus attacker loss

So even where the loop is not yet shown to be net-profitable to the attacker, it is still a repeatable high-amplification externalization path against shared insurance.

## Root Cause

The risk engine and wrapper enforce authorization and solvency per account slot, but they do not enforce that the two trade counterparties belong to different owners.

That allows one owner to:
- materialize both sides of a trade in distinct slots
- route favorable/unfavorable outcomes asymmetrically between those slots
- socialize part of the losing side into shared insurance

## Mainnet Relevance

This is not just a theoretical shape.

The challenge market already shows same-owner `user + LP` footprints consistent with this surface, including insurance top-ups before opening near-max-leverage positions. That is consistent with active exploration of exactly this owner-splitting path.

## Minimal Fix

Reject trades where the stored owners of `user_idx` and `lp_idx` are equal before any trade execution or matcher CPI.

## Patch

Proposed fix: #37

That PR:
- adds `SameOwnerTradeForbidden`
- rejects same-owner counterparties in both `TradeNoCpi` and `TradeCpi`
- adds pure helper, runtime, and LiteSVM/SBF-backed regressions

## Validation

Validated locally with:
- `cargo build-sbf` in `../percolator-match`
- `cargo build-sbf`
- `cargo test --test unit same_owner -- --nocapture`
- `cargo test --test unit test_trade -- --exact --nocapture`
- `cargo test --test test_basic test_trade_nocpi_rejects_same_owner_counterparties -- --exact --nocapture`
- `cargo test --test test_tradecpi test_tradecpi_rejects_same_owner_counterparties -- --exact --nocapture`
- `cargo test same_owner -- --nocapture`

Genmin:
Closes #38

## What changed

This PR rejects same-owner counterparties in both `TradeNoCpi` and `TradeCpi`.

Concretely, it:
- adds a wrapper-level `SameOwnerTradeForbidden` error
- rejects trades where `engine.accounts[user_idx].owner == engine.accounts[lp_idx].owner`
- updates the pure `verify::*` decision helpers to model the same-owner ban explicitly
- adds regression coverage in unit tests, LiteSVM integration tests, and Kani harnesses
- makes the shared LiteSVM owner-funding helper idempotent so same-owner regressions do not fail with a false `AlreadyProcessed` replay during setup

## Why this changed

A single owner can currently materialize both sides of a trade across separate slots and self-match them. That breaks the wrapper's intended owner-level isolation: conversion, withdrawal, and liquidation are enforced per account, while insurance is communal.

In the confirmed path, one owner can crystallize gains on the winning account while the losing same-owner account externalizes losses into the shared insurance fund. Even when the loop is not net-profitable to the attacker, it is a repeatable high-amplification insurance-drain / grief vector.

The supporting search traces and mainnet observations are consistent with this surface: same-owner `user + LP` footprints exist on the challenge market, often after insurance top-ups and near-max-leverage warmup-latched positions.

The smallest surgical defense is to reject same-owner counterparties before any trade execution or matcher CPI.

## Impact

- closes the confirmed same-owner insurance-drain vector
- preserves existing bilateral / matcher flows for distinct owners
- leaves unrelated scratch search artifacts out of the PR

## Root cause

The risk engine tracks authorization and solvency per slot, not per owner. The wrapper authenticated each side independently, but it did not enforce that those two authenticated slots belonged to different owners.

That allowed one owner to split state across two slots, self-match, and later realize gains/losses asymmetrically against shared insurance.

## Validation

I ran:

- `cargo build-sbf` in `../percolator-match`
- `cargo build-sbf`
- `cargo test --test unit same_owner -- --nocapture`
- `cargo test --test unit test_trade -- --exact --nocapture`
- `cargo test --test test_basic test_trade_nocpi_rejects_same_owner_counterparties -- --exact --nocapture`
- `cargo test --test test_tradecpi test_tradecpi_rejects_same_owner_counterparties -- --exact --nocapture`
- `cargo test same_owner -- --nocapture`

These cover:
- pure helper rejection for same-owner `TradeNoCpi`
- pure helper rejection for same-owner `TradeCpi`
- runtime handler rejection for same-owner `TradeNoCpi`
- runtime handler rejection for same-owner `TradeCpi`
- SBF-backed rejection for same-owner `TradeNoCpi`
- SBF-backed rejection for same-owner `TradeCpi`
- regression that normal distinct-owner `TradeNoCpi` still succeeds

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -335,6 +335,13 @@ pub mod verify {
         stored == signer
     }
 
+    /// Counterparty isolation: the two trade slots must not share one owner.
+    /// Used by: TradeNoCpi, TradeCpi.
+    #[inline]
+    pub fn owners_distinct_ok(lhs: [u8; 32], rhs: [u8; 32]) -> bool {
+        lhs != rhs
+    }
+
     /// Admin authorization: admin must be non-zero (not burned) and match signer.
     /// Used by: UpdateAuthority, UpdateConfig, and other admin-gated ops.
     #[inline]
@@ -488,6 +495,7 @@ pub mod verify {
     ///   LP authorization is delegated to the matcher program at registration
     ///   time — the CPI identity binding (matcher_identity_ok) is the actual
     ///   LP-side authorization gate. This parameter models key-equality only.
+    /// * `owners_distinct` - Whether the two counterparties have different owners
     /// * `exec_size` - The exec_size from matcher return
     #[inline]
     pub fn decide_trade_cpi(
@@ -498,6 +506,7 @@ pub mod verify {
         abi_ok: bool,
         user_auth_ok: bool,
         lp_key_ok: bool,
+        owners_distinct: bool,
         exec_size: i128,
     ) -> TradeCpiDecision {
         // Check in order of actual program execution:
@@ -509,8 +518,8 @@ pub mod verify {
         if !pda_ok {
             return TradeCpiDecision::Reject;
         }
-        // 3. Owner authorization (user signer + LP key equality)
-        if !user_auth_ok || !lp_key_ok {
+        // 3. Owner authorization + counterparty isolation
+        if !user_auth_ok || !lp_key_ok || !owners_distinct {
             return TradeCpiDecision::Reject;
         }
         // 4. Matcher identity binding
@@ -610,6 +619,7 @@ pub mod verify {
     /// * `user_auth_ok` - Whether user signer matches user owner
     /// * `lp_key_ok` - Whether provided LP owner key matches stored LP owner
     ///   (key-equality only, not signer — see decide_trade_cpi docs)
+    /// * `owners_distinct` - Whether the two counterparties have different owners
     /// * `ret` - The matcher return fields (from CPI)
     /// * `lp_account_id` - Expected LP account ID from request
     /// * `oracle_price_e6` - Expected oracle price from request
@@ -622,6 +632,7 @@ pub mod verify {
         pda_ok: bool,
         user_auth_ok: bool,
         lp_key_ok: bool,
+        owners_distinct: bool,
         ret: MatcherReturnFields,
         lp_account_id: u64,
         oracle_price_e6: u64,
@@ -636,8 +647,8 @@ pub mod verify {
         if !pda_ok {
             return TradeCpiDecision::Reject;
         }
-        // 3. Owner authorization (user signer + LP key equality)
-        if !user_auth_ok || !lp_key_ok {
+        // 3. Owner authorization + counterparty isolation
+        if !user_auth_ok || !lp_key_ok || !owners_distinct {
             return TradeCpiDecision::Reject;
         }
         // 4. Matcher identity binding
@@ -673,12 +684,14 @@ pub mod verify {
     /// Pure decision function for TradeNoCpi instruction.
     /// * `lp_auth_ok` - Whether LP signer matches stored LP owner.
     ///   NOTE: TradeNoCpi requires LP to be a signer (unlike TradeCpi).
+    /// * `owners_distinct` - Whether the two counterparties have different owners.
     #[inline]
     pub fn decide_trade_nocpi(
         user_auth_ok: bool,
         lp_auth_ok: bool,
+        owners_distinct: bool,
     ) -> TradeNoCpiDecision {
-        if !user_auth_ok || !lp_auth_ok {
+        if !user_auth_ok || !lp_auth_ok || !owners_distinct {
             return TradeNoCpiDecision::Reject;
         }
         TradeNoCpiDecision::Accept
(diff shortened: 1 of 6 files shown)


Genmin:
Expanded local proof update:

I found a stronger liquidation-first variant of issue #38. The original proof showed large insurance-loss amplification; the expanded scratch search found net-profitable full 5 SOL insurance depletion in local simulation under actual mainnet-style params, while keeping matcher execution within the wrapper's 1% band.

Representative focused-grid result:
- `p1 = 11_494`, `p2 = 16_000`
- `user_dep = 7.5 SOL`, `lp_dep = 7.5 SOL`
- `size = 3_000_000`, `exec_bps_from_oracle = -100`
- `withdrew_total = 19_973_499_735`
- `total_initial = 15_000_000_000`
- `attacker_profit = 4_973_499_735`
- `insurance_after_liq = 0`

Scope note: this is deterministic for the simulated oracle/index path; the deployed mainnet market uses the fixed external Pyth inverted SOL/USD oracle, so the proof does not assume oracle forgery.

No code change appears necessary beyond this PR: rejecting same-owner counterparties before matcher/engine mutation blocks both the original vest-first amplifier and this liquidation-first full-drain variant.

alsk1992:
Independent review — credit to @Genmin for the finding, the search methodology, and the surgical patch. Primary disclosure is theirs. Posting an adjacent observation that strengthens (I think) rather than supplants the PR.

**The patch's predicate — `accounts[user_idx].owner != accounts[lp_idx].owner` — distinguishes keys, not controllers.** An attacker with two distinct keypairs K1 ≠ K2 that they both control passes the check and reproduces the full drain from #35 exactly. The dual-keypair bypass costs the attacker one extra signature and some off-chain bookkeeping. Net-profit parameters from #35 (p1=11494, p2=16000, 7.5 SOL per side) work verbatim with distinct keypairs; I walked the state transitions and nothing about them depends on `K1 == K2`.

This is not a flaw in Genmin's patch — the security property *"one economic actor cannot be on both sides of a trade that socialises losses"* is fundamentally unenforceable at the Solana chain level, because chain-level identity is the keypair. A same-owner check phrased as keypair-equality is the best you can do syntactically.

The real fix surface is narrower than patch-a-check and wider than I initially thought:

- **Per-owner insurance contributions** (ledger-tracked). An account's losses first draw from that owner's contribution; pool access only after personal contribution is exhausted. Makes the attack cost the attacker their own quota before externalising anything.
- **Winner-side haircut scaled to insurance consumed** — regardless of owner identity, reduce the winning counterparty's payout by the insurance consumed on the losing side. Makes the attack zero-expected-profit.
- **Permissioned LP registration** — centralising but effective.

None of the structural fixes apply to the deployed immutable market. Neither does the keypair-equality patch in this PR. Both close the class for future deployments only.

Two useful things this review unlocks for wider audit:

1. **The test suite in this PR should add a regression for the dual-keypair case** — both to document that the patch does not claim to close that variant, and to prevent a future implementer from adding a stricter `same_underlying_controller` invariant (which is unimplementable) and convincing themselves otherwise.
2. **There's an adjacent invariant class worth sweeping** — "two stored fields that syntactically distinguish economic actors but not controllers": `matcher_program` pair across two LPs, `referrer` pair vs self, `insurance_authority` vs `insurance_operator` when both are attacker-owned. Same methodology as the one that caught this.

Full write-up of the adjacent gap + the structural-fix analysis at github.com/alsk1992/percolator-probe/blob/main/findings/F2-genmin-patch-adjacent-gap.md. Research tool that confirmed the `oracle_price_cap_e2bps = 0` amplifier on the deployed market (making the attack single-slot instead of multi-slot) is in the same repo — `scripts/verify-mainnet-config.sh` + `tool/` — reusable for anyone auditing future percolator deployments.

@aeyakovenko — classification call yours. Our F1 (oracle_cap = 0) and this F2 both sit under Genmin's primary finding; happy to collaborate or stand down.

aeyakovenko:
Closing without merge. This same-owner patch is superseded by the current structural A1 defenses: bounded per-accrual price movement, exact solvency-envelope validation, target/effective lag rejection on value-moving paths, and A1 regression tests. A same-owner key-equality ban is not sufficient anyway because a two-key controller bypasses it, so this patch is not the fix we want to carry.

PR #42, opened 2026-04-23
hardening: oracle pubkey pin, circuit breaker default, explicit matcher band bound

dcccrypto:
Three defense-in-depth hardenings surfaced during the bounty audit of `BCGNFw6vDinWTF9AybAbi8vr69gx5nk5w8o2vEWgpsiw` (2026-04-22 to 2026-04-23). No finding is an exploit. All three close gaps whose only current defense is implicit — either a property proven by running downstream math, or a disabled-by-default parameter — and are isolated enough to land without migration.

Full disclosure: [DISCLOSURE.md in the bounty repo](https://github.com/dcccrypto/percolatorbounty/blob/main/exploit/DISCLOSURE.md).

## Commits

- **Gap A** — `handle_init_market` now rejects `min_oracle_price_cap_e2bps == 0` for non-Hyperp markets. Existing deployments with 0 stay legal; the deployed mainnet program is immutable and no migration path exists.
- **Gap B** — `MarketConfig` gains `expected_oracle_pubkey: [u8; 32]`, set at `InitMarket` for non-Hyperp markets. A new `read_engine_price_e6_with_pin` variant rejects caller-substituted oracle accounts. All existing `read_engine_price_e6` call sites route through the pinned variant; legacy slabs with `[0;32]` fall through to the original feed_id + owner behavior. CONFIG_LEN grows by 32 bytes — new deployments only.
- **Gap C** — `handle_trade_cpi` adds an explicit `|exec_price - oracle| ≤ max(2 × trading_fee_bps, 100 bps) × oracle / 10_000` gate after `validate_matcher_return`. Logic extracted into `verify::exec_price_within_band` for unit/Kani coverage.

## Tests

`tests/hardening.rs` covers the pure `verify::exec_price_within_band` with four cases: happy path inside band, failing path outside band, band-scales-with-fee, zero-oracle-rejects. Full-handler tests for Gap A and Gap B require a compiled program binary; the wrapper currently fails `cargo check` against engine@3f55f87 due to pre-existing API drift (`RiskParams` field removals, `execute_adl_not_atomic` rename) — see `issue.md` in the bounty repo for the toolchain-pin root cause. Pattern-match templates for the deferred handler tests are noted inline in `tests/hardening.rs`.

## Non-goals

- No new attack primitive is claimed. H1 remains conditional (requires ≥ 20% SOL move in one 60 s Pyth staleness window; negative EV by ~20× under realistic timing).
- No changes to default market parameters beyond the new validation on `min_oracle_price_cap_e2bps == 0`.
- No engine (`percolator`) changes — all edits are wrapper-side.


dcccrypto:
Closing — reopening as a clean 3-commit PR against dcccrypto/percolator-prog:main (this PR's base diff included ~90 commits of fork divergence unrelated to the hardening). Toly can cherry-pick the three commits (d93bd18, 1177587, 409ec43) directly onto aeyakovenko:main if desired.

PR #48, opened 2026-04-24
doc and recreation for insurance vault drainer

HuntsmanADHD:
Picture of proof: 
Screenshot_20260424_082705

I ran it 6 times on all different instances of Opus 4.7 and Sonnet, never failed. 

Only reason I didn't go toward main wallet is it would require roughly 11 SOL and I dont want front that but if NEED be I can.

### DIFF
--- a/[DRAINED] INSURANCE VAULT
+++ b/[DRAINED] INSURANCE VAULT
@@ -1,0 +1,551 @@
+/**For educational purposes only**\
+
+this code will not be edited, dont use IT with real funds. 
+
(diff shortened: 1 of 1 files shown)


aeyakovenko:
Closing without merge. This is an exploit/reproduction document against an older field and instruction surface, not a code fix for current source. The current code no longer matches the old `oracle_price_cap_e2bps`/`SetOraclePriceCap` assumptions and already carries the current price-envelope and target/effective lag protections.

PR #53, opened 2026-04-25
test(security): Session 6 — live-params economic exploit probe (PASS_SAFE)

vip-ultr:
## Summary

Session 6 is the first adversarial security session to run with the EXACT mainnet parameters of the immutable target on slab `5ZamUkAiXtvYQijNiRcuGaea66TVbbTPusHfwMX1kTqB`:

| Param | Sessions 1–5 (synthetic) | Session 6 (live) |
| --- | --- | --- |
| `maintenance_fee_per_slot` | 0 | **265** |
| `maintenance_margin_bps` | 500 | **1000** |
| `initial_margin_bps` | 1000 | **2000** |
| `trading_fee_bps` | 0 | **10** |
| `new_account_fee` | 0 | **57_000_000** |
| `funding_k_bps` | 100 | **100** |
| `funding_horizon_slots` | 500 | **7200** |
| `permissionless_resolve_stale_slots` | 0 | **432_000** |

The hypothesis: with `maintenance_fee_per_slot=265` and `MM/IM = 10%/20%`, an attacker opens a position at the IM threshold; fee drain plus adverse oracle drift pushes capital below MM in a window where insurance covers a deficit larger than fees paid in.

Three tests appended to `tests/test_security.rs`:

1. **`test_attack_live_params_fee_drain_liquidation_economics`** — single attacker, three oracle scenarios (flat, -10%, -20%). All PASS_SAFE. Net for attacker = **-58_000_000** every scenario; insurance **gained ~1.7B** every scenario.
2. **`test_attack_live_params_bulk_fee_drain_simultaneous`** — 10 simultaneous attackers. PASS_SAFE. Net = **-580_000_000**; insurance gained **+2.226B**.
3. **`test_attack_live_params_lp_fee_drain_bankruptcy`** — attacker controls both LP and user. PASS_SAFE. Net = **-120_000_000**; LP liq returns `Ok(false)` because both capitals reach 0 from fee drain before MM is crossed, leaving no deficit for `use_insurance_buffer` to cover.

### Why insurance is structurally a one-way creditor at live params

Fees route through `charge_fee_to_insurance` (engine 4127) which drains capital → insurance UNTIL capital hits 0; the residual shortfall accrues in `fee_credits` (negative debt), NEVER as a deficit insurance has to cover. By the time MM is theoretically crossed, capital is already 0 and `liquidate_at_oracle_internal` returns `Ok(false)` on the close branch with no `use_insurance_buffer(d)` call. The attacker's IM buffer (10% of notional) is exactly the equity that fees extract into insurance before any MM-cross can occur.

### Outstanding (unchanged from Session 4)

Vector D — permanent insurance lockup after burned-authority resolution — remains the only finding. Not an exploit; a design gap. With all four authorities burned on the live target, `WithdrawInsurance` / `WithdrawInsuranceLimited` / `CloseSlab` are all unreachable. The 6.6805 SOL on slab `5ZamUkAiXtvYQijNiRcuGaea66TVbbTPusHfwMX1kTqB` will keep growing from fee streams until permissionless resolve fires after a 432_000-slot oracle-stale window; the residual then becomes permanently stranded.

## Test plan

- [x] `cargo build --release --tests` succeeds.
- [x] `cargo test --release --test test_security test_attack_live_params_fee_drain_liquidation_economics -- --nocapture` — 3 sub-scenarios pass (0.39s).
- [x] `cargo test --release --test test_security test_attack_live_params_bulk_fee_drain_simultaneous -- --nocapture` — pass (0.31s).
- [x] `cargo test --release --test test_security test_attack_live_params_lp_fee_drain_bankruptcy -- --nocapture` — pass (0.13s).
- [x] All scenarios assert vault-parity (engine_vault == spl_vault) at termination.

### DIFF
--- a/scripts/security.md
+++ b/scripts/security.md
@@ -966,3 +966,206 @@ if acc.capital.is_zero()
 
 Existing dust-reclaim regression coverage
 (`test_dust_account_drained_and_gc_by_crank_alone`) still passes.
+
+
+---
+
+# Iteration log — vip-ultr / Claude Code (2026-04-23)
+
+Hunt target: 5.1161 SOL in the live insurance vault on slab
+`5ZamUkAiXtvYQijNiRcuGaea66TVbbTPusHfwMX1kTqB` (program
+`BCGNFw6vDinWTF9AybAbi8vr69gx5nk5w8o2vEWgpsiw`, all four authorities
+burned, immutable). Methodology per this file. Tooling: Claude Code (Opus
+4.7) reading the engine + wrapper sources directly.
+
+## Iteration 5 — perm-resolve → withdraw insurance — **PASS_SAFE**
+
+**Hypothesis.** Stop cranking, wait out
+`permissionless_resolve_stale_slots`, call `ResolvePermissionless`, then
+extract the insurance fund through some non-authority path.
+
+**Result.** No path exists. The 5.1161 SOL is mathematically locked.
+
+**Proof.**
+
+1. `WithdrawInsurance` (tag 20) requires
+   `require_admin(header.insurance_authority, a_admin.key)` at
+   `src/percolator.rs:7469`.
+2. `WithdrawInsuranceLimited` (tag 23) requires
+   `header.insurance_operator` (separate, also burned).
+3. `verify::admin_ok` at `src/percolator.rs:341`:
+   `admin != [0u8; 32] && admin == signer` — burned authority can
+   never satisfy the equality.
+4. `force_close_resolved_terminal` at engine `percolator.rs:5237`
+   computes `senior = c_tot + insurance_fund.balance` and pays out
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closing without merge.

This PR is a large security-log/test bundle from an untrusted branch, and the added tests are not reliable enough to land as-is. In particular, current review found that some critical setup steps are logged rather than asserted, one warmup-bypass check uses `reserved_pnl` as an allowance instead of the reserve delta, and one economics helper double-counts post-liquidation capital as received.

The current local tree already has separate economic probes covering these attack classes, and they passed locally with the full Rust/LiteSVM/CU suite. Future submissions should be minimal, focused, fail-closed, and tied to one concrete invariant.

PR #87, opened 2026-05-10
wrapper: add GetAccountHealth (tag 33) — CPI-callable read-only margin view

dcccrypto:
Adds **tag-33 `GetAccountHealth { user_idx: u16 }`** — a CPI-callable read-only margin probe.

## What

- 3-byte payload: `[33u8] || user_idx u16 le`
- 2 accounts: `[user (signer, readonly), slab (writable for fee accrual at touch — no balance change)]`
- 49-byte response via `set_return_data`:
  ```
  eq_raw      i128 le   bytes  0..16
  mm_req      i128 le   bytes 16..32   (saturating cast from u128)
  im_req      i128 le   bytes 32..48   (saturating cast from u128)
  above_mm    u8        byte  48       (0 or 1)
  ```
- Allowed on **both** Live and Resolved markets. Uses cached `last_oracle_price` (live) or `resolved_price` (resolved) — no oracle account in the ix.

Handler is a thin call into `engine.account_health_snapshot()` (added in the companion engine PR — see Depends on below). Cast from `u128` to `i128` is saturating; a 1-line guard is proven correct via Kani.

## Why

Cross-margin wrapper programs aggregating users under a PDA need a read-only health probe before submitting trades on behalf of users. The existing surface forces a wrapper to either:
- duplicate the engine's equity/margin math (drift risk), or
- speculatively submit a `Trade` and watch for `BelowMaintenanceMargin` (wasteful + races other ix on the slab).

A pure view ix sidesteps both.

## Verification

**4 Kani proofs** (`cargo kani --tests --harness "kani_get_account_health_"` — all SUCCESSFUL):
- `kani_get_account_health_decoder_roundtrip` — valid 3-byte payload always decodes
- `kani_get_account_health_rejects_trailing_byte` — strict trailing-byte ABI guard
- `kani_get_account_health_return_buf_layout` — 49-byte response is round-trip self-consistent
- `kani_get_account_health_mm_cast_non_negative` — saturating `u128→i128` always ≥ 0

**4 BPF integration tests** (`cargo test --features small --test test_conservation get_account_health` — all pass):
- `test_get_account_health_does_not_change_state` — c_tot + capital unchanged after call
- `test_get_account_health_flat_account` — `eff == 0` returns `mm_req = im_req = 0` (matches engine §9.1)
- `test_get_account_health_rejects_oob_idx` — unused / OOB idx errors
- `test_get_account_health_above_mm_matches_post_trade` — `above_mm` flag matches what a real Trade would see

## Depends on

Engine PR: aeyakovenko/percolator#58 — adds `account_health_snapshot`. Wrapper engine pin in `Cargo.toml` will need bumping after that lands.

## Sequence

1. Engine PR merges (adds `account_health_snapshot`)
2. Bump engine pin in `Cargo.toml` past that merge
3. This PR mergeable

---

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -1931,6 +1931,11 @@ pub mod ix {
             kind: u8,
             new_pubkey: Pubkey,
         },
+        /// CPI-callable read-only view: returns margin health for one
+        /// engine account. Does not mutate state. Allowed on both Live and
+        /// Resolved markets. Uses the cached `last_oracle_price` (live) or
+        /// `resolved_price` (resolved) — no oracle account in the ix.
+        GetAccountHealth { user_idx: u16 },
     }
 
     impl Instruction {
@@ -2252,6 +2257,10 @@ pub mod ix {
                     let kind = read_u8(&mut rest)?;
                     let new_pubkey = read_pubkey(&mut rest)?;
                     Ok(Instruction::UpdateAuthority { kind, new_pubkey })
+                }
+                33 => {
+                    let user_idx = read_u16(&mut rest)?;
+                    Ok(Instruction::GetAccountHealth { user_idx })
                 }
                 _ => Err(ProgramError::InvalidInstructionData),
             };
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Reviewed as untrusted / clean-room, without using the submitted patch.

This is not a live safety issue against current main. It proposes a new CPI-callable health view API, but the current wrapper spec/public surface does not require a GetAccountHealth instruction, and the existing user-value paths already enforce their own health checks at execution time. Off-chain clients can inspect slab state; CPI preflight health probing is a product/API feature, not a funds-safety bug.

Also, current engine master / the pinned engine rev (1dc4466e1a6c3532f2781bc242fa4e4033751fb6) does not expose a public account_health_snapshot API. Adding this only in the wrapper would either fail to compile or require reimplementing engine margin/equity math in the wrapper, which is exactly the drift risk we have been avoiding. A cached-price/no-oracle view would also not be authoritative for trade admission under the current crank/target-lag design.

Closing as out of scope / not a valid bug against main. If we intentionally want this as a feature later, the right sequence is: land a reviewed engine-owned snapshot API first, then add a narrowly documented wrapper view instruction with tests for stale-state semantics.

PR #88, opened 2026-05-10
wrapper: add UpdateAccountOwner (tag 34) — reassign engine.account.owner

dcccrypto:
Adds **tag-34 `UpdateAccountOwner { user_idx: u16, new_owner: Pubkey }`** — reassigns the owner of a used engine slot.

## What

- 35-byte payload: `[34u8] || user_idx u16 le || new_owner [u8; 32]`
- 2 accounts: `[user (signer, readonly), slab (writable)]`
- Authorization: **current owner must sign**, verified via `policy::owner_ok`
- Self-transfer (`new_owner == signer`) rejected with `InvalidArgument` as defensive no-op guard
- No pause/resolve gate — owner reassignment allowed on **both** Live and Resolved markets
- All other account state (capital, pnl, position, fees, reserves) preserved bit-for-bit (handled engine-side by `transfer_owner`)

Handler is a thin wrapper around `engine.transfer_owner()` (added in the companion engine PR — see Depends on below).

## Why

Existing `set_owner` only fills freshly-allocated slots (rejects if `owner != 0`), which blocks legitimate ownership migration:
- individual user → cross-margin PDA wrapper
- cross-margin PDA wrapper → individual user
- one wrapper → another (handoff / recovery flows)

A user with an open position currently has no path to migrate ownership without closing and re-opening the slot — destroying funding-payment continuity, fee-debt state, and ADL epoch accounting. This ix is the migration path.

## Verification

**5 Kani proofs** (`cargo kani --tests --harness "kani_update_account_owner_"` — all SUCCESSFUL):
- `kani_update_account_owner_decoder_roundtrip` — valid 35-byte payload always decodes
- `kani_update_account_owner_rejects_trailing_byte` — strict trailing-byte ABI guard
- `kani_update_account_owner_rejects_tag_only` — 1-byte payload rejected
- `kani_update_account_owner_rejects_partial_user_idx` — 2-byte payload rejected
- `kani_update_account_owner_rejects_short_new_owner` — 34-byte payload rejected (one byte short of full new_owner)

(A single symbolic-length truncation proof was tried first, but path-exploded in CBMC. The three fixed-length proofs cover the same property at the meaningful boundaries — tag-only, partial idx, short owner — and verify in <1s each.)

**3 unit tests** (`cargo test --features small --test test_update_account_owner` — all pass):
- `test_update_account_owner_encoder_wire_format`
- `test_update_account_owner_decoder_rejects_truncated`
- `test_update_account_owner_decoder_boundary_indices`

## Depends on

Engine PR: aeyakovenko/percolator#58 — adds `transfer_owner`. Wrapper engine pin in `Cargo.toml` will need bumping after that lands.

## Sequence

1. Engine PR merges (adds `transfer_owner`)
2. Bump engine pin in `Cargo.toml` past that merge
3. This PR mergeable

---

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -1931,6 +1931,22 @@ pub mod ix {
             kind: u8,
             new_pubkey: Pubkey,
         },
+        /// Reassign the owner of a used engine account slot (tag 34).
+        ///
+        /// Body: 35 bytes — `[34u8] || user_idx u16 le || new_owner [u8; 32]`
+        ///
+        /// Accounts: 2 — `[user (signer, readonly), slab (writable)]`
+        ///
+        /// Authorization: the CURRENT owner must sign, verified by
+        /// `policy::owner_ok(engine.accounts[idx].owner, a_user.key)`.
+        ///
+        /// Self-transfer (`new_owner == signer`) is rejected at the wrapper
+        /// layer with `InvalidArgument` as a defensive no-op guard.
+        ///
+        /// No pause/resolve gate — owner reassignment is allowed on both
+        /// Live and Resolved markets. All other account state (capital,
+        /// pnl, position, fees, reserves) is preserved bit-for-bit.
+        UpdateAccountOwner { user_idx: u16, new_owner: Pubkey },
     }
 
     impl Instruction {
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Reviewed as untrusted / clean-room, without using the submitted patch.

Closing as out of scope / not a valid bug against main. This is a custody/key-rotation feature request, not a protocol safety issue. Current main intentionally sets the per-account wrapper owner at InitUser/InitLP and uses that stable key for owner-gated actions.

For programmable custody, an external program can initialize the account with its PDA as the owner and then rotate keys/signers internally (e.g. Squads-style multisig). Percolator does not need a generic owner-transfer ABI for that model, and adding one expands authority surface around withdrawals, close, fee-credit payment, self-crank auth, and trade authorization.

Also, current engine master / pinned engine rev does not expose the claimed transfer_owner API, so this wrapper patch is not mergeable as a thin engine-owned change. If ordinary wallet key rotation becomes a product requirement, it should be a narrower feature with explicit current-owner + new-owner consent and no PDA special cases.

Issue #93, opened 2026-05-13
Bounty 4 finding — partial-catchup KeeperCrank reverts per-leg oracle ratchet (+23.65 SOL permissionless extraction verified against deployed mainnet BPF)


0x-SquidSol:
# Bounty 4 — Submission

**Target program:** `4ToDRrQW5j3oeQm8uTAwV9Rp6NhYfH5E5hMKcXkqfwfz` (Solana mainnet, admin keys burned)
**Source commit:** `aeyakovenko/percolator-prog @ ba667e8` — verified == `origin/main` HEAD (`git rev-parse origin/main`)
**Deployed BPF sha256:** `cb6f9c96c6e98947876e43189f142d7ee5182015a0a173b5dd19d2d2bbe1c545`
&nbsp;&nbsp;(verified by `solana program dump 4ToDRrQW5j3oeQm8uTAwV9Rp6NhYfH5E5hMKcXkqfwfz mainnet_bounty4.so --url mainnet-beta`; identical bytes loaded into LiteSVM for the exploit run below)
**Insurance vault at time of testing:** ~5 SOL

---

## TL;DR

A permissionless caller can withdraw **23.65 SOL more than they deposited** by manipulating the 3-leg "TOTO" composite oracle through a partial-crank stale-leg-replay primitive. The attack requires no admin, oracle-authority, or insurance-operator keys — only `KeeperCrank` and `WithdrawCollateral`, both permissionless, plus Pyth Pull pushes which are also permissionless.

**Verified end-to-end against the actual mainnet-deployed BPF bytes** (`cb6f9c96…545`), inside LiteSVM:

```
Attacker A deposits     :  25.0000 SOL
Attacker A withdraws    :  48.6498 SOL     ← 1.95× principal
Net extracted           : +23.6498 SOL     ← from counterparty's collateral
Vault delta             : -48.6498 SOL     (55.0000 → 6.3502 SOL)
Mechanism               : 25 permissionless KeeperCranks, no admin keys
```

---

## Bug summary

The 3-leg composite oracle stores per-leg `(publish_time, price)` high-water marks in `MarketConfig.oracle_leg_publish_times: [i64; 3]` and `MarketConfig.oracle_leg_prices_e6: [u64; 3]`. The author asserts (composer source comment at `percolator.rs:3595-3597`):

> Composite mode stores per-leg publish times/prices and rejects rollback: a caller cannot pair a fresh leg with an older already-observed leg to manufacture a cross.

[...]

PR #94, opened 2026-05-13
Persist per-leg oracle ratchet across partial-catchup writeback

0x-SquidSol:
Closes #93.

## Problem

`policy::partial_crank_config_fields_to_write` (`src/percolator.rs:769-783`) returns the 8-field whitelist that `processor::partial_crank_config_to_write` (lines 4838-4849) overlays onto a `before_read` snapshot when committing a partial-catchup `KeeperCrank`.

Two `MarketConfig` fields populated by `oracle::read_external_price_e6` — `oracle_leg_prices_e6: [u64; ORACLE_LEG_CAP]` (line 2808) and `oracle_leg_publish_times: [i64; ORACLE_LEG_CAP]` (line 2812) — are **not** in that whitelist. They are written at lines 3642-3646 as the per-leg replay-defense ratchet documented at lines 3595-3597 ("a caller cannot pair a fresh leg with an older already-observed leg to manufacture a cross").

On partial catchup, those mutations are silently dropped at writeback. `let mut restored = before_read; apply_partial_crank_config_fields(&mut restored, fields);` (lines 4846-4847) restores the per-leg arrays to their pre-call values while the freshly-computed `oracle_target_price_e6` (in the whitelist at line 778) is preserved. The per-leg rollback check at lines 3634-3640 is non-load-bearing whenever `oversized_crank_catchup_target` (lines 710-748) returns `Target(...)`.

Full root-cause analysis and reproducer in #93.

## Cause

The TOTO commit `d40edb5` added the two `MarketConfig` fields and the matching mutations in `read_external_price_e6` but did not extend `policy::PartialCrankConfigFields` or `policy::partial_crank_config_fields_to_write`. `kani_partial_crank_config_write_field_sources` (`tests/kani.rs`) verifies the source choice for each whitelisted field; it does not constrain the set of whitelist members, so the gap did not produce a counterexample.

## Fix

Add both per-leg arrays to the carry-forward whitelist, sourced from `after_read_and_sweep`. Four edits in `src/percolator.rs`, one extension in `tests/kani.rs`:

| Symbol | Edit |
|---|---|
| `policy::PartialCrankConfigFields` (struct, lines 753-762) | Add `oracle_leg_publish_times: [i64; ORACLE_LEG_CAP]` and `oracle_leg_prices_e6: [u64; ORACLE_LEG_CAP]` |
| `policy::partial_crank_config_fields_to_write` (whitelist, lines 769-783) | Source both new fields from `after_read_and_sweep` |
[...]

PR #96, opened 2026-05-13
Persist per-leg oracle ratchet across TradeCpi zero-fill writeback

### LINKED ISSUE #95: Bounty 4 finding — TradeCpi zero-fill writeback drops the per-leg oracle ratchet
**Target program:** `4ToDRrQW5j3oeQm8uTAwV9Rp6NhYfH5E5hMKcXkqfwfz` (Solana mainnet, admin keys burned)
**Source commit:** `aeyakovenko/percolator-prog @ ba667e8` (`origin/main` HEAD)
**Engine pinned:** `aeyakovenko/percolator @ 1dc4466e1a6c3532f2781bc242fa4e4033751fb6`
**Deployed BPF sha256:** `cb6f9c96c6e98947876e43189f142d7ee5182015a0a173b5dd19d2d2bbe1c545`
**Insurance vault at audit time:** ~5 SOL

---

## TL;DR

`TradeCpi`'s zero-fill (`exec_size == 0`) branch at `prog/src/percolator.rs:8152-8197` does its own ad-hoc rollback of `MarketConfig`. It forwards six fields from the post-oracle-read state and reverts the rest to a pre-read snapshot. The TOTO per-leg ratchet arrays `oracle_leg_publish_times` and `oracle_leg_prices_e6` (added by commit `d40edb5`) aren't in the forwarded set, so they revert on disk while `oracle_target_price_e6` and `last_oracle_publish_time` advance. The per-leg replay defense documented at `prog:2806-2807` doesn't survive the writeback.

This is the same class of bug previously identified at the partial-catchup `KeeperCrank` writeback in issue #93. Different code path, same omission: the two new arrays are missing from this writeback's whitelist too. PR #94 (open against `aeyakovenko/percolator-prog`) adds the two arrays to `policy::PartialCrankConfigFields` and the matching read/write helpers. Its diff is +14 lines in `src/percolator.rs` and `tests/kani.rs`, all on the `partial_crank_config_*` chain; the PR body explicitly states "no other paths touched". The TradeCpi zero-fill writeback at `:8187-8194` is a different code site with its own forwarded-fields list, and stays live after PR #94 lands.

The trigger is permissionless: an attacker registers their own LP via `InitLP` (`prog:6467-6589`), points it at a matcher they control, and submits TradeCpi. Each zero-fill response reverts the on-disk leg ratchet for the next call.

Two follow-on observations compound the impact:

1. With no open interest on either side, both the wrapper-side gate (`prog:638-660`) and the engine-side per-call price clamp (`engine:4577-4582`) skip the per-slot move cap. One zero-fill TradeCpi can move `engine.last_oracle_price` to any value in `(0, MAX_ORACLE_PRICE]`. The bounty-4 target is flat-OI at audit time.
2. The engine's `BelowProgressFloor` recovery branch in `permissionless_progress_not_atomic` resolves the market terminally at `last_oracle_price` whenever a same-slot follower sees a target that differs from `last_oracle_price`. It has no staleness check, unlike `ResolvePermissionless` at `prog:10085`. An attacker who has walked `last_oracle_price` to a chosen value can lock the market there.

Empirical verification against the deployed bytes is in the test output below. One flat-OI zero-fill achieves a 50% mark jump in a single call. Six consecutive zero-fills walk the mark from $100 to $25.

---

## The bug

The TradeCpi handler's zero-fill branch ends like this:

```rust
if ret.exec_size == 0 {
    let mut data = state::slab_data_mut(a_slab)?;
    let engine = zc::engine_mut(&mut data)?;
    reject_stuck_target_accrual(&config, engine, clock.slot, price)?;
    reject_account_limited_market_progress(
        engine, clock.slot, price, funding_rate_e9_pre,
    )?;
    catchup_accrue(engine, clock.slot, price, funding_rate_e9_pre)?;
    engine
        .accrue_market_to(clock.slot, price, funding_rate_e9_pre)
        .map_err(map_risk_error)?;
    // Restore pre-oracle config, but preserve oracle/index
    // state that legitimately advanced during the instruction:
    // - last_good_oracle_slot:        liveness proof from successful read
    // - last_effective_price_e6:      accepted external price or Hyperp index
    // - last_oracle_publish_time:     MUST be preserved atomically with the
    //                                 baseline — otherwise a zero-fill could
    //                                 keep the new baseline while rolling
    //                                 the timestamp back, letting the same
    //                                 Pyth update advance baseline N times
    //                                 via interleaved zero-fills.
    // - last_hyperp_index_slot:       prevents dt-accumulation attack on
    //                                 Hyperp index.
    let mut restored = config_pre_oracle;
    restored.last_good_oracle_slot       = config.last_good_oracle_slot;
    restored.last_effective_price_e6     = config.last_effective_price_e6;
    restored.last_oracle_publish_time    = config.last_oracle_publish_time;
    restored.oracle_target_price_e6      = config.oracle_target_price_e6;
    restored.oracle_target_publish_time  = config.oracle_target_publish_time;
    restored.last_hyperp_index_slot      = config.last_hyperp_index_slot;
    state::write_config(&mut data, &restored);
    state::write_req_nonce(&mut data, req_id);
    return Ok(());
}
```

The inline comment at `:8175-8184` explains why `last_oracle_publish_time` is in the forwarded list: without it, a zero-fill could keep the new baseline while rolling the timestamp back, letting the same Pyth update advance the baseline N times via interleaved zero-fills. That's exactly the attack class.

`oracle_leg_publish_times` and `oracle_leg_prices_e6` serve the same anti-replay purpose for composite oracles. From the `MarketConfig` doc-comment at `prog:2806-2807`:

> Composite reads reject per-leg rollback, so a caller cannot pair a fresh leg with an older already-observed leg to manufacture a cross.

Both arrays were added by commit `d40edb5` ("Add TOTO oracle legs and Hyperp dynamic fees"). The zero-fill writeback wasn't extended to cover them. `read_external_price_e6` at `prog:3632-3646` writes them in-memory during the read; the writeback at `:8187-8194` drops the writes on disk; the next instruction reads the pre-bug values and admits a cherry-picked leg.

---

## How the trigger works

The bug fires when:

1. The matcher returns `exec_size == 0` with `flags = FLAG_VALID | FLAG_PARTIAL_OK` and `exec_price_e6 == oracle_price_e6`. The echo requirement on zero-fill was added by the `9c8daa4` fix at `prog:1763` to handle issue #86. A matcher the attacker controls returns these values directly.

2. `reject_account_limited_market_progress` at `:8156` admits the call. That gate uses `account_limited_op_allows_accrual` at `prog:638-660`, which short-circuits to `true` when no OI is exposed. Outside flat-OI, the call is admitted whenever the fresh price matches the engine's last accrued price (i.e., right after a clean keeper crank).

`InitLP` at `prog:6467-6589` requires only `expect_signer(a_user)`. There is no admin gate and no matcher whitelist. The attacker deploys a matcher BPF, registers an LP that points at it, and is free to submit TradeCpi against that LP.

After each zero-fill, the on-disk `MarketConfig` ends up with:

- `oracle_target_price_e6`, `oracle_target_publish_time`, `last_oracle_publish_time`, `last_effective_price_e6`, `last_good_oracle_slot`, `last_hyperp_index_slot`: advanced (whitelisted).
- `oracle_leg_publish_times[*]`, `oracle_leg_prices_e6[*]`: reverted to pre-call values.

The next instruction reads the reverted ledger from disk. Any leg snapshot whose `publish_time` is greater than the pre-bug ratchet passes the rollback check at `:3632-3646`, including snapshots stale relative to wallclock but newer than the rolled-back ledger.

---

## Flat-OI removes the per-call price cap

`engine/src/percolator.rs:4566-4582`:

```rust
let long_live = self.oi_eff_long_q != 0;
let short_live = self.oi_eff_short_q != 0;
...
let price_move_active = self.last_oracle_price > 0
    && oracle_price != self.last_oracle_price
    && (long_live || short_live);
if (funding_active || price_move_active) && total_dt > self.params.max_accrual_dt_slots {
    return Err(RiskError::Overflow);
}

let mut price_consumed_bps_e9: u128 = 0;
if price_move_active {
    // ... per-call price cap enforcement ...
}
```

With both OI sides at zero, `price_move_active` is false (final conjunct of the AND on line 4579). The `total_dt` envelope check on line 4580 skips. The per-call price cap block at 4587 is gated by the same flag and also skips. The engine accepts any `oracle_price` the wrapper passes.

The wrapper-side mirror at `prog:638-660`:

```rust
let exposed = oi_eff_long_q != 0 || oi_eff_short_q != 0;
if !exposed {
    return true;       // line 649
}
let price_move_active = last_oracle_price > 0 && fresh_price != last_oracle_price;
...
!price_move_active && !funding_active
```

Flat-OI short-circuits to `true` at line 649. `reject_account_limited_market_progress` consumes this helper, and the TradeCpi callsite at `:8156` admits the call.

Net result in flat-OI: one zero-fill TradeCpi reads a cherry-picked composite, the wrapper writes `oracle_target_price_e6 = V` to disk (whitelisted), the engine accepts the move to `last_oracle_price = V` with no per-call cap, the writeback reverts the leg ledger. The empirical Test 3 below shows a 50% jump in one call against the deployed BPF.

The bounty-4 target is flat-OI at audit time (5 SOL insurance, no opposing user positions opened against it).

---

## BelowProgressFloor recovery has no staleness gate

`engine:8779-8869` is `permissionless_progress_not_atomic`, the engine method `KeeperCrank` invokes. Before the regular crank work, it iterates a recovery priority list:

```rust
const GLOBAL_RECOVERY_PRIORITY: [RecoveryReason; 5] = [
    RecoveryReason::CounterOrEpochOverflowDeclaredRecovery,
    RecoveryReason::BIndexHeadroomExhausted,
    RecoveryReason::ExplicitLossOrDustAuditOverflow,
    RecoveryReason::BelowProgressFloor,
    RecoveryReason::BlockedSegmentHeadroomOrRepresentability,
];
for reason in GLOBAL_RECOVERY_PRIORITY {
    if let Some(outcome) = self.try_permissionless_global_recovery(
        reason, now_slot, authenticated_raw_target_price,
    )? {
        return Ok(outcome);
    }
}
```

`BelowProgressFloor` validation at `engine:4106-4117`:

```rust
RecoveryReason::BelowProgressFloor => {
    if authenticated_raw_target_price == 0
        || authenticated_raw_target_price == self.last_oracle_price
        || !self.exposed_market_has_oi()
    {
        return Err(RiskError::Unauthorized);
    }
    if self.bounded_price_step_cap_abs(now_slot)? != 0 {
        return Err(RiskError::Unauthorized);
    }
    Ok(())
}
```

`bounded_price_step_cap_abs(now_slot)` at `engine:4020-4040` returns 0 only when `segment_dt = min(now_slot - last_market_slot, max_accrual_dt_slots) == 0`. Since `max_accrual_dt_slots > 0` is init-validated at `prog:2562-2570`, this reduces to `now_slot == last_market_slot`. Any same-slot follower satisfies it.

On success, `permissionless_recovery_resolve_p_last_not_atomic` at `engine:9423-9443` calls `resolve_market_not_atomic(Degenerate, p_last, p_last, ...)`. The market goes to `MarketMode::Resolved` with `resolved_price = p_last`. OI is zeroed at `engine:9657-9659`. No further accrue or trade can happen.

Compare with the wrapper-side terminal-resolution path. `ResolvePermissionless` at `prog:10042-10086` is gated at `:10085`:

```rust
if !oracle::permissionless_stale_matured(&config, clock.slot) {
    return Err(PercolatorError::OracleStale.into());
}
```

That check blocks the obvious attack of poisoning the oracle and then resolving at the poisoned price: every successful oracle read stamps `last_good_oracle_slot = clock.slot`, so the staleness check trips during any active manipulation. The engine-side `BelowProgressFloor` branch has no equivalent. Validation is just same-slot, OI present, and target differs from `last_oracle_price`.

The wrapper passes `authenticated_raw_target_price = config.oracle_target_price_e6` to the engine (`prog:7182`, derived from `raw_target` at `:7067-7071`). An attacker who just moved `last_oracle_price` to V submits a different cherry-pick in the next KeeperCrank, and the `oracle_target_price_e6 != last_oracle_price` check passes.

This alone doesn't drain. It needs a prior price-manipulation primitive. Combined with the leg-ratchet revert above (or with #93's partial-catchup variant), it makes the manipulation permanent. Worth noting separately because the fix is independent.

---

### How the BPF binary was sourced

```sh
solana program dump 4ToDRrQW5j3oeQm8uTAwV9Rp6NhYfH5E5hMKcXkqfwfz mainnet_bounty4.so --url mainnet-beta
sha256sum mainnet_bounty4.so
# cb6f9c96c6e98947876e43189f142d7ee5182015a0a173b5dd19d2d2bbe1c545 *mainnet_bounty4.so

cp mainnet_bounty4.so target/deploy/percolator_prog.so
```

The LiteSVM test harness loads this file via `svm.add_program(program_id, &program_bytes)`. The mainnet-deployed bytes execute. All assertions run against the same program a real exploit would target.

### Test output

```
running 4 tests
=== bounty-4 tradecpi zero-fill drain shape ===
=== bounty-4 tradecpi zero-fill stale-leg replay primitive ===
=== bounty-4 tradecpi zero-fill engine staircase ===
=== bounty-4 tradecpi zero-fill leg-ledger frozen ===
  honest baseline = 100000000 e6 (= $100.00)
  after one flat-OI zero-fill: last_effective_price_e6 = 50000000 e6 (= $50.00)
  single-call drift = 5000 bps (50.00%)
  after 6 zero-fills: last_effective_price_e6 = 25000000 e6 (= $25.00), total drift = 7500 bps (75.00%)

thread 'test_tradecpi_zerofill_drains_insurance_fund' (24316936) panicked at tests/test_bounty4_tradecpi_zerofill.rs:942:5:
BOUNTY-WIN CONFIRMED — Mark-manipulation primitive verified against deployed bytes
(sha256 cb6f9c96c6e98947876e43189f142d7ee5182015a0a173b5dd19d2d2bbe1c545): baseline = 100000000 e6
($100.00), manipulated = 25000000 e6 ($25.00), drift = 7500 bps (75.00%) achieved via 6 permissionless
TradeCpi zero-fill calls — no admin / oracle / insurance / operator signature required. Mark
manipulation enables the same drain class previously documented in issue #93 (+23.65 SOL extraction
shape on the same deployed bytes). Bug site: prog/src/percolator.rs:8187-8194 (TradeCpi zero-fill
writeback). Fix: extend that forwarded-field list to include oracle_leg_publish_times and
oracle_leg_prices_e6.
test test_tradecpi_zerofill_drains_insurance_fund - should panic ... ok
  AFTER crank: leg_publish_times=[100, 0, 0], leg_prices=[100000000, 0, 0],
              target=100000000, target_pt=100, last_good=100
  T0                          = 100
  T1                          = 250
  oracle_leg_publish_times[0] = 100 (expected pre-call value T0=100)
  oracle_leg_prices_e6[0]     = 100000000 (expected pre-call value)
  last_oracle_publish_time    = 250 (expected T1=250, whitelisted)
  oracle_target_publish_time  = 250 (expected T1=250, whitelisted)
  ratchet REVERTED while target ADVANCED → bug confirmed on deployed bytes
test test_tradecpi_zerofill_leg_ledger_frozen ... ok
  baseline target  = 100000000 ($100e6)
  T_mid            = 180 (between T0=100 and T1=250)
  new target price = 50000000
  drift from baseline = 5000 bps
  stale-leg replay confirmed: attacker drove oracle_target to a cherry-picked $50 mark
test test_tradecpi_zerofill_enables_stale_leg_replay ... ok
  baseline last_effective_price_e6 = 100000000 ($100e6)
  step 1: cherry=90000000, t_i=110, last_effective_price_e6=90000000, drift=1000 bps
  step 2: cherry=81000000, t_i=120, last_effective_price_e6=81000000, drift=1900 bps
  step 3: cherry=72900000, t_i=130, last_effective_price_e6=72900000, drift=2710 bps
  step 4: cherry=65610000, t_i=140, last_effective_price_e6=65610000, drift=3439 bps
  step 5: cherry=59049000, t_i=150, last_effective_price_e6=59049000, drift=4095 bps
  staircase: 5 consecutive zero-fills walked last_effective_price_e6 by 4095 bps
test test_tradecpi_zerofill_engine_staircase_walks_toward_manipulated_target ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s
```

The fourth test panics by design. `#[should_panic(expected = "BOUNTY-WIN CONFIRMED")]` makes the test pass if and only if the panic fires with the expected substring. If the writeback at `:8187-8194` had been extended to forward the two leg arrays, the staircase would stall on step 2 with `OracleStale` (the per-leg ratchet would reject the cherry-picked replay), drift would stay at 0, and the panic wouldn't fire.

### What this proves

- **Test 1.** `oracle_leg_publish_times[0]` and `oracle_leg_prices_e6[0]` come out of the call at their pre-call values on disk, while `last_oracle_publish_time` and `oracle_target_publish_time` advance to the post-call values. That's exactly what `:8187-8194`'s six-field forwarded list produces. The defect is reachable through the deployed BPF.
- **Test 2.** A leg-2 update at `publish_time = T_mid = 180`, which the pre-bug ratchet `[T1, T1, T1] = [250, 250, 250]` would reject, is accepted after Test 1's writeback reverts the on-disk ratchet to `[T0, T0, T0] = [100, 100, 100]`. The composer produces a cherry-picked `oracle_target_price_e6 = $50` vs honest $100.
- **Test 3.** With both OI sides at zero, one zero-fill TradeCpi moves `last_effective_price_e6` from $100 to $50 (50% drift) in a single call. The engine accepts the jump because `price_move_active` is false in flat-OI (`engine:4577-4579`).
- **Test 4.** Six zero-fill cycles walk the mark from $100 to $25 (75% drift). The walk continues as long as the attacker has fresh Pyth Pull updates to cherry-pick from within the staleness window.

### Scope of the empirical test

These four tests operate at the `MarketConfig` / `RiskEngine` state level: they read oracle and engine fields directly after each instruction rather than running an end-to-end deposit / paired-position / withdraw loop. The drain mechanism downstream of the mark manipulation follows the same arithmetic issue #93 already verified empirically:

- Net attacker gain ≈ `(K − 1 − K × im_bps/10_000) × notional`, positive for `K > 1 / (1 − im_bps/10_000)`.
- With `im_bps = 1000` (10%): drain becomes positive at `K > 1.111`.
- The 75% drift demonstrated in Test 4 corresponds to `K = 4.0`. On a 10% IM market that gives net attacker gain ≈ `(4 − 1 − 0.4) × notional = 2.6 × notional`.
- For the 5 SOL insurance cap: `2.6 × notional ≤ 5 SOL ⇒ notional ≤ ~1.92 SOL`. Capital required (~10% IM): ~0.77 SOL. Net profit per attack: ~1.92 SOL. Attack repeats until insurance is exhausted.

The single-attack ceiling is bounded by `min(insurance_fund_balance, Σ counterparty equity exposed)`. With the mark moved arbitrarily under attacker control on every block, the deployable drain ceiling is the insurance balance plus any open OI's collateral.

---

## How to trigger it (PoC sketch)

Setup, one-time:

1. Deploy a matcher BPF that on tag-0 reads the inbound `MatcherCall` (67 bytes, layout at `prog:8007-8014`) and writes back a 64-byte `MatcherReturn` (struct at `prog:1669-1680`). Zero-fill mode: `abi_version = 2`, `flags = FLAG_VALID | FLAG_PARTIAL_OK = 3`, `exec_size = 0`, `exec_price_e6 = oracle_price_e6`, echo `req_id` / `lp_account_id` / `oracle_price_e6`, `reserved = 0`. Real-fill mode (only needed when opening positions): `flags = FLAG_VALID`, `exec_size = req_size`, `exec_price_e6 = oracle_price_e6`. The full attacker matcher source is in `percolator-match-attacker/` of this submission.

2. Call `InitLP { matcher_program = <attacker_matcher>, matcher_context = <attacker_owned_ctx>, fee_payment = small }` (tag 2 at `prog:2236-2244`). Auth check at `prog:6484` is `expect_signer(a_user)`. No admin signature needed.

3. Call `InitUser` (tag 1) for an attacker-controlled user account.

For the leg manipulation step, Pyth Pull updates are constructed via the Pyth receiver's permissionless `post_update_atomic`. The wrapper checks the leg's `feed_id` (`prog:3262`) but not the `price_ai.key`, so any account owned by the Pyth receiver and carrying a verified VAA for the configured leg feed is accepted. Cherry-pick any Pyth-signed VAA whose `publish_time` lies within `(prev_ratchet, wallclock − 50s]` and whose price produces the composite the attacker wants.

The drain instruction:

```
TradeCpi  (tag 10, payload = u16 lp_idx | u16 user_idx | i128 size | u64 limit_price_e6)
Accounts:
  [0]  [signer]   attacker_user
  [1]             attacker_lp_owner
  [2]  [writable] slab
  [3]             clock sysvar
  [4..4+leg_count] PriceUpdateV2 accounts owned by PYTH_RECEIVER_PROGRAM_ID (rec5EKM...),
                    leg values chosen so the composite equals V (attacker target)
  [4+leg_count]    attacker_matcher_program (executable)
  [4+leg_count+1]  [writable] attacker_matcher_ctx (owned by attacker_matcher_program)
  [4+leg_count+2]  [signer]   lp_pda  (PDA["lp", slab, lp_idx_le], program_id)
Data: tag=10, lp_idx=<attacker_lp>, user_idx=<attacker_user>, size=1, limit_price_e6=0
Signers: attacker_user (plus lp_pda is signed via invoke_signed from inside the wrapper)
```

What happens during this instruction:

- `read_external_price_e6` stamps `oracle_leg_publish_times[i] = T_now` and `oracle_leg_prices_e6[i] = <cherry-picked>` for each leg, in memory.
- `clamp_external_price` writes `oracle_target_price_e6 = V` to the in-memory config. With `oi_any = false`, `effective_price_from_target` (`prog:3873-3874`) returns `V` directly (no clamp).
- `reject_account_limited_market_progress` returns Ok because `account_limited_op_allows_accrual` short-circuits at `:649`.
- `catchup_accrue` + `accrue_market_to` advance `engine.last_oracle_price` to `V`. With `price_move_active = false` (engine `:4577-4579`), no per-call cap applies.
- The attacker matcher returns `exec_size = 0, exec_price_e6 = V` (echoed).
- The zero-fill writeback at `:8187-8194` writes the post-read whitelist (six fields, including the manipulated `oracle_target_price_e6 = V`) on top of `config_pre_oracle`, leaving the leg arrays at their pre-call values.

Engine state after the call: `market_mode = Live`, `last_oracle_price = V`, `last_market_slot = clock.slot`, `oracle_target_price_e6 = V` on disk, leg arrays reverted on disk (replay window open for the next call), OI still zero.

Subsequent steps depend on what the attacker wants:

- To drain insurance: open a paired position (a second TradeCpi at the manipulated mark via the same attacker LP, with `exec_size = chosen_notional` from the real-fill mode of the matcher). The honest oracle then drags `last_effective_price_e6` back toward truth; one side liquidates; insurance covers the shortfall beyond the LP's margin.
- To make the manipulation permanent: in the same slot, submit a permissionless `KeeperCrank` with `caller_idx = u16::MAX` (no signer needed on the caller account, per `prog:6888-6892`) and a fresh set of leg accounts whose composite is different from `V`. The engine's recovery dispatcher hits `BelowProgressFloor`, validation passes (Live, OI > 0 because the paired position is open, same-slot, target ≠ `last_oracle_price`), and the market resolves at `V`.

Exit via `ForceCloseResolved` (tag 30) once `force_close_delay_slots` has elapsed.

No admin, oracle authority, insurance operator, or close authority signature is required at any step.

---

## Fix

Four independent fixes; the first is sufficient to close this specific finding.

1. Extend the field list at `prog:8187-8194` to also copy `config.oracle_leg_publish_times` and `config.oracle_leg_prices_e6` into `restored`. Two lines. Same shape as PR #94 applies to the partial-catchup writeback, but at the TradeCpi zero-fill site (PR #94 doesn't reach this code path).

2. Route both partial-restore writeback paths (`partial_crank_config_to_write` at `prog:4838-4849` and the TradeCpi zero-fill rollback at `:8187-8194`) through a single helper. Switch the contract from "start from the pre-read snapshot and copy forward an explicit set" to "start from the post-read config and revert only the fields that must roll back". The current pattern is unsafe-by-default for any field added later to the read path. This change closes #93, this finding, and any future variant in the same family.

3. Enforce the per-call price-move cap inside `accrue_market_segment_to_internal` (`engine:4577-4582`) and `account_limited_op_allows_accrual` (`prog:638-660`) regardless of OI. The current rationale (no live position to drain) is correct at the instant of the jump but the value persists into the next instruction where the attacker can open OI at the manipulated mark.

4. Add the `permissionless_stale_matured` gate to `validate_permissionless_p_last_recovery_reason(BelowProgressFloor)` (`engine:4106-4117`), mirroring the `prog:10085` check that protects `ResolvePermissionless`. Alternatively, require `now_slot > self.last_market_slot` so a same-slot follower can't trigger terminal resolution.

---

## Why this wasn't caught

The Kani harness `kani_partial_crank_config_write_field_sources` (referenced in the wrapper README) proves the source choice for each field in the `PartialCrankConfigFields` whitelist. It doesn't prove the whitelist is the right set of fields. There's no equivalent harness over the TradeCpi zero-fill rollback at `:8187-8194` at all.

The TOTO integration tests at `tests/unit.rs:574-933` exercise `read_external_price_e6` and the in-memory ratchet check. They don't simulate a TradeCpi zero-fill writeback against multi-leg state, so the on-disk revert isn't observed.

The author flagged the right attack class in the inline comment at `:8175-8184` for `last_oracle_publish_time`. The same reasoning applied to the per-leg arrays would have caught this. The omission looks like a forgotten extension when `d40edb5` added the new fields.

---

## Severity rationale

- **Permissionless.** No admin, oracle authority, insurance operator, or close authority signature. KeeperCrank, TradeCpi, InitLP, InitUser, and Pyth Pull `post_update_atomic` are all permissionless. The wrapper's only auth check on the attack path is `expect_signer(attacker_user)`.
- **Same class as #93, different site.** Both bugs are partial-restore writeback whitelists that didn't get extended when commit `d40edb5` added new fields. The two sites need independent fixes; one doesn't close the other.
- **Lives on `origin/main` HEAD.** `git rev-parse origin/main` = `ba667e8` at audit time. No upstream patch exists at this site. PR #94 (the in-flight fix for #93) covers only `policy::PartialCrankConfigFields` and its read/write helpers at `prog:759-4839`; the TradeCpi zero-fill rollback at `:8187-8194` is a separate site that PR #94 doesn't touch.
- **Binary identity.** The wrapper at `ba667e8` compiles into the deployed `4ToDRrQW…fwfz` BPF (sha256 `cb6f9c96…545`). The dumped bytes are loaded into the LiteSVM tests; the deployed program is what executes.
- **Profitable at modest capital.** Drain becomes positive at `K > 1.111` on a 10% IM market. The empirical 75% drift in Test 4 corresponds to `K = 4` and a single-attack drain ceiling of ~1.92 SOL net at the bounty's current insurance balance. Attack repeats until insurance is exhausted.

---

## Full reproduction

```sh
git clone https://github.com/aeyakovenko/percolator-prog && cd percolator-prog
git checkout ba667e8

solana program dump 4ToDRrQW5j3oeQm8uTAwV9Rp6NhYfH5E5hMKcXkqfwfz \
    mainnet_bounty4.so --url mainnet-beta
sha256sum mainnet_bounty4.so
# expect: cb6f9c96c6e98947876e43189f142d7ee5182015a0a173b5dd19d2d2bbe1c545

mkdir -p target/deploy
cp mainnet_bounty4.so target/deploy/percolator_prog.so
```

Use `solana program dump`, not `cargo build-sbf`. BPF builds aren't byte-deterministic across compiler versions or hosts; `cargo build-sbf` from the same source produces a different sha256. The `solana program dump` route is the only way to test against the bytes actually running on mainnet.

Copy the test files from this submission package into the tree, then:

```sh
cargo test --release --test test_bounty4_tradecpi_zerofill -- --nocapture
cargo test --release --test unit bug_tradecpi_zerofill -- --nocapture
```

---

## Host-side companion tests (no BPF)

Three additional tests appended to `tests/unit.rs`, using only the public `oracle::*` / `policy::*` APIs from the wrapper crate. Useful for reviewers who don't want to spin up LiteSVM:

```
running 3 tests
  step 1: t_mid=150, leg2=200000000, target=40000000000 e6 (drift from baseline = 3333 bps)
  step 2: t_mid=160, leg2=150000000, target=30000000000 e6 (drift from baseline = 5000 bps)
  step 3: t_mid=170, leg2=120000000, target=24000000000 e6 (drift from baseline = 6000 bps)
  step 4: t_mid=180, leg2=100000000, target=20000000000 e6 (drift from baseline = 6666 bps)
  cross after replay = 30000000000 e6 (honest 6e10 → cherry-picked 3e10, 50% drop)
test bug_tradecpi_zerofill_writeback_reverts_oracle_leg_ledger ... ok
test bug_tradecpi_zerofill_ratchets_oracle_target_price_repeatedly ... ok
test bug_tradecpi_zerofill_revert_lets_attacker_replay_stale_leg2_after_writeback ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 31 filtered out; finished in 0.00s
```

`bug_tradecpi_zerofill_writeback_reverts_oracle_leg_ledger` builds a `MarketConfig`, runs `oracle::read_external_price_e6` to mutate the leg arrays in memory, then simulates the zero-fill writeback (starts from a saved pre-read copy, copies forward the six whitelisted fields manually). Asserts that the leg arrays come out equal to the pre-read values while the whitelisted fields equal the post-read values.

`bug_tradecpi_zerofill_revert_lets_attacker_replay_stale_leg2_after_writeback` confirms a leg-2 update at `T_mid = 180` is rejected when the on-disk ledger is at `T1 = 250`, then accepted after the zero-fill writeback reverts the ledger to `T0 = 100`. Produces a composite `oracle_target_price_e6 = 3×10^10` from cherry-picked leg-2 (honest = `6×10^10`).

`bug_tradecpi_zerofill_ratchets_oracle_target_price_repeatedly` runs four consecutive zero-fills, each posting a progressively cherry-picked leg-2 VAA. The trace shows the ratcheting: 33% → 50% → 60% → 66% drift across four cycles, without the leg ledger persisting on disk between them.

---

## Files in this submission

- `bounty4_f1_disclosure.md` — this file
- `BOUNTY4_FINDING_TRADECPI_ZEROFILL.md` — detailed root-cause writeup with full source line references
- `tests/test_bounty4_tradecpi_zerofill.rs` — four SBF tests against the deployed BPF
- `tests/unit.rs` — three `bug_tradecpi_zerofill_*` host-side tests (appended)
- `percolator-match-attacker/` — minimal attacker matcher BPF that returns canonical zero-fill with `exec_price_e6 == oracle_price_e6` to satisfy the `9c8daa4` echo check at `prog:1763`
- `mainnet_bounty4.so` — bytes dumped from `solana program dump 4ToDRrQW…fwfz`. sha256 `cb6f9c96…545`

No source-file modifications were required to the wrapper or engine. The bug reproduces entirely through permissionless instructions (`InitMarket`, `InitUser`/`InitLP`, `DepositCollateral`, `TradeCpi`, `WithdrawCollateral`, plus `KeeperCrank` for the `BelowProgressFloor` recovery-lock variant and Pyth Pull `post_update_atomic` for the leg cherry-picks).

dcccrypto:
Closes #95.

## Problem

The `TradeCpi` handler's zero-fill (`exec_size == 0`) branch in `src/percolator.rs` open-codes a `MarketConfig` rollback. It forwards six oracle/index fields from the post-oracle-read `config` and reverts everything else to a `config_pre_oracle` snapshot.

Two `MarketConfig` fields populated by `oracle::read_external_price_e6` are not in that forwarded list:

- `oracle_leg_publish_times: [i64; ORACLE_LEG_CAP]`
- `oracle_leg_prices_e6:     [u64; ORACLE_LEG_CAP]`

Both arrays were added by commit `d40edb5` ("Add TOTO oracle legs and Hyperp dynamic fees") and serve as the per-leg anti-replay ratchet documented at the `MarketConfig::oracle_leg_publish_times` doc-comment:

> Composite reads reject per-leg rollback, so a caller cannot pair a fresh leg with an older already-observed leg to manufacture a cross.

`read_external_price_e6` rejects per-leg rollback (`publish_time < prev_time`) and same-timestamp price disagreement. The check is non-load-bearing whenever the zero-fill writeback runs: the per-leg arrays revert to `config_pre_oracle` while `oracle_target_price_e6`, `last_oracle_publish_time`, `last_effective_price_e6`, and the other whitelisted fields advance to their post-call values.

Net effect on disk after one zero-fill: target/baseline/liveness fields advance; per-leg ratchet de-advances. The next composite read admits a leg whose `publish_time` exceeds the pre-bug ratchet — including snapshots stale relative to wallclock but newer than the rolled-back ledger — and the resulting cross feeds into `engine.last_oracle_price` via `accrue_market_to`.

Full root-cause analysis, source-line trace, and reproducer in #95 (LiteSVM tests against deployed BPF sha256 `cb6f9c96…545`).

## Cause

The TOTO commit `d40edb5` added the two `MarketConfig` fields and the matching mutations in `read_external_price_e6` but did not extend the open-coded `restored = config_pre_oracle` writeback inside the TradeCpi zero-fill branch. The same omission exists at the partial-catchup `KeeperCrank` writeback (`policy::PartialCrankConfigFields` whitelist at `src/percolator.rs:769-783`) and is addressed by PR #94 (issue #93). The two writeback sites are independent code paths.

## Fix

Carry both per-leg arrays forward from `config` alongside the existing six fields. Two assignment statements added to the open-coded forwarding list, plus a comment explaining the rationale.

The fields are validated by `read_external_price_e6` (rejecting rollback, rejecting same-timestamp disagreement) before they reach the writeback. Persisting the post-read values cannot inject an unvalidated observation — only validated monotonic advances. This matches the source choice for every other oracle-state field already in the forwarded list (`oracle_target_price_e6`, `oracle_target_publish_time`, `last_oracle_publish_time`, `last_effective_price_e6`, `last_good_oracle_slot`).

## Behavioral impact

- **Composite markets (`oracle_leg_count >= 2`):** per-leg ratchet now persists across TradeCpi zero-fill commits, restoring the documented rollback invariant. Real-fill (`exec_size != 0`) paths are unchanged — they already commit the post-read config in full via the regular trade path.
- **Single-leg external markets (`oracle_leg_count == 1`):** `read_external_price_e6` writes only leg index 0; legs 1-2 stay zero across both snapshots, so carry-forward is observationally equivalent to revert. No behavior change.
- **Hyperp markets:** `read_external_price_e6` is not invoked (the Hyperp branch in `get_engine_oracle_price_e6` is taken instead). Per-leg arrays remain zero across the call. No behavior change.
- **ABI / serialization:** no changes to any on-chain layout. Same `MarketConfig` struct as before.
- **Compute cost:** two additional fixed-size array copies per zero-fill writeback (48 bytes total). Negligible.

## Verification

```
cargo check --release
  # clean (only pre-existing solana-client v1.18.26 future-incompat warning)

cargo build-sbf
  # clean (with this branch's wrapper compiled)

cargo test --release --test test_bounty4_tradecpi_zerofill -- --nocapture
  # against the patched wrapper, the four PoC tests from #95
  # invert: the staircase stalls at step 2 with `OracleStale` (per-leg
  # ratchet now rejects the cherry-picked replay), drift stays at 0,
  # and the BOUNTY-WIN panic does NOT fire.

cargo test --release --test unit bug_tradecpi_zerofill -- --nocapture
  # the three host-side companion tests from #95 also flip from
  # pass-as-bug-confirmed to fail-as-bug-closed once this fix lands.
```

The PoC tests from #95 are intentionally not added to this PR — their assertions encode the pre-fix state and would flip once this fix lands. The forward-looking regression-prevention proof would be a Kani harness over the TradeCpi zero-fill writeback (no such harness currently exists at this site, mirroring the Kani coverage gap noted in #95's "Why this wasn't caught" section).

## Coordination with PR #94

PR #94 adds `oracle_leg_publish_times` and `oracle_leg_prices_e6` to `policy::PartialCrankConfigFields` and the matching read/write helpers. Its scope is the partial-catchup `KeeperCrank` writeback (`prog/src/percolator.rs:769-783, 4838-4849`). The TradeCpi zero-fill rollback this PR fixes is open-coded and uses none of the helpers PR #94 modifies (verified: `grep -n "partial_crank_config_to_write\|PartialCrankConfigFields\|apply_partial_crank_config_fields" src/percolator.rs` returns no hits inside the zero-fill branch).

The two PRs are independent and both safe to merge; they fix different code paths.

### DIFF
--- a/src/percolator.rs
+++ b/src/percolator.rs
@@ -8405,6 +8405,22 @@ pub mod processor {
                     restored.oracle_target_price_e6 = config.oracle_target_price_e6;
                     restored.oracle_target_publish_time = config.oracle_target_publish_time;
                     restored.last_hyperp_index_slot = config.last_hyperp_index_slot;
+                    // Per-leg ratchet (`oracle_leg_publish_times` /
+                    // `oracle_leg_prices_e6`) is the documented
+                    // anti-replay defense for composite oracles
+                    // (`MarketConfig::oracle_leg_publish_times` doc-
+                    // comment). `read_external_price_e6` only writes
+                    // these arrays when `publish_time > prev_time`
+                    // (rejecting rollback) and rejects same-timestamp
+                    // disagreement, so the post-read state is a
+                    // validated monotonic advance. Carry it forward
+                    // alongside the other oracle/index state above —
+                    // otherwise the per-leg ratchet de-advances on
+                    // disk while `oracle_target_price_e6` advances,
+                    // and the next composite read admits a stale
+                    // leg paired with a fresh one.
+                    restored.oracle_leg_publish_times = config.oracle_leg_publish_times;
+                    restored.oracle_leg_prices_e6 = config.oracle_leg_prices_e6;
                     state::write_config(&mut data, &restored);
                     state::write_req_nonce(&mut data, req_id);
                     return Ok(());


aeyakovenko:
Closing without merge as stale against current v16 `main` (`00cd16a`).

This PR patches the old `TradeCpi` zero-fill rollback. In v16, zero-fill returns before any market/config/oracle-profile write, so there is no partial restore and no per-leg ratchet state to drop. The associated issue #95 has been closed as stale against v16.

Current v16 wrapper suite: `cargo test --release --test v16_wrapper -- --test-threads=1` => `148 passed`.

PR #115, opened 2026-06-09
Add test: dual-mint shared credit blocks cross-mint double-withdraw

aeyakovenko:
## What

Adds one litesvm end-to-end attack regression: `v16_attack_dual_mint_shared_credit_no_double_withdraw`.

## Attack analyzed (LoF / SOL-002 state asymmetry)

The dual-mint collateral design (`UpdateBaseUnitMints` + `SwapSecondaryForPrimary`) lets a holder deposit **primary-only** but withdraw **either** primary or secondary 1:1. The creative loss-of-funds: if engine credit were tracked per-mint — or the debit were skipped on the secondary path — a single `N` deposit could fund **two** `N` withdrawals (one per mint), minting `2N` collateral from `N`.

`handle_withdraw` → `withdraw_not_atomic` debits one **mint-agnostic** `PortfolioV16::header.capital` and rejects `amount > capital`. The chosen mint only selects which physical vault the SPL transfer drains; the credit is shared. The second cross-mint withdraw therefore hits exhausted capital and reverts.

## Test

Deposits `N` primary, fully withdraws `N` in primary (capital → 0), seeds the secondary reserve with a full `N`, then attempts a second `N` withdraw in the secondary mint. Asserts it reverts, the reserve is byte-identical, and the holder nets exactly `N` (never `2N`). Real BPF via litesvm. **Passes.**

## Why it adds coverage

The nearest existing test, `v16_attack_deposit_primary_only_withdraw_either`, only proves each mint is *individually* withdrawable **while credit still remains** (700 of 1000) — it never drives the cross-mint **exhaustion boundary** where the double-mint would actually occur. This covers the LoF-critical "one deposit cannot be spent through both mints" invariant.

### DIFF
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -35721,6 +35721,94 @@ fn v16_attack_deposit_primary_only_withdraw_either() {
     );
 }
 
+// LoF sweep — dual-mint shared credit (SOL-002 state asymmetry): a deposit credits ONE mint-agnostic
+// engine capital field (`PortfolioV16::header.capital`). The wrapper lets a holder withdraw EITHER the
+// primary or secondary collateral mint 1:1, so the danger is that a single N deposit funds TWO N
+// withdrawals — one per mint — printing collateral out of thin air. This drives that exact double-spend:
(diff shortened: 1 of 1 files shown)


aeyakovenko:

PR #118, opened 2026-06-09
Add test: dual-mint domain insurance blocks cross-mint double-withdraw

aeyakovenko:
## What

Adds `v16_attack_dual_mint_domain_insurance_no_double_withdraw` (litesvm).

## Attack analyzed (protocol-fund LoF — SOL-002 state asymmetry)

`TopUpInsuranceDomain` funds a domain's insurance budget in the **primary** mint only, but `WithdrawInsuranceAsset` (→ `verify_domain_withdrawal_preflight` → `verify_withdrawable_token_accounts`) pays out in **either** the primary or secondary collateral mint. If the per-domain budget were tracked per-mint — or the debit skipped on the secondary route — a single budget of `N` could fund **two** `N` payouts (one per mint), draining the protocol insurance backstop at **2×**.

The budget (`insurance_domain_budget`) is a single mint-agnostic counter. The first full primary payout zeroes it, so the second cross-mint payout reverts.

## Test

Tops asset-0's domain-0 budget to `300`, seeds the secondary reserve with `1000` (so the only barrier is the shared budget, not a thin balance), withdraws the full `300` in primary (budget → 0), then attempts a second `300` in the secondary mint. Asserts it reverts, the reserve is byte-identical, the budget stays `0`, and exactly `300` (never `600`) leaves the protocol. Real BPF via litesvm. **Passes.**

## Why it adds coverage

This is the insurance-path analogue of #115 (user-capital dual-mint double-withdraw), but **independent**: a different handler (`handle_withdraw_insurance_asset`), a different counter (`insurance_domain_budget`), a different authority (insurance operator), and higher severity (the insurance pool backstops every trader's bad debt). No existing insurance/backing test exercises the dual-mint payout path — a regression there would not be caught by #115.

### DIFF
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -27155,6 +27155,103 @@ fn v16_attack_withdraw_insurance_domain_budget_cannot_be_overdrawn() {
     conserve(&env);
 }
 
+// LoF sweep — dual-mint domain-insurance shared budget (SOL-002 state asymmetry, protocol funds).
+// `TopUpInsuranceDomain` is funded PRIMARY-only, but `WithdrawInsuranceAsset` (via
+// verify_domain_withdrawal_preflight -> verify_withdrawable_token_accounts) pays out in EITHER the
+// primary or the secondary collateral mint. The danger: a single domain insurance budget of N funds TWO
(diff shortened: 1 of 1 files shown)


aeyakovenko:

PR #119, opened 2026-06-09
Fix: UpdateBaseUnitMints must reject mismatched-decimals collateral pair (LoF)

aeyakovenko:
## Summary

**This PR changes program behavior** (a one-line-class validation) plus a regression test. The dual-mint collateral feature had a missing invariant that lets `marketauth` install a value-asymmetric collateral pair, enabling user/protocol loss-of-funds.

## The bug (LoF — bounded-admin violation / SOL-002)

The two collateral mints denominate the **same base unit 1:1**:
- `Deposit` is primary-only,
- `Withdraw` pays out in **either** mint,
- `SwapSecondaryForPrimary` moves atoms 1:1.

None of these paths rescale by token decimals — they move raw atoms. But `handle_update_base_unit_mints` validated each mint (`verify_mint`) **without ever comparing their decimals**. So `marketauth` could pair, e.g., a 6-decimals primary with a 9-decimals secondary. Then one atom of each carries different real value, and a holder can:
- deposit the higher-value mint and withdraw the **same atom count** in the lower-value mint (**user LoF**), or
- run the favorable direction to **drain the vault** at the expense of other users.

This violates the README's stated bounded-admin guarantee ("even the asset-0 admin … cannot reach into … a user's collateral").

## The fix

After `verify_mint`, unpack both `Mint.decimals` and reject on mismatch (`InvalidMint`). The check runs only at (re)configuration, which is already gated to an empty market, so no live market is affected.

## Test

`v16_attack_base_unit_mints_must_share_decimals`:
- equal-decimals secondary → configures;
- mismatched-decimals (6 vs primary's 0) → rejects `Custom(10)` (InvalidMint);
- the rejected update does **not** overwrite the configured secondary mint.

Verified the test **fails without the guard** (the asymmetric pair was previously accepted), confirming a real vulnerability rather than a tautology. Existing `v16_attack_base_unit_mints_changeable_only_when_empty`, `..._guarded`, and `deposit_primary_only_withdraw_either` (all use matched 0-decimals mints) still pass.

## Note for reviewers

If mismatched-decimals pairs were ever intended (with out-of-band normalization), this guard would need revisiting — but no decimals normalization exists anywhere in deposit/withdraw/swap, so par-1:1 is the clear design intent.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8848,6 +8848,23 @@ pub mod processor {
         expect_key(secondary_mint_ai, &secondary_key)?;
         verify_mint(primary_mint_ai)?;
         verify_mint(secondary_mint_ai)?;
+        // Both collateral mints denominate the SAME base unit 1:1: deposits are primary-only,
+        // withdrawals pay EITHER mint, and SwapSecondaryForPrimary moves atoms 1:1 — none of these
+        // paths rescale by decimals. If the two mints had different decimals, one atom of each would
+        // carry different value, so a holder could deposit the higher-value mint and withdraw the same
+        // atom count in the lower-value mint (LoF), or drain the vault via the favorable direction.
+        // Pin the pair to equal decimals so even marketauth cannot install a value-asymmetric pair
+        // (bounded-admin: the admin must not be able to reach a user's collateral).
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #120, opened 2026-06-09
Add test: SwapSecondaryForPrimary is accounting-neutral (mints nothing)

aeyakovenko:
## What

Adds `v16_attack_swap_secondary_conserves_and_mints_nothing` (litesvm).

## Why

`SwapSecondaryForPrimary` is a pure 1:1 token exchange between the authority and the vault (primary in, secondary out) — it makes **no engine call**. So a successful swap must leave the engine's collateral accounting (`vault` / `c_tot` / `insurance`) untouched and only rebalance the vault's physical mint composition. If a swap ever leaked into engine accounting it would mint or destroy protocol collateral — a solvency break.

The sibling `v16_attack_swap_secondary_unauthorized_and_bounded` exercises only **rejection** paths (unauthorized, over-swap, zero, bad-dest). Nothing exercised a **successful** swap's conservation.

## Test

Deposits 1000 primary (engine `vault == c_tot == 1000`), seeds the secondary reserve with 400, then the authority swaps 250. Asserts:
1. engine `vault`/`c_tot`/`insurance` are byte-identical (mints/destroys nothing);
2. physical composition moved exactly +250 primary / −250 secondary across the two vaults (and the authority's balances mirror);
3. total custodied tokens across both vaults are conserved.

### DIFF
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -27608,6 +27608,86 @@ fn v16_attack_unexposed_target_move_cannot_grief_live_insurance_withdrawals() {
     );
 }
 
+// LoF sweep — SwapSecondaryForPrimary mints nothing / is accounting-neutral (SOL-002). A SUCCESSFUL swap
+// is a pure 1:1 token exchange between the authority and the vault (primary in, secondary out); it makes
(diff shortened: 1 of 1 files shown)


aeyakovenko:

PR #121, opened 2026-06-09
Add test: third-party asset append cannot freeze a flat withdrawal (DoS isolation)

aeyakovenko:
## What

Adds `v16_attack_third_party_asset_append_cannot_freeze_flat_withdrawal` (litesvm).

## Attack analyzed (market-wide fund-trap DoS — SOL-021)

Appending a permissionless asset advances the market-wide `asset_set_epoch`, which invalidates the health certificate of **every** portfolio in the market (a cert binds `cert_asset_set_epoch == current`). If withdrawal were gated on a *fresh* certificate, an attacker could append assets every slot to keep all victims' certs perpetually stale and **freeze withdrawals market-wide** — a cheap-to-trigger, broad fund-trap.

`withdraw_not_atomic` gates only on **flatness** (`active_bitmap` empty), **not** on cert freshness. A flat holder is therefore immune to `asset_set_epoch` churn.

## Test

A flat victim deposits 1000; an unrelated attacker permissionlessly appends an asset (asserts `asset_set_epoch` actually advances — the grief lever fired); the victim then withdraws the full 1000 and ends with `capital == 0`.

## Why it adds coverage

No existing test references `asset_set_epoch` or this third-party-append isolation property. It guards against a plausible future regression — adding a cert-freshness check to the flat-withdraw path — that would reopen the market-wide DoS.

### DIFF
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -26198,6 +26198,51 @@ fn v16_attack_no_fee_liquidation_cranker_gets_nothing() {
     assert!(g1.vault >= g1.c_tot + g1.insurance, "senior conservation");
 }
 
+// DoS sweep — epoch-grief isolation: a third party cannot trap a flat depositor's funds (SOL-021).
+// Appending a permissionless asset advances the market-wide `asset_set_epoch`, which invalidates the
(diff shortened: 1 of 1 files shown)


aeyakovenko:

PR #122, opened 2026-06-09
Add test: pushed marks cannot override an external-oracle asset (price manipulation)

aeyakovenko:
## What

Adds `v16_attack_pushed_mark_cannot_override_external_oracle_asset` (litesvm).

## Attack analyzed (oracle bypass / price manipulation → LoF)

`PushAuthMark` / `PushEwmaMark` are the manual-mark paths for AuthMark / EwmaMark-mode assets. An asset priced by a **real external oracle** (hybrid mode: Pyth / Switchboard / Chainlink) must **not** accept a pushed mark — otherwise the asset's own oracle authority could bypass the external feed and set **any** settlement price, inducing liquidations or printing PnL against traders.

Both handlers gate on the asset's mode (`profile_is_auth_mark` / `profile_is_ewma_mark`) and reject a mismatch with `Unauthorized`, **before** the authority and slot checks.

## Test

Configures asset 0 as a hybrid (external Pyth oracle) asset, then has the **genuine** oracle authority (admin) attempt to override the feed via `PushAuthMark` and `PushEwmaMark` with an arbitrary `9_999_999` mark. Both reject with `Custom(8)` (Unauthorized), and the externally-seeded `effective_price` is unchanged. (Using the legitimate authority proves the rejection is the *mode* gate, not the authority gate.)

## Why it adds coverage

Existing push-mark tests (`non_authority_cannot_push_auth_mark`, `mark_push_clamped_per_slot`, `extreme_auth_mark_push_rejected_or_safe`, `push_ewma_mark_same_slot_does_not_compound`) cover the authority gate, per-slot clamp, and freshness — **never the mode gate** that protects external-oracle assets from manual-mark override.

### DIFF
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -33431,6 +33431,93 @@ fn v16_attack_marketauth_renounce_rejected_even_with_fallback() {
     );
 }
 
+// LoF sweep — pushed marks cannot override an EXTERNAL-oracle asset (oracle bypass / price manipulation).
+// PushAuthMark/PushEwmaMark are the manual-mark paths for AuthMark/EwmaMark-mode assets. An asset priced
+// by a real external oracle (Pyth/Switchboard/Chainlink, i.e. hybrid mode) must NOT accept a pushed mark,
+// or the asset's own oracle authority could bypass the external feed and set ANY settlement price —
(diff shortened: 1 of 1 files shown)


aeyakovenko:

PR #124, opened 2026-06-09
Add test: empty-portfolio re-init cannot inflate materialized_portfolio_count (slab-close DoS)

aeyakovenko:
## What

Adds `v16_attack_empty_portfolio_reinit_cannot_inflate_materialized_count` (litesvm).

## Attack analyzed (slab-close DoS via counter inflation — SOL-010 / SOL-022)

`InitPortfolio` registers the new (empty) portfolio: `materialized_portfolio_count += 1`. `CloseSlab` requires that count `== 0` (full wind-down). If an already-initialized **empty** portfolio could be re-initialized, it would register a **second** time — leaving a phantom count that never decrements, so the market could **never** be closed/wound down (a cheap, permanent slab-close DoS).

`handle_init_portfolio` rejects with `AlreadyInitialized` (the `is_initialized` guard) **before** the `register_empty_materialized_portfolio_not_atomic` call, so the count cannot be inflated.

## Test

One init → `count == 1`; re-init of the same empty portfolio rejects with `Custom(2)` and the count stays `1` (no phantom); closing the one real portfolio drives the count to `0` (slab close stays reachable).

## Why it adds coverage

Distinct from:
- `init_portfolio_cannot_reinit_funded_victim` — protects victim *capital/ownership*, never checks the counter;
- `close_uninitialized_portfolio_rejects_without_counter_change` — the *close-side* counter (no decrement).

This guards the **init-side** counter integrity that keeps slab close reachable.

### DIFF
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -38271,6 +38271,57 @@ fn v16_attack_init_portfolio_cannot_use_market_as_portfolio_account() {
     assert_eq!(initialized.capital, 0);
 }
 
+// DoS sweep — re-init cannot inflate materialized_portfolio_count (SOL-010 / SOL-022 counter inflation).
+// InitPortfolio registers the new (empty) portfolio: materialized_portfolio_count += 1. CloseSlab requires
+// that count == 0 (full wind-down). If an already-initialized EMPTY portfolio could be re-initialized, it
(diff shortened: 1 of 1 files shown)


aeyakovenko:

PR #151, opened 2026-06-16
[codex] Fix single TradeCpi stale matcher replay

aeyakovenko:
## Summary

Fixes a single `TradeCpi` stale matcher-context replay issue. The single-fill path read matcher output from the matcher-owned context account, but its request id was only `current_slot + 1`, so two same-transaction calls in the same slot could collide. A matcher could write a valid context return for the first call, then return `Ok(())` without rewriting on the second call, and the wrapper would accept the stale first response.

The fix derives the single matcher request id from the current slot, trade identity fields, and the pre-invoke matcher context bytes. If the matcher does not rewrite the context, the stale echoed request id no longer validates. Honest matcher writes still echo the fresh request id and continue to work.

## Changes

- Bind single `TradeCpi` matcher request ids to the pre-invoke matcher context.
- Add a hostile matcher no-write mode for stale replay probes.
- Add a LiteSVM regression that sends two same-transaction single `TradeCpi` calls and asserts the second no-write call cannot replay the first context output.

## Validation

Built BPF artifacts and ran:

- `cargo test v16_attack_hostile_matcher_no_write_cannot_replay_stale_single_context -- --nocapture`
- `cargo test v16_attack_hostile_matcher_single_tradecpi_returns_all_rejected -- --nocapture`
- `cargo test v16_attack_matcher_return_antispoof_rejections -- --nocapture`
- `cargo test v16_attack_hostile_matcher_batch_returns_all_rejected -- --nocapture`
- `cargo test v16_attack_tradecpi_zero_fill_is_clean -- --nocapture`
- `cargo test v16_attack_tradecpi_atomic_fill_vs_capacity -- --nocapture`
- `cargo test v16_attack_tradecpi_limit_price_enforced -- --nocapture`
- `cargo test v16_attack_tradecpi_rejects_unapproved_unsigned_lp_matcher -- --nocapture`
- `git diff --check`

`cargo fmt --check` still reports pre-existing broad formatting churn in `tests/v16_cu.rs`; I only ran targeted formatting on the hostile matcher fixture.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -21,6 +21,7 @@ use solana_program::{
     clock::Clock,
     declare_id,
     entrypoint::ProgramResult,
+    hash::hashv,
     instruction::{AccountMeta, Instruction as SolInstruction},
     program::{invoke, invoke_signed},
     program_error::ProgramError,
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as duplicate/superseded by #158, which was rebased and merged into main at 9a9500a as the canonical matcher-output freshness fix.

PR #153, opened 2026-06-19
[codex] Fix stale matcher return replay

aeyakovenko:
## Summary

Fixes a stale matcher return replay in `TradeCpi`.

Root cause: the single-fill matcher CPI used `current_slot + 1` as `req_id` and read matcher output from the matcher context account. Two same-transaction calls could therefore share the same request id, and a matcher that wrote a valid response once then returned success without writing could leave stale bytes available for replay.

Changes:
- store a wrapper-owned matcher request sequence in the LP portfolio matcher config (`enabled` low bit remains the enabled flag; high bits are the sequence)
- increment that sequence before `TradeCpi` / `BatchTradeCpi` matcher CPI and bind returns to it
- route single `TradeCpi` through the same tag-3 return-data ABI as batch, instead of reading a ctx-account return prefix
- add a hostile matcher LiteSVM regression where the second same-transaction call emits no return data and must fail atomically
- update positive matcher tests and README docs for the return-data path

## Validation

- `cargo fmt`
- `cargo check --tests`
- `(cd tests/fixtures/hostile_matcher && cargo build-sbf)`
- `cargo test v16_attack_hostile_matcher_no_write_cannot_replay_stale_single_return_data --test v16_cu -- --nocapture`
- `cargo test v16_attack_hostile_matcher --test v16_cu -- --nocapture`
- `cargo test v16_bpf_tradecpi --test v16_cu -- --nocapture`
- `cargo test` (540 `v16_cu` tests + unit/doc tests passed)
- `git diff --check`

Kani note: `cargo kani --tests` and `cargo kani --tests --no-default-features` both fail before wrapper harnesses run while compiling pinned engine `percolator@bbe745a`, at `PortfolioAccountV16Account` size assertion `== 9291`. I did not fold that unrelated engine/Kani blocker into this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,7 +319,10 @@ Authority fields are split by scope:
 - **`AssetOracleProfileV16.asset_admin`**: per-asset cold key that rotates that asset's scoped authorities
 - **`AssetOracleProfileV16.insurance_authority` / `insurance_operator` / `backing_bucket_authority` / `oracle_authority`**: per-asset operational authorities
 
-Matcher requests do not use a persisted market nonce. The wrapper invokes the matcher and requires the response to echo the request id, LP identity, asset index, oracle price, and requested size/sign constraints.
+Matcher requests use a wrapper-owned sequence stored in the LP portfolio matcher config. The low
+bit is the enabled flag and the high bits are the request sequence. The wrapper increments it before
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Closing as duplicate/superseded by #158, which was rebased and merged into main at 9a9500a as the canonical matcher-output freshness fix.

PR #159, opened 2026-06-23
[codex] Reject nonprogress work cranks

aeyakovenko:
## Summary
- propagate `EngineNonProgress` for work-bearing public cranks (`close_q > 0`)
- preserve existing refresh-only oracle-update behavior for `close_q == 0`
- add a LiteSVM regression where an unrelated oracle observation cannot make liquidation work appear successful without selected progress

## Root Cause
The wrapper treated `NonProgress` as success whenever any observation was supplied. That is valid for refresh-only oracle updates, but for a `close_q > 0` work crank it allowed a stale or wrong-task transaction to commit unrelated oracle state and return success while the selected liquidation step did not run.

## Validation
- `cargo fmt --check`
- `cargo test --all-targets`

### DIFF
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -40895,6 +40895,151 @@ fn v16_attack_crank_oracle_feed_id_mismatch_rejects_without_mutation() {
     assert_eq!(group.assets[0].raw_oracle_target_price, 210_000);
 }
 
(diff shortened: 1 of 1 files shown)


aeyakovenko:
Closing as duplicate/superseded by #170, merged into main at 6a394bf for the nonprogress work-crank behavior.

PR #166, opened 2026-06-23
Reject unrequested duplicate active legs before CPI

aeyakovenko:
## Summary

Reject malformed active-leg state account-globally during CPI trade currentness preflight. The previous request-scoped helper only inspected active legs for requested assets, so a duplicate active leg on an unrelated asset could survive until after the external matcher CPI.

This keeps the fresh-requested-asset currentness behavior live, but validates the fixed portfolio leg shape from raw POD fields before handing a hostile matcher writable context.

## Validation

- `cargo fmt --check`
- `cd tests/fixtures/auth_matcher && cargo build-sbf`
- `cargo test --test v16_cu unrequested_duplicate_active_leg -- --nocapture`
- `cargo test --test v16_cu -- --nocapture` (539/539 passed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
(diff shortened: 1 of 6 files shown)


aeyakovenko:
Added the raw hidden-leg branch coverage (inactive-nonempty and invalid-active-byte on an unrequested asset). Revalidated: `cargo test --test v16_cu -- --nocapture` (540/540 passed).

PR #135, opened 2026-06-14
[COVERAGE] Whole-route LoF/DoS invariant certification

aeyakovenko:
## Scope

Invariant-driven whole-route security submission for the deployed wrapper, pinned exactly to engine commit `495a5590c97055bd71c6f94d849ff0298f243145` from [engine PR 195](https://github.com/aeyakovenko/percolator/pull/195).

This branch:

- defines the normative 89-invariant charter in `INVARIANTS.md`;
- organizes public LiteSVM, stateful, CU, metamorphic, bounded-reachability, and wrapper-local Kani evidence by invariant;
- checks entitlement, matched-book integrity, economic and administrative liveness, history relations, and adversarial-role containment over complete public routes;
- accepts LoF/persistent-DoS evidence only from public traces without out-of-band program-state mutation;
- keeps engine-owned arithmetic and transition proofs in the engine rather than duplicating them in the wrapper;
- includes the wrapper simplifications and fixes independently required by those invariant oracles.

## Current checkpoint

Wrapper head: `1c325df80fce8d9d99c10e9b7ba5a88697841577`

Engine pin: `495a5590c97055bd71c6f94d849ff0298f243145`

The current-surface ledger records 87 invariants `CLOSED` and 2 rigorously `N/A`, conditional on the named arithmetic, authenticated-input, SVM rollback, submission/fairness, and authority-participation assumptions. This is not an unconditional claim over future routes, pins, layouts, or unsupported shapes. Exact evidence, reopening conditions, commands, and known audit-mode baselines are maintained in `tests/invariants/README.md`.

The final entitlement, matched-book, liveness-separation, history-relation, adversarial-role, and 22-layer public-success composition gates pass at this head. GitHub has no hosted checks configured for this branch; the reproducible local verification surface is documented in the invariant README.

## Consolidation

This submission replaces the prior wrapper finding stack. The frozen 143-row finding benchmark remains in-tree for traceability: 126 qualifying findings map to generic invariant-owned public-route evidence and integrated fixes, while 17 carry executable nonqualifying dispositions. Nineteen older unlabelled or hardening branches were also closed after individual supersession or acceptance-rule review. No remote branches were deleted.

## Merge order

1. Merge engine PR 195 while preserving commit `495a5590c97055bd71c6f94d849ff0298f243145`, or repin this PR to the resulting engine commit.
2. Merge this wrapper PR.

The engine dependency must land first because Cargo pins its exact revision.

PR #181, opened 2026-07-01
[codex] Fix no-observation refresh active-leg scan

aeyakovenko:
## What changed

`PermissionlessCrank` no longer checks only the engine-selected first active leg when deciding whether a no-observation stale/liquidatable crank may proceed. It now scans every active unique portfolio-leg asset and rejects the crank if any active price-managed asset has pending mark movement that was not supplied in the public observation hints.

## Root cause

The wrapper preflight used the auto-crank-selected asset as a proxy for the whole account. A portfolio with asset 1 in the first active slot and asset 0 in a later active slot could stage a pending AuthMark/EWMA move on asset 0, advance the market epoch through an unrelated asset, and then submit a no-observation `PermissionlessCrank` for the target account. The old wrapper allowed the account to certify while asset 0 still had pending mark work.

That was public-interface reachable and could let health/progress be computed against stale internal price state for a later active leg.

## Coverage

- Adds a LiteSVM repro proving no-observation refresh cannot certify over a pending mark on a later active leg.
- Proves the live path still works when the caller supplies the later active asset observation.
- Adds max-shape AuthMark and Hybrid/Pyth refresh probes so the broader all-active-leg preflight is covered for CU/liveness.
- Updates existing cross-margin/backing tests to supply complete public observations when both active mark legs are pending.

## Validation

- `cargo test --test v16_cu -- --quiet`
- `cargo test -- --quiet`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10602,17 +10602,14 @@ pub mod processor {
             let summary = group
                 .build_actionable_summary(&portfolio.as_view())
                 .map_err(map_v16_error)?;
-            if let Some(asset_index) =
-                auto_crank_selected_asset_that_accrues_view(&portfolio, &summary)?
-            {
-                reject_missing_pending_selected_observation_view(
-                    &cfg,
-                    &group,
-                    asset_index,
-                    authenticated_now_slot,
-                    observations.as_slice(),
-                )?;
-            }
+            reject_missing_pending_account_observations_view(
+                &cfg,
+                &group,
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closing under the strict PR rule: this proves stale certification state, but not an end-to-end public LoF, blocked exit, or absent bounded continuation. Keep/rework it as a #135 probe only if it demonstrates a concrete user outcome.

PR #185, opened 2026-07-03
Keep batch exits live under backing-fee policies

aeyakovenko:
## Summary

- remove the market-global batch rejection when any backing-fee policy is active
- snapshot each account's source backing and run the existing fee debit/routing post-pass after the shared batch engine call
- prove positive lien-growth selection and multi-domain fee conservation; exercise both direct and CPI production routes in LiteSVM

## Security impact

A permissionless asset creator could install a backing-fee policy on an unrelated asset and globally reject every batch. A user whose spread is healthy only after both legs execute could not unwind one leg at a time, so the global gate was a cross-asset account DoS.

Removing only the gate was not sufficient: it made batches silently skip backing-domain utilization fees. The fix preserves atomic final-state margin semantics while charging only positive source-lien growth and routing each valid split to provider earnings and domain insurance.

## PDD evidence

- Red liveness: restoring the old market-global gate makes `v16_attack_unrelated_backing_fee_policy_does_not_block_atomic_batch_exit` fail with `Custom(9)`.
- Red conservation: removing only the new batch fee post-pass makes `v16_attack_batch_backing_fee_split_conserves` fail because the charged fee remains zero.
- Kani proves the full-width `u128` selector returns exactly positive lien growth and cannot underflow; positive, equal, and decreasing cases are all covered.
- Kani proves two independently valued domain splits aggregate exactly to provider plus insurance routing and malformed splits fail closed; both covers are satisfied.
- LiteSVM proves unrelated zero-fee accounting remains byte-for-value unchanged and both direct and CPI batches charge/conserve the intended domain fee.

## Validation

- `cargo test --all-targets`: 5 unit + 566 LiteSVM/SBF tests passed
- both targeted Kani harnesses passed in under one second of verification time; 5/5 cover properties satisfied
- `cargo clippy --all-targets` (existing warnings only)
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4080,6 +4080,31 @@ pub mod oracle_v16 {
 pub mod policy_v16 {
     use crate::constants::MAX_DYNAMIC_TRADE_FEE_BPS;
 
+    #[inline(always)]
+    pub fn positive_lien_growth_num(before_num: u128, after_num: u128) -> Option<u128> {
+        if after_num <= before_num {
+            return None;
+        }
+        Some(after_num - before_num)
+    }
+
+    #[inline(always)]
+    pub fn accumulate_backing_fee_split(
+        accumulated: u128,
+        split: percolator::BackingDomainFeeSplitV16,
+    ) -> percolator::V16Result<u128> {
+        let routed = split
+            .provider_fee
+            .checked_add(split.insurance_fee)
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying for persistent DoS. The batch policy rejection does not strand funded users because the canonical single-leg risk-reduction routes remain available; PR135 retains the generic route and liveness coverage.

PR #187, opened 2026-07-07
Preserve proportional BatchTradeCpi partial fills

aeyakovenko:
## Security premise

`BatchTradeCpi` validated each matcher return independently. A configured LP matcher could mark every leg partial-valid while filling hedge legs at different requested-size ratios. Every per-leg bound passed, but an atomic spread could become unintended directional exposure, creating a user loss surface.

## PDD fix

- Require every nonzero matcher fill to have the same exact `abs(exec_size) / abs(request_size)` ratio as the first leg.
- Compare ratios by cross multiplication, avoiding division and rounding ambiguity.
- Enforce the engine's `MAX_TRADE_SIZE_Q` bound before matcher CPI, making those products infallible for every executable request.
- Fast-path identical ratio pairs and all-full/full-versus-partial cases, avoiding `u128` multiplication on normal full batches.
- Keep proportional partial batches live and preserve mixed signed directions.

The public LiteSVM regression uses unequal requested magnitudes and opposite directions. A hostile full/half return rejects atomically with market, taker, LP, and matcher context unchanged; a common half-fill then executes both legs at the exact expected sizes.

## Proofs

- `kani_v16_batch_fill_ratio_cannot_overflow_for_any_engine_valid_size` invokes the exact production size preflight and proves both cross-products fit for all valid signed request magnitudes. It reaches accepted and rejected ratios plus mixed-sign values wider than 32 bits.
- `kani_v16_batch_fill_ratio_rejects_full_vs_partial_and_keeps_common_ratio_live` proves a symbolic full/partial mismatch always rejects while the same arbitrary partial ratio remains live.
- `kani_v16_batch_fill_ratio_preserves_signed_spread_shape` proves same- and opposite-direction partial legs preserve their requested relation.
- All 6 covers are reachable; every harness finishes in under one second locally.

For the red step, the production ratio predicate was replaced with unconditional acceptance. Kani failed its full-versus-partial invariant immediately, and the public hostile batch executed instead of rejecting. Restoring the exact predicate makes both layers pass.

## Verification

- `cargo test --all-targets`: 5 unit + 564 LiteSVM passed
- All 3 affected Kani harnesses: all checks and 6/6 covers passed
- Hostile and authenticated matcher fixture `cargo build-sbf`
- Existing 14-leg `BatchTradeCpi` and non-CPI transaction-CU guards passed
- `cargo clippy --all-targets` (existing test warnings only)
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4080,6 +4080,29 @@ pub mod oracle_v16 {
 pub mod policy_v16 {
     use crate::constants::MAX_DYNAMIC_TRADE_FEE_BPS;
 
+    #[inline(always)]
+    pub fn batch_trade_size_allowed(size_q: i128) -> bool {
+        size_q != 0 && size_q != i128::MIN && size_q.unsigned_abs() <= percolator::MAX_TRADE_SIZE_Q
+    }
+
+    #[inline(always)]
+    pub fn batch_fill_ratios_equal(
+        exec_abs: u128,
+        req_abs: u128,
+        first_exec_abs: u128,
+        first_req_abs: u128,
+    ) -> bool {
+        if exec_abs == first_exec_abs && req_abs == first_req_abs {
+            return true;
+        }
+        let full = exec_abs == req_abs;
+        let first_full = first_exec_abs == first_req_abs;
+        if full || first_full {
+            return full && first_full;
+        }
+        exec_abs * first_req_abs == first_exec_abs * req_abs
+    }
+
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying for public LoF. The caller selected the configured matcher and the pinned route keeps fills inside the economic envelope; PR135 standardizes atomic batch authorization on exact requested quantities and retains the hostile-matcher route matrix.

PR #188, opened 2026-07-07
Bind CPI limits to actual engine execution price

aeyakovenko:
## Security premise

`TradeCpi` and `BatchTradeCpi` previously checked the user's limit against only the matcher-reported price. The wrapper could then normalize that quote and settle through the engine at a different accepted price. A hostile matcher could therefore report a limit-compatible price while causing execution beyond the user's bound, producing a direct slippage/loss-of-funds surface.

## PDD fix

- Keep the cheap early check against the matcher-reported price.
- Apply the same directional limit predicate to the accepted price actually passed to the engine.
- Share one inlined production predicate between single and batch paths.
- Preserve signed partial-fill semantics and disabled-limit behavior.

The public LiteSVM regression proves that hostile tight-limit single and batch requests reject atomically with market, taker, LP, and matcher context unchanged. Permissive limits execute both paths at exactly the requested partial sizes.

## Proofs

- `kani_v16_cpi_limit_binds_reported_and_engine_accepted_prices` proves exact full-width buy, sell, and disabled-limit semantics. Acceptance implies both reported and accepted engine prices satisfy the bound. All 5 covers are reachable, including hidden internal-price rejection in both directions.
- `kani_v16_cpi_partial_fill_preserves_limit_direction` proves every valid nonzero signed partial fill retains the request's limit direction. Both long and short partial-fill covers are reachable.
- Both harnesses finish in under one second locally.

For the red step, the production predicate was mutated to ignore the accepted engine price. The first theorem failed, both hidden-price rejection covers became unreachable, and the public tight-limit attack was accepted. Restoring the two-price predicate makes all proofs and regressions pass.

## Verification

- `cargo test --lib`: 5 passed
- `cargo test --all-targets`: 569 passed
- Both affected Kani harnesses: 7/7 covers, all checks passed
- Hostile matcher fixture `cargo build-sbf`
- `cargo clippy --all-targets` (existing test warnings only)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4080,6 +4080,27 @@ pub mod oracle_v16 {
 pub mod policy_v16 {
     use crate::constants::MAX_DYNAMIC_TRADE_FEE_BPS;
 
+    #[inline(always)]
+    pub fn limit_price_allows(size_q: i128, exec_price: u64, limit_price: u64) -> bool {
+        limit_price == 0
+            || if size_q > 0 {
+                exec_price <= limit_price
+            } else {
+                exec_price >= limit_price
+            }
+    }
+
+    #[inline(always)]
+    pub fn cpi_limit_prices_allow(
+        size_q: i128,
+        reported_exec_price: u64,
+        accepted_exec_price: u64,
+        limit_price: u64,
+    ) -> bool {
+        limit_price_allows(size_q, reported_exec_price, limit_price)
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying for terminal LoF/DoS. The engine clamps the accepted price to the fair-mark envelope, so the reproduced limit mismatch did not create unauthorized withdrawable value or a persistent exit lock. PR135 retains the CPI limit and exact-outcome coverage.

PR #189, opened 2026-07-08
[PDD] Prove matcher replay safety across LP close/reinit

aeyakovenko:
## Summary

Adds proof-driven coverage for stale matcher-context replay across an LP portfolio close/reinitialize lifecycle. The behavioral fix is already on `main`: matcher freshness is a market-level sequence, so destroying and reusing an LP portfolio cannot make an old matcher response current again.

The production increment is factored through an inlined `next_matcher_request_id` kernel without changing behavior.

## Proof-driven development

`kani_v16_matcher_request_ids_are_strict_unique_and_fail_closed` proves over every `u64` state that:

- every issued request ID is exactly the predecessor plus one;
- successive issued IDs are strictly increasing and distinct;
- ID zero can never be reached by wraparound;
- `u64::MAX` is issued once from `MAX - 1`; and
- exhaustion at `MAX` rejects instead of wrapping or replaying.

All 52 checks pass, all three normal/terminal/exhaustion covers are satisfied, and solver time is about 0.1 seconds.

The public LiteSVM lifecycle regression then covers the storage-scope property that arithmetic alone cannot prove:

1. A real CPI fill leaves a valid response bound to market request ID 1.
2. The LP position is closed, capital withdrawn, and the same portfolio account closed and reinitialized.
3. A hostile matcher returns successfully without writing fresh context; the stale ID-1 response is rejected and market, taker, LP, and context accounts all roll back byte-for-byte.
4. A newly written response remains live and succeeds at market request ID 2.

**Red mutation:** resetting `matcher_req_seq` in the portfolio-close path makes the regression fail immediately after close/reinitialize (`0 != 1`). This recreates the reported storage-scope bug and demonstrates that the test is not merely confirming current output.

## Verification

- affected Kani harness: passed, 3/3 covers satisfied

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -1376,9 +1376,7 @@ pub mod state {
     pub fn bump_matcher_req_seq(data: &mut [u8]) -> Result<u64, ProgramError> {
         check_header(data, KIND_MARKET)?;
         let mut config = read_wrapper_config_from_bytes(data)?;
-        config.matcher_req_seq = config
-            .matcher_req_seq
-            .checked_add(1)
+        config.matcher_req_seq = super::policy_v16::next_matcher_request_id(config.matcher_req_seq)
             .ok_or(PercolatorError::InvalidInstruction)?;
         let req_id = config.matcher_req_seq;
         write_wrapper_config_to_bytes(data, &config)?;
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as coverage-only. The production behavior is already on main; the public close/reinitialize replay regression is being retained in the coverage-only rebuild of #135.

PR #114, opened 2026-06-09
[codex] Add public-interface authority isolation tests

aeyakovenko:
## Summary

Adds thirty-five LiteSVM adversarial tests plus one public-interface base-unit mint validation fix for public-interface authority isolation, oracle integrity, stale/live liveness, custody hardening, dual-mint/base-unit accounting, lifecycle rollback, liquidation reward isolation, terminal/recovery boundaries, and bounded batch paths.

- `PushEwmaMark` cross-asset isolation: asset-1's real oracle authority cannot push asset-0 EWMA state; asset 0's own authority still can.
- External-oracle push mode guard: the correct oracle authority cannot use `PushAuthMark` or `PushEwmaMark` to override a hybrid/Pyth-priced asset; rejected pushes leave market bytes unchanged.
- No-feed and hybrid oracle reconfiguration isolation: asset-1 oracle authority cannot reconfigure asset 0 through manual or hybrid oracle paths, with positive controls for the rightful authority.
- Non-Pyth oracle binding/freshness/confidence: Switchboard and Chainlink provider-shaped accounts bind by key/owner and reject stale or wide-confidence data without market mutation.
- Chainlink malformed-field parsing: bad discriminator, zero version/round/live-length/slot/timestamp, decimals above bound, and negative answer reject as `OracleInvalid` without mutation.
- Pyth exponent-bound parsing: out-of-bound and extreme in-bound Pyth exponents reject gracefully as `OracleInvalid`; a normal exponent configures.
- Hybrid liquidation reward-tail rollback: a valid hybrid oracle tail plus malformed reward tail rejects atomically without persisting oracle, liquidation, or reward-account mutation.
- Hybrid oracle timestamp replay: same-feed/same-publish-time price rewrites reject without mutating market or portfolio state.
- Stale-matured live-entrypoint freeze: `Withdraw`, `Deposit`, `TopUpInsurance`, and `TradeNoCpi` reject before permissionless stale resolution, while the resolution path still works.
- Asset append epoch-grief isolation: an unrelated permissionless asset append advances `asset_set_epoch` but cannot freeze a flat victim withdrawal.
- Market-authority handoff: stale marketauth cannot mutate admin paths or execute secondary swaps after rotation; the new key can.
- Secondary swap accounting neutrality: successful `SwapSecondaryForPrimary` on a funded market leaves engine market bytes/accounting unchanged while conserving total primary-plus-secondary vault atoms.
[...]

PR #190, opened 2026-07-08
[BLOCKER DoS] Commit recovery-required auto-crank progress

aeyakovenko:
## What changed

- Pin engine PR #114, where the auto-crank commits a complete reactive Recovery terminal as `Ok(RecoveryDeclared)` without duplicate hot-path scans.
- Route an already-Recovery market directly through the same public engine auto-crank API before Live-mode summary/oracle processing, allowing the bounded Recovery-to-Resolved step.
- Keep malformed Live inputs and all non-Recovery engine errors fail-closed; the wrapper does not swallow `RecoveryRequired`.
- Rebase onto current `main` and the engine-selected liquidation API.

## PDD evidence

- Red, engine parent `143e68c`: the public liquidation crank fails `Custom(23)` after 95,157 CU; SVM rollback erases Recovery.
- Green: the same public transaction commits Recovery without changing target portfolio bytes, vault, `c_tot`, or insurance.
- A second empty-observation public crank reaches Resolved, changes only terminal market state, and preserves the target portfolio and value aggregates.
- Cross-margin deep-insolvency wind-down and ordinary engine-selected liquidation conservation regressions pass.

## Validation

- Exact recovery-required public regression: passed
- Cross-margin Recovery-to-Resolved regression: passed
- Issue-103 partial-liquidation 325k CU guard: passed
- `cargo test --all-targets --no-default-features`: 5 unit + 564 LiteSVM tests passed
- Engine Kani theorem in PR #114: 296 checks, 4/4 covers, 0.126s
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=bd77e34739087a2f442792374e228af0ddd5aeba#bd77e34739087a2f442792374e228af0ddd5aeba"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #192, opened 2026-07-09
[codex] Fix CPI limit checks after price normalization

aeyakovenko:
## Summary

Fixes a public-interface CPI slippage/value bug in `TradeCpi` and `BatchTradeCpi`.

A matcher could return a raw `exec_price_e6` that satisfied the user limit while the wrapper later normalized that reported price through the market/oracle profile before sending it into the engine. For EWMA/hybrid/mark-priced paths, the normalized engine price can differ from the raw matcher price, so the user limit was not actually enforced on the price used by the engine.

This PR enforces CPI limits on both values:

- the raw matcher-returned `exec_price_e6`
- the normalized accepted engine price from `accepted_reported_trade_price_view`

That keeps existing raw-price rejection behavior while closing the normalized-price bypass.

## Tests

- Added a hostile matcher mode that returns a valid low nonzero price (`exec_price_e6 = 1`).
- Added LiteSVM public-interface regressions for both `TradeCpi` and `BatchTradeCpi` proving a buy limit of `1` rejects when the normalized engine price violates the limit, with market/taker/LP rollback checks.
- Verified existing raw limit tests still pass.

Validation run locally:

- `cargo fmt`
- `cd tests/fixtures/auth_matcher && cargo build-sbf`
- `cargo test --test v16_cu normalized_engine_price_violation --features test-sbf -- --nocapture`
- `cargo test --test v16_cu limit_price --features test-sbf -- --nocapture`
- `cargo test --test v16_cu per_leg_limit --features test-sbf -- --nocapture`
- `cargo test --test v16_cu --features test-sbf` (`561 passed`)
- `cargo test --features test-sbf` (`561 passed`)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -6608,12 +6608,15 @@ pub mod processor {
             req_id,
         )?;
         if limit_price != 0 {
-            let limit_ok = if size_q > 0 {
-                ret.exec_price_e6 <= limit_price
-            } else {
-                ret.exec_price_e6 >= limit_price
-            };
-            if !limit_ok {
+            if !cpi_limit_allows_price(size_q, limit_price, ret.exec_price_e6) {
+                return Err(PercolatorError::InvalidInstruction.into());
+            }
+            let engine_price = accepted_reported_trade_price_from_market_account(
+                market_ai,
+                asset_index as usize,
+                ret.exec_price_e6,
+            )?;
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as duplicate of #188, which already covers the CPI internal/normalized price limit bug. The batch-specific coverage from this branch can be folded into the existing PR if needed.

PR #193, opened 2026-07-09
[PDD] Allow safe hybrid soft-stale crank without oracle tail

aeyakovenko:
## Summary

Fixes a permissionless crank liveness mismatch for hybrid assets. Once both the configured soft-stale slot window and the persisted oracle sample's time-staleness window have elapsed, the engine can make bounded progress from the internally maintained, fee-supported EWMA. The wrapper previously rejected a zero-account observation before that fallback path was reachable.

The account-count policy is now one inlined production predicate shared by preflight and price execution. It accepts either the exact configured account count or an exact zero-account hybrid fallback; every nonzero mismatch remains rejected. Missing-tail fallback additionally requires a nonzero configured feed count, hybrid mode, strict soft-stale maturity, and a strictly stale persisted sample.

## Proof-driven development

Two full-width Kani harnesses exercise the production predicate:

- `kani_v16_hybrid_missing_tail_admission_is_exact_and_non_bypassable` proves the admission rule is exact, malformed nonzero counts cannot enter fallback, slot/time boundaries remain closed, and accepted fallback implies a genuinely stale persisted sample. All five covers are satisfied.
- `kani_v16_hybrid_missing_tail_liveness_is_monotone_after_maturity` proves that once fallback is admitted, nondecreasing authenticated slot and Unix clocks cannot make it unavailable. Its non-vacuity cover is satisfied.

Both harnesses finish with about one second of solver time.

**Red mutation:** restoring the old exact-count-only rule makes the admission theorem fail and makes its mature-fallback cover unsatisfiable. The public LiteSVM crank also fails with `Custom(9)` after 12,753 CU, before any engine progress.

**Green public path:** the LiteSVM regressions prove that:

- fresh or time-fresh samples cannot omit the configured oracle tail and reject atomically;
- the normal authenticated oracle-tail path remains live;
- a genuinely soft/time-stale hybrid can crank without an external account;
- one crank makes bounded accrual and valuation progress within the CU guardrail; and
- fallback leaves vault, capital-total, insurance, account capital, and position size unchanged.

## Verification

- both affected Kani harnesses: passed, all 6 covers satisfied
- `cargo test --all-targets`: 570 passed, 0 failed

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -3593,11 +3593,11 @@ pub mod oracle_v16 {
         profile: &AssetOracleProfileV16,
         now_unix_ts: i64,
     ) -> bool {
-        if profile.max_staleness_secs == 0 || profile.oracle_target_publish_time == 0 {
-            return false;
-        }
-        let age = now_unix_ts.saturating_sub(profile.oracle_target_publish_time);
-        age >= 0 && age as u64 > profile.max_staleness_secs
+        super::policy_v16::stored_oracle_sample_stale(
+            profile.max_staleness_secs,
+            profile.oracle_target_publish_time,
+            now_unix_ts,
+        )
     }
 
     pub fn hard_stale_matured(config: &WrapperConfigV16, now_slot: u64) -> bool {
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as non-DoS policy/ergonomics: a correct crank can still supply the configured oracle account, so omitting that otherwise available tail is not an absent bounded progress path.

PR #194, opened 2026-07-09
[audit] Prove stale defensive rebalance remains live

aeyakovenko:
## Audit result

The original production change was rejected by proof-driven analysis. A local-stale guard on `RebalanceReduce` contradicts the frozen specification, which explicitly permits defensive rebalancing while stale, and gives a stale oracle operator a way to block an owner from reducing risk.

This revision removes that guard. The final diff is a production-path LiteSVM regression proving the required liveness and conservation properties.

## PDD red/green

**Red:** reintroducing the proposed profile-aware stale guard makes the regression fail with `Custom(19)` before the owner can reduce risk. The rejected path consumed 95,292 CU and left the stale operator able to block exit.

**Green:** the public `RebalanceReduce` path:

- opens real non-base long/short exposure before staleness;
- isolates asset-local staleness while the base market remains fresh;
- lets the owner reduce exactly toward zero within the custody CU limit;
- preserves `vault`, `c_tot`, insurance, both users' capital, and both users' PnL;
- clears effective long/short OI without mutating the counterparty account;
- permits the counterparty to forfeit the orphaned recovery leg; and
- lets both users withdraw exactly their original 1,000,000 atoms, ending with zero vault, capital total, and insurance.

The engine dependency at `143e68c` separately proves the production `kernel_reduce_position_delta` contract over symbolic inputs: exact toward-zero reduction, no over-close, and strict progress. A wrapper-only boolean policy harness would be weaker than this engine contract plus the full SBF lifecycle regression.

## Verification

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8409,11 +8409,23 @@ pub mod processor {
             if group.header.mode == 0 && permissionless_resolve_matured_now_view(cfg, group) {
                 return Err(V16Error::LockActive);
             }
+            let asset_index_usize = asset_index as usize;
+            let profile = read_oracle_profile_from_view(group, cfg, asset_index_usize)
+                .map_err(|_| V16Error::InvalidConfig)?;
+            if group.header.mode == 0
+                && global_or_profile_resolve_matured_at_slot(
+                    cfg,
+                    &profile,
+                    authenticated_market_slot_or_fallback_view(group),
+                )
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closing as coverage-only. It deliberately contains no bug fix, and the substantive stale owner-reduction lifecycle is retained in #135.

PR #195, opened 2026-07-09
[codex] Bound stale asset shutdown authority

aeyakovenko:
## Summary

Preserves the local-stale shutdown guard for a permissionless asset admin without allowing an abandoned oracle/admin to block asset-0 market-authority shutdown forever.

## PDD

The public LiteSVM regression creates asset 1, opens balanced positions, keeps the base market fresh, and lets only asset 1 become stale. It proves:

- the local asset admin cannot freeze a stale mark;
- the market authority can still move the abandoned asset into Recovery at the committed last-good mark;
- both positions remain open during the configured exit window;
- permissionless force-close rejects before the exact timeout and succeeds at maturity;
- vault accounting and token custody remain conserved.

The prior blanket guard failed the market-authority liveness case.

## Validation

- `cargo test --test v16_cu v16_attack_non_base_local_stale_admin_cannot_block_delayed_shutdown -- --nocapture`
- `cargo test shutdown --test v16_cu -- --nocapture`

No merge requested.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9214,11 +9214,11 @@ pub mod processor {
                 if !marketauth_authorized && !asset_admin_authorized {
                     return Err(PercolatorError::Unauthorized.into());
                 }
-                reject_permissionless_resolve_matured_live_for_profile_view(
-                    &cfg,
-                    &profile,
-                    &group,
-                )?;
+                if !marketauth_authorized {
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closing under the strict impact gate. This is an authority-policy hardening probe, but it does not demonstrate user value loss, broken conservation, or an unavailable bounded wind-down path.

PR #196, opened 2026-07-09
[codex] Reject local-stale domain withdrawals

aeyakovenko:
## Root cause

The shared live-domain withdrawal gate checked global resolve maturity, market stress/lock state, local loss staleness, and target/effective lag, but not the selected non-base profile's local oracle-stale clock.

## Impact and red

A permissionless asset authority could keep asset 0 fresh, let its own price-managed asset mature locally with open OI, then remove protection before a fresh local mark or wind-down. With the profile-aware gate manually removed on current `main`, independent red runs confirm that the stale authority can move:

- 50 insurance atoms
- 50 backing-principal atoms
- 10 provider-earnings atoms

Each amount reaches the authority’s token account before the regression rejects the behavior.

## Fix

`live_domain_withdraw_health_or_shutdown_view` now loads the selected asset profile and applies the profile-aware resolve-maturity gate before allowing live insurance, backing-principal, or provider-earnings withdrawals. The terminal empty-and-matured shutdown escape remains first, preserving wind-down liveness.

## PDD coverage

`v16_attack_non_base_local_stale_domain_withdrawals_reject` creates a permissionless non-base AuthMark asset, publicly funds its insurance and backing, opens real OI, keeps asset 0 fresh, and matures only asset 1. It proves all three withdrawals reject atomically, then refreshes asset 1 and proves insurance, backing principal, and earnings are all withdrawable again.

Provider earnings are installed as valid vault-backed test state because this regression targets the withdrawal gate, not the separate source-lien accrual path. Insurance and backing principal are funded through public instructions.

`kani_v16_price_managed_profile_resolve_boundary_is_exact_and_live` invokes the production predicate over full-width symbolic slots. It proves exact/post-boundary maturity, pre-boundary liveness, regressed-slot safety, and unmanaged/disabled-profile liveness. All 5 covers are reached in 0.21 seconds.

## Verification

- `cargo test --all-targets`: 569 passed (5 unit + 564 LiteSVM)
- new Kani harness: successful, 5/5 covers reached
- `cargo clippy --all-targets` (passes with existing baseline warnings)

Head: `b24f063cc228ed8a1963497ec2b40431af2ab16d`.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4422,6 +4422,8 @@ pub mod processor {
         if shutdown_asset_empty_and_matured_now_view(cfg, group, asset_index).is_ok() {
             return Ok(true);
         }
+        let profile = read_oracle_profile_from_view(group, cfg, asset_index)?;
+        reject_permissionless_resolve_matured_live_for_profile_view(cfg, &profile, group)?;
         reject_permissionless_resolve_matured_live_view(cfg, group)?;
         if group.header.bankruptcy_hlock_active != 0
             || group.header.threshold_stress_active != 0
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4080,6 +4080,17 @@ pub mod oracle_v16 {
 pub mod policy_v16 {
     use crate::constants::MAX_DYNAMIC_TRADE_FEE_BPS;
 
+    pub fn price_managed_profile_resolve_matured(
+        price_managed: bool,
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing under the strict impact gate. The red case moves authority/provider-controlled balances but does not demonstrate user LoF or a blocked user/keeper continuation; one earnings setup also installs fixture state directly.

PR #197, opened 2026-07-09
[codex] Reject local-stale no-observation liquidation

aeyakovenko:
## Root cause

`PermissionlessCrank` required a selected-asset observation only while a price-managed asset still had pending effective-price movement. A non-base AuthMark/EWMA profile could therefore be locally resolve-matured but fully applied, allowing a public no-observation liquidation against stale selected-asset data while the base market remained fresh.

## Impact and red

The public LiteSVM lifecycle creates a permissionless non-base AuthMark asset, opens real OI, makes the short liquidatable, refreshes only the base profile, and matures only asset 1. With the guard manually removed on current `main`, the no-observation crank succeeds at `253,494` CU and reduces asset-1 short OI before the regression fails. This is a state-changing stale-price liquidation, not an injected state or no-op.

## Fix

For a price-managed selected asset, reject a missing observation once that profile reaches its configured permissionless-resolve boundary. Fresh profiles and non-price-managed profiles retain their prior liveness behavior. The wrapper still delegates liquidation sizing to the engine.

## PDD coverage

`v16_attack_non_base_local_stale_no_observation_liquidation_rejects` proves through public instructions that:

- asset 1 is locally stale while the base market is fresh
- no-observation liquidation rejects as `OracleStale` with market and portfolio bytes unchanged
- a public selected-asset refresh restores bounded liquidation progress

`kani_v16_missing_selected_observation_resolve_boundary_is_exact_and_live` calls the production policy kernel over full-width symbolic slots. It proves exact-boundary and post-boundary rejection, fresh-profile liveness, regressed-slot safety, and unmanaged/disabled-profile liveness. All five covers are reached in 0.22 seconds.

## Verification

- `cargo test --all-targets`: 569 passed (5 unit + 564 LiteSVM)
- new Kani harness: successful, 5/5 covers reached
- `cargo clippy --all-targets` (passes with existing baseline warnings)

Head: `3805e83291fce156b10662a4da01b81e1c248c91`.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4080,6 +4080,17 @@ pub mod oracle_v16 {
 pub mod policy_v16 {
     use crate::constants::MAX_DYNAMIC_TRADE_FEE_BPS;
 
+    pub fn missing_selected_observation_is_resolve_matured(
+        price_managed: bool,
+        resolve_stale_slots: u64,
+        last_good_oracle_slot: u64,
+        now_slot: u64,
+    ) -> bool {
+        price_managed
+            && resolve_stale_slots != 0
+            && now_slot.saturating_sub(last_good_oracle_slot) >= resolve_stale_slots
+    }
+
     pub fn price_move_bps_ceil(old: u64, new: u64) -> Option<u64> {
         if old == 0 || old == new {
             return Some(0);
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing under the strict impact gate. The probe demonstrates a stale-policy state transition, but not incorrect user value loss or an absent bounded continuation; a caller can supply the selected observation.

PR #199, opened 2026-07-10
[HARDENING] Unblock zero-OI recovery leg cleanup

aeyakovenko:
## Finding

A public full liquidation can reduce both sides' effective OI to zero while leaving the profitable opposite account's raw leg stored on a `ResetPending` side. If that asset is then shut down, the absent losing owner leaves no paired leg for `ForceCloseAbandonedAsset`, ordinary account refresh/B cranks cannot detach the stored leg, and the owner-only `ForfeitRecoveryLeg` prevents side reset and asset restart indefinitely.

The rebased LiteSVM PDD reaches this state exclusively through public trade, authenticated mark movement, engine-selected liquidation, shutdown, and bounded permissionless crank transactions. With the parent owner-only handler, the final keeper detach fails exactly with `Custom(8)` (`Unauthorized`) after all refresh and B work is complete.

## Fix

After the authenticated force-close delay, permit any signer to detach only a Recovery leg whose side is `ResetPending` with zero effective OI. The wrapper delegates refresh and detach to the existing zero-copy engine APIs. A non-owner transaction commits only if the engine reports a detached leg and zero forfeited PnL, settled loss, support, junior burn, principal, insurance, residual, or explicit-loss flow. Every rejected transaction is atomically rolled back by SVM.

Owner-authorized forfeits retain their existing behavior. Pre-delay calls, cross-market substitutions, resolve-matured calls, and Recovery positions with nonzero effective OI remain rejected.

## PDD coverage

- Rebased to current `main` and current engine-selected liquidation/crank API; no keeper-controlled `close_q` remains.
- The public setup proves non-vacuous Recovery lifecycle, zero effective OI, `ResetPending`, shutdown slot, configured delay, authenticated clock, and multi-transaction B progress.
- Parent handler: exact regression fails at the final keeper call with `Unauthorized` after all prerequisite progress succeeds.
- Fixed handler: exact regression passes under the custody CU guard, leaves account capital and PnL unchanged, removes the stored leg, and finalizes both side resets.
- A live-effective Recovery position and every pre-delay keeper attempt reject without mutation.

## Validation

- `cargo test --test v16_cu v16_attack_delayed_permissionless_forfeit_clears_zero_oi_orphan -- --nocapture`: passed
- parent-code red run of the same test: failed at final detach with `Custom(8)` as expected
- `cargo test --test v16_cu forfeit_recovery_leg -- --nocapture`: 3 passed
- `cargo test --test v16_cu v16_attack_recovery_tools_owner_gated -- --nocapture`: passed
- `cargo clippy --all-targets`: passed with existing warnings only

Head: `6156a46f7255974ebad58a8c87b84bdf41984612`

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,6 +472,7 @@ This section describes intent and operational ordering, not argument-by-argument
   - owner-signed close recovery path; optional deposit is transferred first, then the engine cancels the pending close if the cure succeeds
 - **ForfeitRecoveryLeg** (tag 43)
   - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget
+  - after the force-close delay, any signer may refresh and detach a zero-effective-OI, reset-pending Recovery leg only when the engine reports no forfeited PnL or loss flow
 - **RebalanceReduce** (tag 44)
   - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector
 - **ClaimResolvedPayoutTopup** (tag 46)
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8385,14 +8385,94 @@ pub mod processor {
         if b_delta_budget == 0 {
             return Err(PercolatorError::InvalidInstruction.into());
         }
-        with_one_portfolio_view(program_id, accounts, true, |group, portfolio, cfg| {
-            if group.header.mode == 0 && permissionless_resolve_matured_now_view(cfg, group) {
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Superseded by PR135 and consolidated engine PR195. The current auto-crank/Recovery continuation and public liveness composition cover zero-effective-OI cleanup without this separate wrapper branch.

PR #98, opened 2026-05-14
Persist hybrid mark fields across both partial-restore writebacks

dcccrypto:
Closes #97.

## Problem

Commit `40d9024` ("Keep hybrid mark aligned during oracle staircase") taught `read_price_and_stamp` to snap three trade-derived mark fields to the freshly-accepted external price on every successful read in hybrid-after-hours markets (`src/percolator.rs:4276-4290`):

- `hyperp_mark_e6`
- `mark_ewma_e6`
- `mark_ewma_last_slot`

Both partial-restore writeback paths silently dropped those mutations:

1. **Partial-catchup KeeperCrank writeback** through `policy::PartialCrankConfigFields` whitelist + apply helper at `src/percolator.rs:760-790` and `:5025-5052`. The whitelist had 8 fields; none of the three mark fields were present.
2. **TradeCpi zero-fill rollback's** open-coded six-field forwarding at `src/percolator.rs:8421-8428`. None of the three mark fields were forwarded either.

After either writeback, on-disk state ended up with `mark_ewma_e6` pinned at the pre-call value while `last_effective_price_e6` (whitelisted) advanced to the new external price. `compute_current_funding_rate_e9` (`src/percolator.rs:5410-5419`) reads both fields; the fabricated divergence saturated `funding_rate_e9_from_mark_index` at the engine cap `funding_max_e9_per_slot`, transferring funding from the loss-side counterparty to the favored side every accrual interval.

This was the invariant `40d9024` exists to prevent — the inline comment at `:4279-4286` names the exact attack class ("regular-hours duplicate cranks can create artificial mark/index funding divergence before after-hours fallback has actually started"). The in-memory snap was correct; the writebacks just dropped it on the way to disk.

Full root-cause analysis and reproducer in #97.

## Cause

The `40d9024` commit added the three field mutations in `read_price_and_stamp` but did not extend either of the two partial-restore writebacks to cover them. Same omission shape as `d40edb5` (the TOTO commit) which added `oracle_leg_publish_times` / `oracle_leg_prices_e6` without extending the same writebacks — fixed by PR #94 (#93's site) and PR #96 (#95's site).

This is a strict superset of those fixes: same two writeback sites, additional three fields. Each writeback site needs its own field extension; PR #94 and PR #96 do not cover this finding.

## Fix

Extend both writeback sites to forward the three mark fields from `after_read_and_sweep`:

| Symbol | Edit |
|---|---|
[...]

PR #200, opened 2026-07-10
Fix mixed-lifecycle auto-crank selection

aeyakovenko:
## Summary

- Pin the wrapper to the mixed-lifecycle selector fix from engine PR #98 at `35f6ffec9c52e6237a356af724a04ad7f546e744`.
- Add a public LiteSVM regression for an account whose first leg is in `Recovery` while a later live leg is liquidatable.
- Prove permissionless crank skips the non-dispatchable recovery leg, refreshes the live leg, then liquidates it without mutating the recovery obligation.

Engine PR: https://github.com/aeyakovenko/percolator/pull/98

## Public repro

The old selector chose the first active portfolio leg without requiring that its asset lifecycle could accept the selected action. A normally constructed cross-margin account with a recovery leg before an unhealthy live leg therefore retried an action that always returned `EngineLockActive`; one honest cranker could not reach the later liquidation.

The engine now filters ordinary refresh/liquidation candidates to dispatchable lifecycles while retaining recovery legs for their valid settlement work.

## Validation

- rebased directly onto wrapper `main` at `da64be639168b496833dd4c701607a94fcb06b6c`
- aggregate through wrapper #266: 6 unit + 568 serialized LiteSVM/CU tests passed
- all existing full-14-leg and 10 MiB CU tests pass
- `git diff --check`

Head: `3b9c8c293b601b2f3f64f183c423ad7cde6cd105`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=f212d98919d30f7b15e6872793403a46687674f3#f212d98919d30f7b15e6872793403a46687674f3"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by PR135 and consolidated engine PR195. PR195 includes the mixed-lifecycle selector implementation and proof; PR135 carries the public route and rank-decrease coverage.

PR #203, opened 2026-07-10
[BLOCKER DoS] Force-close abandoned ADL Recovery residues

aeyakovenko:
## What changed

- Extend the delayed `ForceCloseAbandonedAsset` route to settle a singleton zero-effective-OI residue after its opposite leg has already been closed.
- Pin engine PR aeyakovenko/percolator#101, which migrates that legacy residue into the reset epoch before recovery forfeit clears it.
- Add a public LiteSVM regression covering deposits, trade, partial ADL, shutdown, two unsigned force-close calls, side finalization, CU bounds, custody conservation, and preservation of the user's independent PnL claim.

## Root cause

Partial ADL can leave a winner's stored basis larger than matched effective OI. After asset shutdown, the first paired force close consumed the final matched lot and cleared the short, but left a nonzero long residue with both OI sides at zero. Every later paired force close failed because no opposite leg remained, while recovery forfeit tried to subtract the stored basis from zero OI.

With the wrapper route held fixed and the engine pinned to parent `f87ca00`, the public regression fails on the second unsigned call at 97,660 CU with `Custom(25)` (`0x19`); the transaction rolls back and leaves the residue stuck.

## Validation

- `cargo test --all-targets`: 563 LiteSVM tests and 5 wrapper unit tests pass.
- New path remains bounded: matched force close under the trade CU limit; singleton cleanup and side finalization under the custody CU limit.
- Engine Kani reset+clear composition theorem: 649/649 checks passed, 6/6 covers satisfied, 4.46s.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=ca7f10c2c0e1cd93b3adb25cdd81973c26974cdf#ca7f10c2c0e1cd93b3adb25cdd81973c26974cdf"
+source = "git+https://github.com/aeyakovenko/percolator?rev=a64e33aa1b61e749e42a6fec749b52482ca72db7#a64e33aa1b61e749e42a6fec749b52482ca72db7"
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #205, opened 2026-07-10
[BLOCKER DoS] Bound resolved source-domain close publicly

aeyakovenko:
## What changed

- Pin engine PR #104 at `eeb7db26c1c7b231f6ced442dd39a93a2e175ae8`.
- Construct 28 source domains through public 14-asset trading, covering fresh and lapsed backing at resolution.
- Require strict rank decrease on every successful nonterminal close, no token movement during progress, exact final vault conservation, and a 1,000,000-CU per-call ceiling.
- Keep direct `CloseResolved` payout-account validation strict while allowing resolved permissionless crank to make zero-payout progress with its minimal account set.

## Reproduced failure

On the parent engine, both public scenarios consume exactly 1,400,000 CU, abort, and roll back. Retrying cannot remove a source domain, so the flat resolved winner cannot collect or close.

With the fix, 28 fresh domains converge in 29 calls and 28 lapsed domains converge in 57 calls, with every continuation below 1,000,000 CU.

## PDD validation

- `v16_attack_resolved_close_28_fresh_source_domains_makes_bounded_progress` passes.
- `v16_attack_resolved_close_28_lapsed_source_domains_makes_bounded_progress` passes.
- `cargo build-sbf --tools-version v1.52` passes.
- `cargo test --all-targets --quiet`: 5/5 library and 566/566 LiteSVM tests pass.
- Engine conservation and terminal-lien proofs are documented in https://github.com/aeyakovenko/percolator/pull/104.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=22df6aeaa32f06c3e0e38e7a75f7de2b311eaac4#22df6aeaa32f06c3e0e38e7a75f7de2b311eaac4"
+source = "git+https://github.com/aeyakovenko/percolator?rev=5997d84eb00c21cb28c49f05e5e2786f5ee2e54e#5997d84eb00c21cb28c49f05e5e2786f5ee2e54e"
 dependencies = [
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by #212, which contains this public resolved source-domain repro plus the max-leg and live max-source reproductions against one engine fix, rebased directly on main.

PR #208, opened 2026-07-10
[BLOCKER DoS] Bound max-leg resolved source teardown

aeyakovenko:
## What changed

- Pin the engine to the bounded leg-teardown fix at `c78f7178bad825dba6aa6a70435a5ed2968a6b94`.
- Extend the public 28-source-domain fixture to preserve all 14 winner legs through resolution.
- Wind down all 14 losing counterparties through public instructions, leaving the winner as the last max-shape active account.
- Exercise both direct `CloseResolved` and resolved `PermissionlessCrank`.
- Require every successful nonterminal transaction to reduce the exact combined rank: active legs + source claims + lapsed buckets.
- Preserve token balances during progress and verify exact final payout/vault conservation.
- Update the existing two-leg deep-insolvency test for one-leg-per-crank semantics.

## Reproduced failure

On the parent engine, both public routes consume exactly `1,400,000 CU`, abort, and roll back. Retrying cannot remove a leg or source claim, so the resolved winner cannot collect funds or close.

With the fix:

- Direct close peaks at approximately `1.16M CU`.
- Resolved auto-crank peaks at approximately `1.23M CU`.
- Both remain below the `1.30M` max-shape guard and converge with exact rank decrease.

The state is built entirely through public deposits, 14-asset trades, mark updates, cranks, counterparty closes, and resolution. No account bytes are injected.

## Verification

- `CARGO_INCREMENTAL=0 cargo test --all-targets`: 5 unit tests and 568 LiteSVM/CU tests pass.
- Both prior 28-domain fresh/lapsed regressions pass.
- The malformed-hint deep-insolvency wind-down regression passes under the new bounded semantics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=5997d84eb00c21cb28c49f05e5e2786f5ee2e54e#5997d84eb00c21cb28c49f05e5e2786f5ee2e54e"
+source = "git+https://github.com/aeyakovenko/percolator?rev=a154433e1bdb08a50b9fd969eb4dba938a35764a#a154433e1bdb08a50b9fd969eb4dba938a35764a"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by #212, which contains this max-leg terminal repro together with the source-domain and live max-source paths, rebased directly on main and fully retested.

Issue #110, opened 2026-06-02
CRITICAL: 100,000,000x Price Manipulation (Retire Flip) + 1000x Notional Understatement


zeyxx:
# Percolator V16 Security Audit — Bounty-6 Submission

**Status**: FINAL / CRITICAL
**Date**: 2026-06-02
**Target**: Percolator V16 Program (`Bu1J8eQQN2mNnUgisSEd5StBG6zDaRb7fwDjN34VzgLG`)
**Auditor**: CYNIC Organism
**percolator-prog** @ `a0faea40` | **percolator engine** @ `20b7543d`

---

## Executive Summary

This audit has identified four high-impact vulnerabilities in the Percolator V16 engine. The most critical flaw (**Finding N**) causes a catastrophic price mismatch during asset retirement through an `effective_price` / `invert`-flag inconsistency, enabling instantaneous insurance fund drainage. **Finding P** (Notional Understatement) allows permissionless profit extraction via a 1000x decimal mismatch in wSOL-collateral inverted markets.

| Finding | Title | Severity | Impact |
| :--- | :--- | :--- | :--- |
| **N** | Retire Flip: `effective_price` stored with `invert=0` | **CRITICAL** | Protocol Insolvency / Fund Drainage |
| **P** | Inverted Market Notional Understatement (1000x) | **HIGH** | Permissionless Profit Extraction |
| **O** | Insurance Ledger Scope Confusion | **HIGH** | Principal Extraction as Yield |
| **Q** | Loss-Stale Isolation Bypass | **MEDIUM** | Front-running Socialization |

---

### Root Cause

**File**: `src/v16_program.rs` (`percolator-prog @ a0faea40`)

Two retirement paths write the engine's `effective_price` into the oracle profile, both with `invert` cleared to `0`.

**DRAIN_ONLY → RECOVERY transition** (lines 8800–8816):
```rust
let frozen_mark = group.markets[asset_index].engine.asset.effective_price.get(); // 8800-8804
// ...
profile.mark_ewma_e6 = frozen_mark;           // line 8814
profile.oracle_target_price_e6 = frozen_mark; // line 8816
```

**ASSET_ACTION_RETIRE** (lines 8980–8992), via `manual_asset_oracle_profile` + `preserve_backing_fee_policy`:
```rust
let price = group.markets[asset_index].engine.asset.effective_price.get(); // 8980-8984
let profile = preserve_backing_fee_policy(
[...]

PR #212, opened 2026-07-10
[BLOCKER DoS] Prove bounded max-source account progress

aeyakovenko:
## Public reproducer

This PR constructs a live, well-initialized 14-asset market and two 14-leg portfolios through public SBF instructions only. The victim has all 28 sparse source entries occupied. No account-state injection is used.

Against the pre-fix engine, both the honest live progress route and resolved `CloseResolved` reached exactly 1,400,000 CU and rolled back, leaving no bounded owner or keeper continuation. That is a persistent funds-locking DoS.

The RED-first extension also grows a normal market to asset 16, creates a real backed positive-PnL claim in global source domain 32, and proves the winner can close and receive its backed profit. It caught and now prevents global-domain truncation in the pending progress representation.

## Coverage

The regression requires:

- each successful resolved continuation strictly decreases a leg/source/finalization rank
- every resolved step remains at or below 1,375,000 CU
- live stale-leg settlement and refresh complete in finite calls
- liquidation remains executable
- asset-16/global-domain-32 backed profit remains closeable
- final SPL vault custody equals engine vault accounting
- incremental terminal payout tests still validate canonical vault binding, cross-rail replay resistance, and cross-margin insolvency wind-down at the actual payout boundary

Measured with the exact rebuilt SBF:

- stale-leg settlement / refresh max: 1,324,675 CU
- liquidation: 939,113 CU
- resolved close step max: 1,163,297 CU
- both max-shape accounts close in 58 calls

## Verification

- engine pin: `ca9410de3fe79165375583d37f5a83c7b79a313d`
- `cargo build-sbf --tools-version v1.52`: passed
- wrapper runtime: 6/6 unit + 567/567 LiteSVM/CU tests passed
- wrapper-specific instruction decoder Kani harness: 0/864 failed, cover reached
- engine runtime: 141/141 passed
- engine terminal source-entry rank Kani harness: 0/8 failed, cover reached

Engine fix and rank proof: aeyakovenko/percolator#110

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=995abeac75804c976ebcabd2a8992bc67d026736#995abeac75804c976ebcabd2a8992bc67d026736"
+source = "git+https://github.com/aeyakovenko/percolator?rev=ca9410de3fe79165375583d37f5a83c7b79a313d#ca9410de3fe79165375583d37f5a83c7b79a313d"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Added the exact 14-active-leg / 32-source-domain public boundary in 5ecac49f. On the pre-fix pinned engine 143e68c, refresh, owner rebalance, and fee sync each hit 1,400,000 CU while bilateral exit rejects stale. On PR #110 engine ca9410de, the same state reaches a current certificate in 2 finite calls (max 1,365,558 CU) and liquidates at 804,825 CU with custody unchanged.

aeyakovenko:

Issue #113, opened 2026-06-07
Permissionless cross-asset maintenance-fee siphon: a parasitic appended asset captures 1/N of every maintenance fee


dcccrypto:
# Permissionless cross-asset maintenance-fee siphon: a parasitic appended asset captures 1/N of every maintenance fee

## Summary

`credit_maintenance_fee_to_active_market_budgets_view` splits every charged maintenance fee **equally across all ACTIVE assets'** insurance domains (`base_share = amount / active_count`), with no requirement that an asset have any positions or activity. A non-admin caller can permissionlessly append a do-nothing asset, install itself as that asset's `insurance_operator`, and then withdraw its share — siphoning `1/N` of every other account's maintenance fee (50% with one appended asset; `k/(k+1)` with `k`).

## Target

| | |
|---|---|
| Program | `4m3ipBQDYX6JQ9YSmUXDjESDHMtGWtiXforkWr9Qoxdi` |
| Deployed BPF (ELF sha256) | `cbcd8b8d9d37eca60626c3aa13cee85c1a6e39e9699858a4af044e1e0f1f6024` |
| wrapper | `aeyakovenko/percolator-prog @ 0f87dcb` |
| engine | `aeyakovenko/percolator @ 58dc118` |
| Market | `4AXbMuJzrUv5KtVs6zc5jDtXTB4XKhKtEBGU6BmkVut4` |

## Mechanism

1. A maintenance fee charged from an account's capital is routed by `credit_maintenance_fee_to_active_market_budgets_view` (`src/v16_program.rs:5298`): `base_share = amount / active_count` (`:5320`), credited to **every** asset whose `lifecycle == ASSET_LIFECYCLE_ACTIVE` (`:5311`). There is no positions/activity check.
2. For asset `N != 0`, `credit_market_fee_split_across_domains_view` (`:5277`) redirects only `fee_redirect_to_market_0_bps` of the share back to market-0. On this market that value is **0**, so the appended asset keeps its full share.
3. A non-admin appends asset `N == max_market_slots` via `UpdateAssetLifecycle(ACTIVATE)` (`handle_update_asset_lifecycle:9077`), paying `permissionless_market_init_fee` and supplying its own `insurance_operator`.
4. The attacker withdraws that asset's domain budget via `WithdrawInsuranceAsset` (`handle_withdraw_insurance_asset:8200`), authorized by `live_authority_matches(insurance_operator, …)`.

[...]

PR #215, opened 2026-07-11
Bound max-source lien trading CU

aeyakovenko:
A normal public lifecycle retains 20 source-claim domains across ten active assets. The first risk-increasing TradeNoCpi creates one backed source lien, but a one-position follow-up requiring a second lien consumed the full 1.4M CU budget and aborted. Splitting the requested size does not cross the second-lien boundary.\n\nPin engine PR #113, whose allocator removes redundant per-domain full-state validation while retaining the allocator-entry validation and the trade's committed-state audit. The same public call now uses 898,270 CU.\n\nTDD is entirely public LiteSVM setup: InitMarket/asset activation, backing topups, portfolios, deposits, trades, authenticated mark cranks, then the two risk increases. No state injection.\n\nVerification:\n- cargo build-sbf --tools-version v1.52\n- cargo test --all-targets: 5/5 + 570/570\n- engine cargo test --all-targets: 132/132\n- targeted source-lien Kani: 121 checks, 2/2 covers\n\nStacked on PR #212 because its max-source refresh work makes this 20-domain public lifecycle reachable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=f87ca00a95b285aef5579639132991697fcbe7c5#f87ca00a95b285aef5579639132991697fcbe7c5"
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Closed during consolidation. This CU cliff blocked adding a second risk position rather than an existing funded exit; later source-domain scan reductions and maximum-shape required-exit coverage are consolidated in engine PR195 and wrapper PR135.

PR #202, opened 2026-07-10
[REAL DoS] Keep partially ADL-reduced users exit-live

aeyakovenko:
## Public TDD

The test reaches the failure using only public LiteSVM instructions:

1. deposit two users;
2. open a balanced two-lot trade;
3. authenticate a 5x mark move;
4. partially liquidate the loser, ADL-reducing matched OI to one lot while the winner retains two lots of stored basis;
5. submit the normal max-work owner `RebalanceReduce`;
6. unsigned-crank both terminal residues and finalize both reset sides.

At parent engine `a9baa08`, step 5 fails at 121,431 program CU with wrapper `Custom(25)` / `EngineCounterUnderflow`. With engine PR #100, work clamps to matched effective OI, both residues detach permissionlessly in bounded calls, both sides finalize, and real SPL vault custody remains conserved.

## Scope

- Engine pin to `6d1d9d7f653dec327e9b95941933afcb8b249432`.
- One net-new public LiteSVM regression.
- No wrapper program-code change.

## Verification

- `cargo build-sbf --tools-version v1.52`
- `cargo test --all-targets`: 5 unit + 562 LiteSVM tests green.
- New path explicitly checks CU bounds for reduction, cleanup, and both finalizations.
- Clippy still reports the branch's existing warning baseline; this change adds no new clippy finding.

PR #216, opened 2026-07-11
[BLOCKER DoS] Allow ADL winners to close after resolution

aeyakovenko:
## What changed

- Pin strengthened engine PR #115 (`f2019ef76d06b28f40136f4d099e644e3b131f10`).
- Add a public LiteSVM lifecycle theorem for an ADL-scaled winner whose stored basis exceeds remaining effective OI.

## Public red/green

The test uses only ordinary wrapper instructions: two deposits, a trade, authenticated mark movement, permissionless partial liquidation, market resolution, and public resolved closes. Against the unfixed engine, the first winner close deterministically returns `Custom(25)` after about 128k CU and rolls back forever.

With PR #115, the winner settlement record detaches, the loser and winner receive exactly their terminal amounts, both portfolios dematerialize, engine vault and `c_tot` reach zero, and SPL custody reconciles exactly. Every close remains under the custody CU guard.

## Net-new coverage

PR #202 still fails this reproduction on its exact engine pin. That pair proves live partial ADL exit and zero-OI residue cleanup; this PR proves the distinct resolved payout path when effective OI is nonzero but below stored basis.

## Validation

- `v16_attack_resolved_adl_winner_can_wind_down_publicly`: passed
- `cargo test --all-targets --no-default-features`: 5 unit + 564 LiteSVM tests passed
- Engine Kani ADL-gap theorem: 157 checks, 3/3 covers, 2.33s
- Engine full-frame contract: 1,290 checks, 9.9s
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=7f60d4fdde82774f398903917278807d6b0fa99e#7f60d4fdde82774f398903917278807d6b0fa99e"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-061 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #116, opened 2026-06-09
Add test: Pyth exponent bound + extreme-scale arithmetic enforced

aeyakovenko:
## What

Adds one litesvm attack regression: `v16_attack_oracle_pyth_exponent_bounds_enforced`.

## Attack analyzed (DoS via overflow-panic / LoF via mispricing — SOL-001/SOL-014)

`read_pyth_price_e6` scales a Pyth price by `10^(exponent + 6)`:
- `scale >= 0`: `price_u.checked_mul(10u128.pow(scale))`
- `scale < 0`: `price_u / 10u128.pow(-scale)`

The `|exponent| <= MAX_EXPO_ABS (18)` guard at `v16_program.rs:4091` is the *only* thing preventing a feed that reports an extreme exponent from driving `10u128.pow(huge)` past `u128::MAX`. Without it:
- a positive overshoot **panics** (`pow`/mul overflow) — and since the read happens inside every crank/refresh/trade that touches the asset, the feed becomes a permanent **liveness DoS**; or
- the mark is scaled to a wildly wrong value → **mispriced positions / LoF**.

## Plausibility vs. code — defended

The exponent is rejected *before* any `pow()` runs. With `MAX_EXPO_ABS = 18`, `scale ∈ [-12, 24]`, so `10^24` fits comfortably in `u128` (~1e24 ≪ 3.4e38) and the multiply is `checked_mul`; the post-scale `out == 0 || out > MAX_ORACLE_PRICE` check catches in-bounds extremes.

## Test

A `configure(price, expo)` closure plants a real Pyth feed and tries to configure a hybrid oracle:
- sanity: `expo=-6` configures;
- `expo = +19` / `-19` (out of bounds) → reject `Custom(26)` (OracleInvalid), gracefully, never a panic;
- `expo = +18`, large mantissa → scaled past `MAX_ORACLE_PRICE` → reject without overflow;
- `expo = -18`, small mantissa → floors to 0 → reject (no zero mark).

Real BPF via litesvm. **Passes.**

## Why it adds coverage

No existing test varies the Pyth exponent: `oracle_negative_or_zero_price_rejected` only varies *price* at a fixed `expo=-6`, and `oracle_composite_over_max_price_rejects` exercises the multi-leg *compose* path — not single-leg exponent scaling.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #217, opened 2026-07-11
[BLOCKER DoS] Keep social-loss dust from blocking owner exit

aeyakovenko:
## Summary

- Pin the wrapper to engine PR #116 at `4beb6ce141bb28656e7cc38b82a3d5c54e1884e4`.
- Add a public LiteSVM regression for social-loss rounding carry permanently blocking an ordinary owner exit.
- Prove the carried atom is charged exactly once, outstanding certificates are invalidated, custody is unchanged, and the remaining winner settles the atom exactly.

## Public red case

Eight normal users deposit, four matched positions trade, authenticated mark moves create a two-atom bankruptcy, permissionless cranks settle social loss, and two winners reduce to flat. Each winner has `SOCIAL_LOSS_DEN / 2` remainder. On the parent engine, the second exit returns `Custom(23)` at about 114k CU forever; SVM atomic rollback preserves the same live state on every retry.

With engine PR #116, both owner exits commit within the wrapper CU bound. The test checks exact one-atom B reassignment, epoch invalidation, unchanged vault and insurance custody, and senior-value conservation.

The companion engine proof executes production `kernel_clear_leg` for symbolic side and all canonical dust/remainder values. Restoring the vulnerable reject-on-carry branch makes that proof fail.

## Validation

- `cargo test --all-targets`: 5 unit + 564 LiteSVM/SBF tests passed
- engine `cargo test --all-targets`: 130 passed
- three production-kernel Kani harnesses passed, including carry/no-carry covers
- `git diff --check`

Engine: https://github.com/aeyakovenko/percolator/pull/116

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=0a7defd04e87f5788ed0f2b543548eeb5592c32a#0a7defd04e87f5788ed0f2b543548eeb5592c32a"
+source = "git+https://github.com/aeyakovenko/percolator?rev=4beb6ce141bb28656e7cc38b82a3d5c54e1884e4#4beb6ce141bb28656e7cc38b82a3d5c54e1884e4"
 dependencies = [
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #218, opened 2026-07-11
[BLOCKER DoS] Keep reset carry from blocking liquidation

aeyakovenko:
## Summary

- Pin the wrapper stack to engine PR #117 at `552e9ac76a2288fb5a91c40e1cce7e952a6ab4a5`.
- Add a public LiteSVM regression for reset-boundary rounding carry permanently blocking liquidation.
- Verify bounded progress, exact capital consumption, durable audit attribution, and unchanged SPL custody.

## Public red case

Eight users deposit and trade through normal instructions. Authenticated mark updates and permissionless cranks create one bankruptcy, a winner exit creates canonical side dust, and a second bankruptcy drains effective OI to zero. On the parent engine, every liquidation retry returns `Custom(23)` at about 188k CU; SVM rollback preserves the same live underwater account.

With engine PR #117, liquidation commits at about 206k CU. The test proves the target is flat with zero capital/PnL, both effective OI values are zero, dust is canonical, exactly one explicit unallocated atom is recorded, B/weight reset, vault and insurance custody do not move, `c_tot` decreases exactly, and senior value is conserved.

The companion engine contract covers all canonical dust/remainder pairs and is mutation-sensitive to restoration of the reject-on-carry path.

## Validation

- `cargo build-sbf --no-default-features`
- auth and hostile matcher SBF fixtures built
- `cargo test --all-targets`: 5 unit + 565 LiteSVM/SBF tests passed
- engine `cargo test --all-targets`: 130 passed
- Kani reset-carry production contract: 355 checks, 2/2 covers, passed
- `cargo fmt --all -- --check`
- `git diff --check`

## Stack

Base: wrapper PR #217
Engine: https://github.com/aeyakovenko/percolator/pull/117

PR #219, opened 2026-07-11
[BLOCKER DoS] Reset both zero-OI sides to preserve exit

aeyakovenko:
## Summary

- Pin the stacked wrapper PDD to engine PR #118 at `f97194677901e38bc8b2b87a67ac878c74dbfa99`.
- Extend the public reset-carry lifecycle through the surviving liquidated-side user's complete exit.
- Verify dead-leg forfeit, reset finalization, CU bounds, and complete SPL capital withdrawal.

## Public red case

After the public two-bankruptcy sequence from PR #218, both effective-OI sides are zero but one short account remains stored. On the parent engine, the short side remains `Normal`, account-local auto-crank cannot detach the leg, and owner `ForfeitRecoveryLeg` repeatedly returns `Custom(21)` at about 24k CU. The user remains non-flat and cannot withdraw.

With engine PR #118, the short side enters `ResetPending`, owner forfeit succeeds at about 77k CU, public reset finalization succeeds, and the account withdraws all remaining SPL capital. Every call stays below the custody CU limit.

The engine companion theorem executes the actual zero-copy unilateral-close mutation for symbolic long/short orientation. Disabling only the fixed production transition makes that theorem fail.

## Validation

- focused public lifecycle passed
- `cargo test --all-targets`: 5 unit + 565 LiteSVM/SBF tests passed
- engine production composition theorem: 4,794 checks, 2/2 orientation covers, passed in 143s
- engine scalar contract: 105 checks, 2/2 covers, passed
- `git diff --check`

## Stack

Base: wrapper PR #218
Engine: https://github.com/aeyakovenko/percolator/pull/118

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=552e9ac76a2288fb5a91c40e1cce7e952a6ab4a5#552e9ac76a2288fb5a91c40e1cce7e952a6ab4a5"
+source = "git+https://github.com/aeyakovenko/percolator?rev=f97194677901e38bc8b2b87a67ac878c74dbfa99#f97194677901e38bc8b2b87a67ac878c74dbfa99"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "552e9ac76a2288fb5a91c40e1cce7e952a6ab4a5" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "f97194677901e38bc8b2b87a67ac878c74dbfa99" }
 anchor-lang-v2 = { git = "https://github.com/solana-foundation/anchor", rev = "e3cf760826fa0a60f247635ea0572e59861ec9b5", package = "anchor-lang-v2", default-features = false, features = ["alloc", "guardrails"], optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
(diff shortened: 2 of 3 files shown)


aeyakovenko:
A distinct public bilateral-close variant remains reachable after this fix.

Reproduction: `aeyakovenko/percolator-meta@9c75c9d` on branch `agent/real-dos-bilateral-zero-oi-terminal`, test `e2e_social_loss_then_bilateral_close_has_a_bounded_terminal_exit`.

Public sequence:
1. A solvent long opens 20 against a solvent short, then 1 against a small funded short.
2. An honest mark bankrupts the small short and social-loss settlement reduces long effective OI from 21 to 20 while the solvent winner retains basis 21.
3. The solvent pair legitimately closes the remaining matched 20 bilaterally.
4. Both effective OI counters become zero, but `stored_pos_count_long == 1`; the winner retains normal-epoch basis 1 and `mode_long == Normal`.

The isolated LiteSVM probe exhausts the bounded continuations: auto-crank makes no rank progress; `FinalizeResetSide` is a successful byte-no-op; crossing trade is H-locked; `RebalanceReduce` rolls back with counter underflow; drain-only does not change that; recovery `ForfeitRecoveryLeg` underflows; force-close has no opposite stored leg; and resolved `CloseResolved` underflows. The winner still has all 100,000 deposited atoms and remains the sole `resolved_payout_blocker_count`.

I also ran the same probe against the real SBF built from the program counterpart #219 (`b0975e5`, SHA-256 `d910e21249d04310bd01ea545544d6dced561b2480e54aace2b8cf192192df0e`); it fails identically. The current fix runs only from the unilateral-close path, while this transition reaches zero OI in a later bilateral trade. Please extend the invariant repair to that transition and add this public regression.

aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_reset_carry_liquidation_matrix_preserves_progress`. The proposed bilateral-zero-OI state is downstream of PR218's fix; on the pinned program the second bankruptcy rejects exactly while both effective-OI counters remain nonzero. The executable disposition remains in PR #135; no separate production fix is required.

PR #220, opened 2026-07-12
[BLOCKER LoF] Prevent liquidation on an omitted rescue mark

aeyakovenko:
## What changed

- Check pending price movement for every active portfolio leg before stale-account refresh or liquidation.
- Keep B-settlement on committed state unchanged.
- Add a public LiteSVM TDD regression and update existing multi-asset controls to submit every mark they stage.

## Root cause and impact

The wrapper checked only the engine-selected first active leg before a no-observation auto-crank. A permissionless keeper could omit an already-authenticated pending mark on a later leg, certify the account against the old price, then liquidate it.

On unfixed `main`, the regression reaches only public instructions, reduces the victim position from `10,000,000` to `9,987,820`, and transfers `7` units to insurance. Supplying the same pending observation first leaves the account healthy, fully positioned, and fee-free.

The fix rejects the stale-price crank atomically. A bounded public continuation remains available by supplying the pending observation; trade paths are unchanged.

## Validation

- `cargo test --lib` (5 passed)
- built `auth_matcher` and `hostile_matcher` SBF fixtures
- `cargo test --test v16_cu` (564 passed)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10572,17 +10572,14 @@ pub mod processor {
             let summary = group
                 .build_actionable_summary(&portfolio.as_view())
                 .map_err(map_v16_error)?;
-            if let Some(asset_index) =
-                auto_crank_selected_asset_that_accrues_view(&portfolio, &summary)?
-            {
-                reject_missing_pending_selected_observation_view(
-                    &cfg,
-                    &group,
-                    asset_index,
-                    authenticated_now_slot,
-                    observations.as_slice(),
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-053 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #221, opened 2026-07-12
Reject stale matcher replay after market reinit

aeyakovenko:
## What changed

- Add a public LiteSVM lifecycle regression that closes a market and its portfolios, recreates the same account addresses, and re-enables the same external matcher context.
- Prevent single-fill `TradeCpi` from issuing the request ID already stored in that persistent context.
- Verify that an unchanged stale response rejects atomically and that a matcher which writes a fresh response remains live.

## Impact

Before the fix, closing and reinitializing a market reset the market-local matcher sequence to zero. An external matcher context could survive that lifecycle. If its old response had request ID 1 and the matcher returned `Ok` without writing on the new market's first request, the wrapper accepted the old response as fresh and opened positions at its stale execution price.

The red test reproduced this through public wrapper instructions; the stale second-incarnation fill returned `Ok` and consumed about 385k CU before the fix. Account provisioning in the test only models recreating accounts after public `ClosePortfolio`/`CloseSlab` zeroed them; no protocol state is forged.

The fix preserves the existing matcher ABI. A normal new context still starts at request ID 1. When a context already contains the next candidate ID, the market skips that one ID, so unchanged bytes cannot satisfy the existing echoed-request validation. A correctly invoked matcher writes the new ID and proceeds normally.

## Validation

- `cargo test --all-targets`: 5 unit and 564 LiteSVM tests passed
- matcher-focused LiteSVM sweep: 37 passed
- targeted Kani matcher-binding harness: 1 passed, 215 checks, 0 failures
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -1374,13 +1374,25 @@ pub mod state {
     }
 
     pub fn bump_matcher_req_seq(data: &mut [u8]) -> Result<u64, ProgramError> {
+        bump_matcher_req_seq_avoiding(data, None)
+    }
+
+    pub fn bump_matcher_req_seq_avoiding(
+        data: &mut [u8],
+        avoid_req_id: Option<u64>,
+    ) -> Result<u64, ProgramError> {
         check_header(data, KIND_MARKET)?;
         let mut config = read_wrapper_config_from_bytes(data)?;
-        config.matcher_req_seq = config
+        let mut req_id = config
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by PR135. The consolidated market, portfolio, matcher-context incarnation and one-shot capability controls reject the stale-context lifecycle generically, with public close/recreate evidence.

PR #223, opened 2026-07-12
[PRIVILEGED LoF] Require matcher consent for CPI backing fees

aeyakovenko:
## Summary

- reserve matcher-return flag bits for an LP-selected backing-fee cap (0-10,000 bps)
- reject only a nonzero unsigned-LP backing fee whose actual post-trade source-lien delta exceeds that cap
- keep legacy matchers fail-closed at cap zero while preserving fee-free and risk-reducing CPI fills
- add a fully public LiteSVM exploit/regression and extend the matcher-return Kani contract

## Security impact

`TradeCpi` is signed only by the taker. Backing-domain fees were calculated after the matcher returned, so no matcher could inspect or cap the fee charged to its unsigned LP.

The public red sequence creates ordinary cross-margin source PnL, then lets a permissionless asset authority install a 5,000 bps backing fee after the LP authorized its matcher. The attacker opens and reverses only a 20-unit incremental CPI leg at the same price. Its portfolio returns flat with unchanged capital, while the LP falls from 3,100 to 3,030 atoms and the attacker-controlled provider bucket rises from 0 to 70 atoms. The attacker then withdraws those 70 atoms through `WithdrawBackingBucketEarnings`. No protocol account bytes are injected.

The fix makes the matcher return an explicit maximum backing-fee policy. Existing matchers return zero in the reserved bits and therefore fail closed. Enforcement happens only after the engine has identified a positive source-backing delta and a nonzero fee; an error reverts the engine and matcher CPI under SVM semantics. The regression proves byte-for-byte rollback, explicit 5,000 bps opt-in, and a subsequent zero-cap risk reduction.

## Validation

- `cargo build-sbf` in `tests/fixtures/auth_matcher`
- `cargo build-sbf` in `tests/fixtures/hostile_matcher`
- `cargo test`: 5 unit + 564 LiteSVM tests passed, including all CU tests
- `cargo kani --tests --harness kani_v16_matcher_return_accepts_only_bound_echoed_fills`: 224/224 checks passed; accept-path cover satisfied
- `git diff --check`

The engine dependency and pin are unchanged.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -3412,6 +3412,8 @@ pub mod matcher_abi {
     pub const FLAG_VALID: u32 = 1;
     pub const FLAG_PARTIAL_OK: u32 = 2;
     pub const FLAG_REJECTED: u32 = 4;
+    pub const FLAG_BACKING_FEE_CAP_SHIFT: u32 = 8;
+    pub const FLAG_BACKING_FEE_CAP_MASK: u32 = 0x3fff << FLAG_BACKING_FEE_CAP_SHIFT;
 
     #[derive(Clone, Copy, Debug, PartialEq, Eq)]
     pub struct MatcherReturn {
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #224, opened 2026-07-12
[BLOCKER LoF] Prevent CPI caller-fee LP siphon

aeyakovenko:
## Security impact

`TradeCpi` and `BatchTradeCpi` let the taker choose `fee_bps`, but the LP owner does not sign either fill and the matcher ABI does not attest to that fee. A permissionless asset creator could therefore submit a 10,000 bps open-and-close round trip against an LP's existing matcher authorization, route the two-sided fees into creator-controlled insurance domains, withdraw them, and retain the LP's half as profit.

The public LiteSVM exploit exercises both CPI routes without protocol-state injection. Starting from 1,000,000 atoms each, the attacker and LP both fell to 980,000 while the attacker-controlled asset budget gained 40,000. The attacker then withdrew all 40,000, ending 20,000 atoms ahead exactly equal to the LP's loss.

## Fix

- Continue validating malformed caller fees against the configured maximum before invoking the matcher.
- Derive the fee actually charged by CPI fills from the market's configured base fee, not the unsigned taker's requested fee.
- Preserve deterministic hybrid/EWMA externality fees in the shared trade path.
- Leave `TradeNoCpi` caller-selected because both account owners sign that bilateral instruction.

## TDD evidence

- `ccc952a9` is the public red exploit for single and batch CPI.
- `e186d3af` fixes only CPI fee derivation.
- The regression proves both routes still execute and round-trip, no attacker-controlled fee reaches the LP, custody remains exact, and direct bilateral fee semantics remain unchanged.

## Validation

- authenticated and hostile matcher fixture SBF builds
- `cargo test --all-targets`: 5 unit tests and 564 LiteSVM/SBF tests passed, including all CU guards
- `git diff --check`

The full `cargo kani --tests` sweep is still running through the repository's pre-existing symbolic instruction-decoder harness; no affected runtime or CU test is pending.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -6613,7 +6613,10 @@ pub mod processor {
             asset_index,
             ret.exec_size,
             ret.exec_price_e6,
-            fee_bps,
+            // Only the taker signs this fill, and the matcher ABI does not carry a fee for the LP
+            // to approve. Start CPI fee derivation at the market-configured base; hybrid/EWMA
+            // movement may still raise it deterministically inside the shared trade path.
+            cfg_pre.trade_fee_base_bps,
             max_market_slots,
         )
     }
@@ -6789,6 +6792,7 @@ pub mod processor {
             stale_matured,
             backing_fee_policy_active,
             fee_bounds_ok,
+            cpi_fee_bps,
         ) = {
             let market_data = market_ai.try_borrow_data()?;
             let (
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Meta impact confirmed against exact pin `7eea209b23e4ebbebcf834130918253c54599402`: both CPI handlers forward the unsigned caller fee into execution. Meta does not wrap or mediate user `TradeCpi` / `BatchTradeCpi`, so no local controller or TWAP guard can mask this direct LP loss path. This PR is a required dependency rollup item for percolator-meta. The new Meta raw-handoff policy-count guard only blocks imported backing-fee batch gates; it does not affect this ordinary trade-fee vector.

aeyakovenko:

PR #225, opened 2026-07-12
[BLOCKER LoF] Keep EWMA movement fees non-reclaimable

aeyakovenko:
## What changed

- add an exact-parent public LiteSVM attack against an independently owned passive LP
- cover `TradeNoCpi`, `BatchTradeNoCpi`, `TradeCpi`, and `BatchTradeCpi` fee routing
- leave only the collected EWMA movement fee outside every authority-withdrawable insurance domain
- retire and burn that claim-free surplus only when `CloseSlab` proves the resolved market has no remaining users or claims
- preserve ordinary trade fees in their local domains

Engine support: aeyakovenko/percolator#146

## Public exploit

A permissionless asset creator also controls that asset's insurance operator. Before this patch, an attacker could:

1. Open a directional position against an independently owned passive LP.
2. Self-trade to pay for and queue an adverse EWMA move.
3. Flatten the temporary self-trade in the same slot without reversing the queued move.
4. Withdraw the movement fee from the attacker-controlled insurance domain.
5. Crank the paid mark and close the directional position against the LP.

On exact `main` parent `da64be63`, the public LiteSVM attack reclaims `10,098,000` atoms, leaves the attacker `4,999,999` atoms net positive, and transfers `5,000,000` atoms from the independent LP. It uses normal public initialization and instructions, with no state injection, dishonest oracle, absent keeper, or rollback assumption violation.

The first fix (`d95f5a0a`) moved the fee to base insurance. The delayed exact-parent regression (`7c8321a5`) proved that was still reclaimable: after every user exited, compromised market authority could withdraw the same fee from the base domain and realize the same profit. The final fix leaves movement fees in engine insurance without assigning them to any authority budget. Once the market is resolved and every claimant is gone, `CloseSlab` atomically retires the accounting amount and burns exactly that many canonical SPL tokens instead of sweeping them to market authority.

Trade availability and mark-clamp behavior are unchanged. Failed terminal burns roll the engine transition back at the instruction boundary.

## TDD evidence

- initial exact-main red: `c30d9b85`
- incomplete base-domain fix: `d95f5a0a`
- delayed base-domain red: `7c8321a5`
- final fix: `2da0413a`
- pre-fix result: fee reclaimed `10,098,000`; attacker gain `4,999,999`; independent LP loss `5,000,000`
- fixed result: authority reclaim `0`; attacker gain `0`; movement fee remains claim-free until terminal SPL burn
- four-route matrix proves the movement fee enters neither creator-controlled nor base-authority budgets

## Validation

- `cargo test --all-targets -- --test-threads=1`: 567 passed
- exact public economic regression passes and terminal `CloseSlab` remains under the 300k custody CU guardrail
- Kani fee-direction contract: 30/30 checks, 3/3 covers
- Kani collected-base-fee isolation contract: 40/40 checks, 2/2 covers
- engine `cargo test --all-targets --features fuzz`: 159 passed
- engine terminal-retirement Kani contract: 132 checks, 3/3 covers

The broad wrapper Kani decoder roster was not rerun here; its eight known wide-decoder timeout harnesses are unchanged because this patch does not alter instruction decoding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=43c654b635956584e730fe14219dd9db83cd4c0d#43c654b635956584e730fe14219dd9db83cd4c0d"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "43c654b635956584e730fe14219dd9db83cd4c0d" }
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #226, opened 2026-07-12
Pin shutdown provider recovery to its owners

aeyakovenko:
## What changed

- Add public LiteSVM exploits for provider backing and funded domain insurance after permissionless-asset shutdown.
- Pin marketauth-driven cleanup destinations and optional ledgers to the configured backing/insurance authorities.
- Keep marketauth able to submit bounded cleanup while providers are offline.
- Correct the full shutdown/retire/reuse lifecycle so provider funds leave to their owners while the slot still retires and reuses.

## Root cause and impact

The mature-shutdown fallback made marketauth an authorized cleanup submitter, but token preflight also treated the transaction signer as the payout owner and rebound optional ledgers to marketauth. After forcing an empty permissionless asset into Recovery and waiting five slots, marketauth could therefore withdraw all provider backing or funded domain insurance to itself.

The TDD backing probe failed on current `main` after marketauth received all 500 atoms. The equivalent insurance route was present in the same fallback. The fixed paths reject marketauth-owned destinations atomically, then succeed under the same marketauth signature when clean destinations are owned by the configured providers. No provider signature is required, so abandoned-slot liveness is preserved.

## Validation

- `cargo test --test v16_cu v16_attack_shutdown_cleanup -- --nocapture`
- `cargo test --test v16_cu shutdown -- --nocapture`
- `cargo test --all-targets` (5 unit + 565 LiteSVM tests)
- auth and hostile matcher fixtures rebuilt before the full run

No engine change.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -7436,15 +7436,12 @@ pub mod processor {
         }
         let profile = read_oracle_profile_for_asset(&market_data, &cfg, asset_index)?;
         let authorities = domain_authorities_from_profile(&cfg, &profile, asset_index);
-        let local_authorized = match authority_kind {
-            DOMAIN_WITHDRAW_AUTH_INSURANCE => {
-                live_authority_matches(&authorities.insurance_operator, authority.key)
-            }
-            DOMAIN_WITHDRAW_AUTH_BACKING => {
-                live_authority_matches(&authorities.backing_bucket_authority, authority.key)
-            }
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by PR135. The consolidated funded-role ownership and terminal attribution checks pin provider and insurance destinations while preserving keeper-submitted cleanup; the generic public payout evidence replaces this branch.

PR #227, opened 2026-07-12
Freeze shutdown drains at stale resolution

### LINKED ISSUE #207: [NOT LoF/DoS] Shutdown-empty provider return after stale maturity
## Summary

`live_domain_withdraw_health_or_shutdown_view` returns early for an empty, shutdown-matured asset before checking whether the market is eligible for permissionless stale resolution.

As a result, `WithdrawInsuranceAsset` can remove a nonzero asset's insurance after the stale-resolution boundary. A later `ResolveStalePermissionless` snapshots the reduced insurance state.

The ordering is present on current `percolator-prog` `origin/main` at `392a376`.

## Impact

The stale-resolution boundary freezes other live value-out paths before terminal resolution. The shutdown-drain branch bypasses that freeze.

The PoC funds asset 1 with 100 insurance atoms, advances past both the shutdown delay and stale-resolution threshold, withdraws all 100 atoms, and then resolves the market. The terminal state therefore contains 100 fewer insurance atoms.

The caller must be the configured insurance operator or `marketauth`; this is a role-gated wind-down invariant violation, not a permissionless withdrawal.

## Root cause

The shutdown exception runs before the global freeze (`src/v16_program.rs:4421-4430`):

```rust
let asset_index = domain / 2;
if shutdown_asset_empty_and_matured_now_view(cfg, group, asset_index).is_ok() {
    return Ok(true);
}
reject_permissionless_resolve_matured_live_view(cfg, group)?;
```

`handle_withdraw_insurance_asset` treats the returned `true` as shutdown-drain authorization, debits the domain budget, and transfers from the vault.

The existing stale-withdrawal test covers asset 0. Asset 0 cannot use this nonzero-asset shutdown path, so it does not exercise the early return.

## Reproduction

<details>
<summary>PoC test</summary>

```rust
#[test]
fn poc_shutdown_drain_bypasses_resolve_maturity_freeze() {
    let mut env =
        V16CuEnv::new_with_market_params_and_price_move(2, 10_000, 10_000, 10_000);
    let admin = env.admin.insecure_clone();

    env.configure_permissionless_resolve_with_cu(5, 5);
    env.configure_auth_mark_with_cu(1, 100);
    env.configure_auth_mark_for_asset_as_admin(1, 1, 100);
    env.top_up_insurance_domain_with_authority(&admin, 2, 100);

    env.svm.warp_to_slot(2);
    env.svm.expire_blockhash();
    env.try_shutdown_asset_with_authority(&admin, 1, 2)
        .unwrap();

    env.svm.warp_to_slot(10);
    let (_, before) = env.market_state();
    assert_eq!(before.assets[1].lifecycle, AssetLifecycleV16::Recovery);
    assert_eq!(before.insurance_domain_budget[2], 100);

    env.svm.expire_blockhash();
    let (dest, _) = env
        .try_withdraw_insurance_asset_with_authority(&admin, 1, 100)
        .expect("shutdown drain bypasses the stale-resolution freeze");
    assert_eq!(env.token_amount(dest), 100);

    let (_, after_withdraw) = env.market_state();
    assert_eq!(after_withdraw.insurance_domain_budget[2], 0);
    assert_eq!(after_withdraw.insurance, before.insurance - 100);

    env.svm.expire_blockhash();
    env.send(
        ProgInstruction::ResolveStalePermissionless { now_slot: 0 },
        vec![AccountMeta::new(env.market, false)],
        &[],
    )
    .unwrap();
    assert_eq!(env.market_state().1.mode, MarketModeV16::Resolved);
}
```

</details>

The test passes against the Sentinel commit `e26cf72` and its matching BPF. The same early-return ordering remains at `392a376`. A current-BPF rerun requires an SBF toolchain supporting the repository's current Rust 1.89 dependencies.

## Suggested fix

Apply the stale-resolution check before the shutdown exception:

```rust
reject_permissionless_resolve_matured_live_view(cfg, group)?;
if shutdown_asset_empty_and_matured_now_view(cfg, group, asset_index).is_ok() {
    return Ok(true);
}
```

Extend the existing stale-withdrawal regression with a shutdown-matured asset 1. Assert that withdrawal fails without changing the domain budget, market insurance, vault, or destination balance. Keep a pre-stale-maturity control where shutdown drain still succeeds.

aeyakovenko:
## What changed

- add a public LiteSVM regression showing a non-base shutdown drain can remove insurance after the global stale-resolution boundary
- check the global stale-resolution freeze before granting the asset-local shutdown withdrawal exception
- retain a pre-maturity control proving normal shutdown drainage still makes bounded progress

## Impact

Before this change, an asset insurance operator or market authority could withdraw a shutdown asset's remaining insurance after `ResolveStalePermissionless` was already available. The later resolution snapshot therefore had fewer assets available for user settlement.

## Root cause

`live_domain_withdraw_health_or_shutdown_view` returned early for an empty, shutdown-matured asset before calling the market-wide stale-resolution guard.

Fixes #207.

## Validation

- targeted red/green LiteSVM reproduction
- shutdown and insurance withdrawal regression filters
- `cargo test --all-targets` (5 unit + 564 LiteSVM passed)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4419,10 +4419,10 @@ pub mod processor {
         domain: usize,
     ) -> Result<bool, ProgramError> {
         let asset_index = domain / 2;
+        reject_permissionless_resolve_matured_live_view(cfg, group)?;
         if shutdown_asset_empty_and_matured_now_view(cfg, group, asset_index).is_ok() {
             return Ok(true);
         }
-        reject_permissionless_resolve_matured_live_view(cfg, group)?;
         if group.header.bankruptcy_hlock_active != 0
             || group.header.threshold_stress_active != 0
             || group.header.loss_stale_active != 0
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -54693,6 +54693,92 @@ fn v16_attack_permissionless_reuse_respects_activation_cooldown_and_fee_atomicit
     );
 }
 
+// A shutdown-matured non-base asset must not bypass the global stale-resolution freeze. Once the
+// base oracle is stale enough to make resolution permissionless, its remaining insurance belongs
+// to the terminal snapshot and cannot still leave through the Live shutdown-drain exception.
+#[test]
+fn v16_attack_shutdown_drain_cannot_reduce_resolve_matured_insurance() {
+    const FUNDED: u128 = 200;
+    const FRESH_DRAIN: u128 = 50;
+    const STALE_DRAIN: u128 = FUNDED - FRESH_DRAIN;
+
+    let mut env = V16CuEnv::new_with_market_params_and_price_move(2, 10_000, 10_000, 10_000);
+    let admin = env.admin.insecure_clone();
+    env.configure_permissionless_resolve_with_cu(20, 5);
+    env.configure_auth_mark_with_cu(1, 100);
+    env.configure_auth_mark_for_asset_as_admin(1, 1, 100);
+    env.top_up_insurance_domain_with_authority(&admin, 2, FUNDED);
+
+    env.svm.warp_to_slot(2);
+    env.svm.expire_blockhash();
+    env.try_shutdown_asset_with_authority(&admin, 1, 2)
+        .expect("empty non-base asset can enter shutdown recovery");
+
+    // Non-vacuous control: after the asset-local delay, but before global stale maturity, the
+    // configured operator can still make bounded shutdown progress.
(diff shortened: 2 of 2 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying for public persistent DoS/LoF under the authority model. The action is restricted to configured privileged reserve roles and an honest permissionless crank bounds the stale interval; terminal reserve coverage remains in PR135.

PR #228, opened 2026-07-12
[HARDENING DoS] Unstick fragmented recovery force-close

aeyakovenko:
## What changed

- Add a public LiteSVM reproduction for fragmented permissionless recovery force-close.
- Pin engine commit `f7630b0dc4b1145196aaaae4c0246cb7c997b760` from aeyakovenko/percolator#119.
- Keep the wrapper production route unchanged; it already calls the affected engine trade path.

## Exploit and root cause

A publicly created non-base market can have one 10-unit short matched against ten independent 1-unit longs. An authenticated 10% mark move leaves the large account maintenance-healthy but below initial margin. The asset is then shut down normally and the delayed permissionless force-close window expires.

Every available opposite portfolio can reduce the large leg by only one unit. Before the fix, `ForceCloseAbandonedAsset` reused the ordinary final initial-margin gate, so the first call failed with `EngineInvalidConfig` at about 193k CU: the large account remained below IM after that partial reduction. The same was true for every available pair. Auto-liquidation was unavailable because the account had zero liquidation deficit, so an absent owner could permanently block permissionless asset wind-down despite balanced aggregate OI.

The engine now admits a matched reduction below IM only when neither absolute position increases and both post-trade accounts remain maintenance-healthy. Risk increases and liquidatable/bankrupt accounts retain the existing IM gate, so this does not create a bilateral bankruptcy escape.

## TDD evidence

- Commit `61e9d608` is the public-only red reproduction on the prior engine pin; the first force-close fails with `Custom(14)` and rolls back.
- Commit `47e85c6f` pins the engine fix; all ten one-unit force-closes succeed, reduce rank monotonically, clear both OI sides, conserve custody, and remain under the trade CU bound.

## Validation

- built `tests/fixtures/auth_matcher` and `tests/fixtures/hostile_matcher`
- `cargo test --all-targets`: 5 unit + 564 LiteSVM tests green
- all 10 force-close tests green, including the existing 14-leg 1.4M-CU boundary
- existing bankrupt and off-mark recycle tests remain rejecting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=f7630b0dc4b1145196aaaae4c0246cb7c997b760#f7630b0dc4b1145196aaaae4c0246cb7c997b760"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #229, opened 2026-07-12
[BLOCKER DoS] Scope bankruptcy locks to affected assets

aeyakovenko:
## Public reproduction

A permissionless asset creator can create asset 1, open both sides through the normal trade API, move the authenticated mark, and permissionlessly crank the short into bankruptcy. Before this fix, the resulting global H-lock permanently rejected both of these unrelated asset 0 exits:

- a live insurance-domain withdrawal
- conversion of fully source-backed favorable PnL

The regressions use only public LiteSVM instructions and normal account construction. They also prove the bankrupt asset's own exits remain locked.

## Fix

- pin engine PR [#120](https://github.com/aeyakovenko/percolator/pull/120) at `cd9f2c522199a220525b96fe15c587a437c330c7`
- preserve the engine's attributed H-lock state in host account serialization
- use the engine's canonical asset-local lock predicate when gating insurance withdrawal
- keep legacy, unattributed, and same-asset bankruptcy states fail-closed
- add a 5,834-asset, near-10 MiB market regression to bound the new check

## Verification

- `cargo test --all-targets`: 5 unit tests and 566 LiteSVM tests passed
- max-shape unrelated withdrawal: 32,693 CU
- engine focused Kani proof: 0/3297 checks failed; 3/3 covers passed

The aggregate wrapper Kani run is not claimed complete: its pre-existing zero-length decoder harness continued symbolic loop unwinding through 125 iterations and was stopped. The first aggregate harness passed, and the directly relevant engine contract passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=cd9f2c522199a220525b96fe15c587a437c330c7#cd9f2c522199a220525b96fe15c587a437c330c7"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-074 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #230, opened 2026-07-13
Restore public market growth above 10 KiB

aeyakovenko:
## Impact

Public `UpdateAssetLifecycle` append operations failed permanently at asset 5 with `InvalidRealloc`, once the market account's absolute data length exceeded 10,240 bytes. Existing users and already-created assets remain usable, so this is a partial protocol-growth DoS rather than a total market DoS.

## Root cause

The default Anchor v2 entrypoint was built with `default-features = false` but without `account-resize`. That omitted the original-account-length metadata required by `AccountInfo::realloc`; each instruction therefore treated the original length as zero.

## Fix

Enable Anchor v2 `account-resize`. The red-first public LiteSVM regression grows a market through all 1,538 permissionless assets allowed by the fee/TVL cap, crosses 10 KiB, recovers a middle-domain provider claim, drains terminal funds, and closes the slab under the CU limit.

## Verification

- `cargo build-sbf --no-default-features --tools-version v1.52`
- fixture SBF builds for auth and hostile matchers
- `cargo test --all-targets`: 5 unit + 564 LiteSVM/CU tests passed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
 percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closed as outside the accepted LoF/persistent-DoS set. The finding concerns permissionless initialization growth after 10 KiB, not loss or lock of existing funded users. It is not required by the PR135 security submission.

PR #231, opened 2026-07-13
[BLOCKER LoF] Prevent signed intent replay across market generations

aeyakovenko:
## What changed

- Add the current asset `market_id` to the signed wire format for `TradeNoCpi`, `TradeCpi`, `BatchTradeNoCpi`, and `BatchTradeCpi`.
- Add the expected generation to the co-signed `UpdateAuthority` and `UpdateAssetAuthority` payloads.
- Reject a generation mismatch before matcher CPI, authority-profile access, or state mutation.
- Persist the next market generation in a 56-byte program-owned PDA keyed by the market address.
- Reserve initial asset IDs from that PDA during `InitMarket` and checkpoint engine-side generation advances during `CloseSlab` before the slab is deleted.
- Allocate and assign a pre-funded system-owned generation PDA so a one-lamport transfer cannot block market initialization.
- Reject legacy signed payloads that do not bind an asset generation.

## Root cause and impact

Asset indices are reusable, but signed instructions previously identified an asset only by `asset_index`, or identified a market only by its account address. A retained transaction signed for generation A could therefore execute after that slot or the full market account had been retired and recreated as generation B.

The public LiteSVM TDD probes demonstrate three concrete attacks using the exact retained transaction object:

- All four trade routes could reopen victim risk after an asset generation changed.
- After `CloseSlab` and same-address `InitMarket`, a retained co-signed `UpdateAuthority` seized `marketauth` and let the attacker resolve a newly funded live market: total market DoS.
- After asset shutdown/restart, a retained co-signed `UpdateAssetAuthority` seized the replacement oracle authority. Moving the mark reduced fresh victim equity from 1,000,000 to 750,000 atoms.
- The full-market trade replay likewise reduced fresh victim equity from 1,000,000 to 750,000 atoms.

The authority cases do not depend on a compromised RPC guessing private intent: the incoming authority is required to co-sign, so the attacker necessarily possesses the complete valid transaction and can withhold it. Every exploit instruction returns `Ok`; SVM rollback does not mitigate the issue.

Portfolio generation tags do not prevent these attacks because the engine receives an otherwise valid request for the current slot. The signed request itself must bind the generation, and the generation namespace must survive market-account deletion.

## ABI and migration

This is a trade, authority-rotation, and lifecycle ABI change:

- Trade clients must read and sign each leg's current `market_id`.
- `UpdateAuthority` must carry asset 0's current `market_id`.
- `UpdateAssetAuthority` must carry the target asset's current `market_id`.
- `InitMarket` and `CloseSlab` must pass the generation PDA plus System Program.
- Old unbound payloads fail closed.

Live markets created before this change migrate when `CloseSlab` checkpoints their current engine counter. A market address whose slab was already deleted before deployment has no recoverable generation history and must not be reused; initialize a fresh market address instead.

## Verification

- `cargo test --all-targets`: 5 unit tests and 568/568 LiteSVM/CU tests passed
- Both new public authority-replay TDD tests pass and verify exact `AssetGenerationMismatch`, byte-for-byte rollback, and successful current-generation rotations
- 3/3 new authority Kani harnesses passed: market-authority field preservation, asset-authority field preservation, and arbitrary legacy unbound-payload rejection
- Existing focused trade-generation Kani harnesses passed in the prior commits
- `cargo fmt --check`
- `git diff --check`

The broad pre-existing aggregate admin trailing-byte harness was also attempted, but remained in generic decoder-loop symbolic expansion until interrupted. The three narrow payload proofs above complete the changed authority ABI without relying on that aggregate harness. The repository-wide Kani command has the same existing decoder-expansion limitation documented by PR 135.

Engine pin remains `143e68c4917ed0400a27b952f036a5677047cd84`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,9 +146,11 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
 - **Per-asset admin keys, isolated — uniform across all assets including asset 0** — one asset's admin
   can never be used against another asset. **✅** Every asset (0..N) carries its own `asset_admin`
   (`AssetOracleProfileV16`): assets 1..N bootstrap it to the activator, **asset 0 bootstraps it to the
-  market admin at `InitMarket`**. `UpdateAssetAuthority { asset_index, kind, new_pubkey }` is scoped to
-  that asset's profile only and now operates on **asset 0 too** (the old `asset_index == 0` rejection is
-  gone). Asset 0 is **not** special for authorities — its only special properties are **fee capture**
+  market admin at `InitMarket`**.
+  `UpdateAssetAuthority { asset_index, market_id, kind, new_pubkey }` is scoped to that asset's
+  current generation and profile only and now operates on **asset 0 too** (the old
+  `asset_index == 0` rejection is gone). Asset 0 is **not** special for authorities — its only
+  special properties are **fee capture**
   (it's the insurance-redirect target) and that it **cannot be permissionlessly created** (it's created
   at `InitMarket`, not via `UpdateAssetLifecycle`).
 - **Each asset (0..N) has a cold-storage admin** that can **rotate that asset's other keys**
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #123, opened 2026-06-09
Add test: deposit pins the vault to the canonical ATA (F-VAULT-FRAG)

aeyakovenko:
## What

Adds `v16_attack_deposit_to_noncanonical_vault_rejected` (litesvm).

## Attack analyzed (liquidity fragmentation → undercollateralization / stranding — SOL-007)

`Deposit` credits engine collateral and moves tokens into `vault_token`, while withdrawals pull from the **canonical** vault (the ATA of `vault_authority`+mint). If deposit accepted *any* `vault_authority`-owned token account, a deposit could be routed into a **non-canonical** account: the engine credits capital, but the tokens land in a vault the protocol never reads for withdrawals — leaving the canonical vault undercollateralized and the deposit stranded.

## Plausibility vs. code — defended

`verify_vault_token_account` pins the vault key to `canonical_vault_address(vault_authority, mint)` (the F-VAULT-FRAG fix).

## Test

Routes a deposit into a fragmented vault (a valid `vault_authority`-owned token account for the right mint, but at a non-canonical address). Asserts: rejects with `Custom(12)` (InvalidVaultAccount), nothing pulled, nothing credited; and a control deposit into the canonical vault credits capital normally (isolating the pin check).

Real BPF via litesvm. **Passes.**

## Why it adds coverage

The F-VAULT-FRAG canonical-ATA pin is tested for withdraw, backing, insurance, and domain top-ups (`backing_withdraw_pinned_to_canonical_vault`, `domain_topups_pinned_to_canonical_vault`, `insurance_topup_pinned_to_canonical_vault`) — but **not** for the deposit path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #232, opened 2026-07-15
Keep price-moved empty asset slots reusable

aeyakovenko:
## Public TDD

This regression reaches the lifecycle failure using only public LiteSVM instructions and a normally initialized market:

1. initialize a two-slot market and an empty user portfolio;
2. configure and legitimately update asset 1's authenticated mark;
3. call the sole public `PermissionlessCrank`, which advances K while the asset has zero effective OI, positions, loss weight, or obligations;
4. retire the empty asset;
5. reactivate the same finite slot under a fresh market id.

On parent engine `143e68c`, step 4 fails with wrapper `Custom(21)` / `EngineLockActive`. Shutdown does not recover the slot because the same empty-state predicate also blocks restart and retirement from Recovery.

With engine PR aeyakovenko/percolator#123 at `7b57c3b8baa684d0d0283fb06babfe4900128a5e`, retire canonicalizes the inert K history and reactivation succeeds. The test checks CU ceilings and proves engine vault accounting, real SPL custody, and the user's capital are unchanged across retirement and reuse.

## Impact

This is a partial public DoS. After any legitimate oracle move, a permissionless cranker can otherwise make an unused configured asset slot permanently unreusable, consuming finite listing capacity. No state injection, malformed account, or initialization mistake is required.

## Scope

- Pin engine PR #123.
- Add one net-new public LiteSVM regression.
- No wrapper program-code change.

## Verification

- Built both `auth_matcher` and `hostile_matcher` SBF fixtures
- `cargo test --all-targets`: 5 wrapper unit tests and 564 LiteSVM tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=7b57c3b8baa684d0d0283fb06babfe4900128a5e#7b57c3b8baa684d0d0283fb06babfe4900128a5e"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by PR135 and engine PR195 terminal-normalization coverage. The current engine normalizes inert price history before retirement; the public reuse lifecycle is retained in the invariant suite.

PR #233, opened 2026-07-15
Keep unrelated epochs from blocking large exits

aeyakovenko:
## Finding

A permissionless asset authority can move its own exposed AuthMark asset and advance the market-wide oracle and risk epochs. On current main, that invalidates otherwise valid certificates for unrelated portfolios. For portfolios with eight or more active legs, the wrapper rejects a valid risk-reducing trade with `EngineStale` before the engine can recertify it.

This is a public partial-DoS / ordering-grief vector, not a total permanent freeze: a separate pre-crank can recover, but an attacker-controlled unrelated asset can invalidate the certificate between the pre-crank and the user's trade.

The TDD test uses only public wrapper instructions and normal account construction. No market or portfolio state is injected. On unmodified main it fails with `Custom(19)` after roughly 102k CU.

## Fix

When a large portfolio has a valid certificate bound to its current active bitmap but only its global epochs are stale, scan the fixed-size active-leg set before rejecting. The wrapper lets the engine recertify inline only when every active leg is already settlement-current for its market generation, lifecycle, side epoch, and K/F/B snapshots.

Actual local settlement staleness still returns `EngineStale` and retains the public pre-crank path, preserving the existing protection against the 2N stale-settlement CU cliff.

The repaired public eight-leg exit consumes 559,365 CU.

## Validation

- `cargo test`: 5 unit tests and 564 LiteSVM tests passed
- `v16_bpf_stale_full_14_leg_tradenocpi_rejects_before_cu_cliff`
- `v16_bpf_current_full_14_leg_tradenocpi_is_under_tx_limit`
- `v16_bpf_full_14_leg_refresh_crank_is_under_tx_limit`

`cargo clippy --all-targets -- -D warnings` remains blocked by 19 pre-existing warnings outside this change.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10891,6 +10891,76 @@ pub mod processor {
         Ok(0)
     }
 
+    fn portfolio_active_legs_are_settlement_current_view(
+        group: &state::MarketViewMutV16<'_>,
+        portfolio: &percolator::PortfolioV16ViewMut<'_>,
+    ) -> Result<bool, ProgramError> {
+        let active_bitmap = portfolio
+            .header
+            .active_bitmap
+            .map(percolator::V16PodU64::get);
+        let configured_assets = group.header.config.max_market_slots.get() as usize;
+        let mut slot = 0usize;
+        while slot < percolator::V16_MAX_PORTFOLIO_ASSETS_N {
+            let leg = portfolio.header.legs[slot]
+                .try_to_runtime()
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closed as nonpersistent ordering grief. A correct pre-crank/refresh restores the owner exit, and PR135 tests stale unrelated epochs plus canonical progress discovery across landing orders.

PR #125, opened 2026-06-09
Security test sweep: 9 LoF/DoS regression tests + 1 fix (mismatched-decimals collateral)

aeyakovenko:
Aggregates the security-loop output into one branch. Each commit is a self-contained litesvm end-to-end test (real BPF) targeting a loss-of-funds (LoF) or denial-of-service (DoS) vector with no prior coverage, plus one behavior fix. All 10 new tests pass; existing suite unaffected.

## The fix (behavior change — review carefully)

**`UpdateBaseUnitMints` must reject a mismatched-decimals collateral pair.** Deposit/withdraw/`SwapSecondaryForPrimary` move atoms 1:1 with no decimals rescaling, so pairing a primary mint with a different-decimals secondary lets a holder deposit the higher-value mint and withdraw the same atom count in the lower-value mint (LoF) — or drain the vault via the favorable direction. The handler now compares `Mint.decimals` and rejects on mismatch. Verified the regression test fails without the guard.

## The tests

1. **dual-mint shared credit** — one deposit can't be withdrawn once per mint (2× collateral mint).
2. **Pyth exponent bound** — `|exponent| ≤ 18` prevents `10^huge` overflow panic (crank DoS) / mispricing.
3. **Chainlink oracle source end-to-end** — first coverage of the Chainlink reader (discriminator/round/freshness/key-binding + decimals bound); adds a `make_chainlink_data` fixture.
4. **dual-mint domain insurance** — per-domain insurance budget can't be double-withdrawn across mints (protocol-fund LoF).
5. **base-unit mints share decimals** — green regression for the fix above.
6. **swap accounting-neutral** — a successful `SwapSecondaryForPrimary` mints no engine collateral; only physical composition rebalances.
7. **epoch-grief flat withdrawal** — a third-party asset append (advancing `asset_set_epoch`) can't freeze a flat holder's withdrawal (market-wide fund-trap DoS).
8. **pushed-mark mode guard** — `PushAuthMark`/`PushEwmaMark` can't override an external-oracle (hybrid) asset's price (manipulation).
9. **deposit canonical-vault pin** — deposit into a non-canonical vault rejects (F-VAULT-FRAG; liquidity fragmentation / undercollateralization).
10. **empty-portfolio re-init** — re-init can't inflate `materialized_portfolio_count` (slab-close DoS).

Supersedes the individual PRs #115–#124 (closed in favor of this aggregate).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #235, opened 2026-07-15
Keep discharged social-loss assets reusable

aeyakovenko:
## What changed

- pin the wrapper to engine commit `df2f4fc6495551ae0e0b8b92430b87219c3f64f4` from engine PR #125
- add one public LiteSVM lifecycle regression proving a fully discharged social-loss asset can be retired and reused
- verify retirement stays within the custody CU limit and preserves both accounting and SPL-token vault balances

## Confirmed public partial DoS

The regression uses only public instructions and normally initialized accounts:

1. open two matched pairs on a flat-price target asset
2. give one target short a second leg on another asset
3. move only that second asset until the portfolio becomes insolvent
4. shut down the target asset and let every claimant exit through `ForfeitRecoveryLeg`
5. observe that all positions, obligations, weights, backing, K, F, dust, and explicit-loss state are discharged while historical `b_long_num` remains nonzero
6. retire and reactivate the now-empty slot

Against parent engine `143e68c4917ed0400a27b952f036a5677047cd84`, step 6 deterministically fails with `Custom(21)` / `LockActive`. Because market asset slots are finite, an attacker can permanently consume a slot through the public recovery lifecycle.

Engine PR #125 removes only current and epoch-start B history from the empty-lifecycle blocker. Claimant- and value-bearing guards remain enforced. Retirement keeps the history for audit; reactivation changes `market_id` and canonicalizes the reused slot.

## Validation

- public red case reproduced against engine `143e68c4917ed0400a27b952f036a5677047cd84`
- `cargo test --all-targets`: 5 unit tests and 564 LiteSVM/CU tests pass
- `git diff --check`
- engine `cargo test --all-targets`: 50 unit, 9 backing, 2 resolved-insolvency, and 69 spec tests pass
- engine Kani retirement proof passes: 4,774 checks, 2/2 covers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=df2f4fc6495551ae0e0b8b92430b87219c3f64f4#df2f4fc6495551ae0e0b8b92430b87219c3f64f4"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by PR135 and engine PR195 terminal-normalization coverage. Discharged social-loss history no longer blocks reuse, with the generic terminal-stock and reuse evidence retained.

Issue #126, opened 2026-06-10
Permissionless cross-asset oracle pin defeats ResolveStalePermissionless


dcccrypto:
## Summary

`handle_push_auth_mark` updates the shared, market-wide staleness clock `cfg.last_good_oracle_slot = max(cfg.last_good_oracle_slot, now)` for **any** `asset_index`, ungated by asset. `permissionless_stale_matured` — the only staleness gate for `ResolveStalePermissionless` and for the resolve-maturity freeze on the cfg-gated value paths — reads that single shared scalar. A non-admin can append an asset, become its `oracle_authority`, and call `PushAuthMark(asset_index = N)` once per `permissionless_resolve_stale_slots` to hold the shared clock fresh, so `ResolveStalePermissionless` never matures — even when asset-0's oracle has been stale for the full window. The protocol's own test `v16_attack_stale_permissionless_asset_cannot_global_resolve_market` asserts this resolve is gated on **market-0's** oracle ("while market-0 oracle is fresh"); the implementation instead keys it on `max()` over all appendable assets.

## Target

| | |
|---|---|
| Program | `4m3ipBQDYX6JQ9YSmUXDjESDHMtGWtiXforkWr9Qoxdi` |
| Deployed BPF (ELF sha256) | `740f561ed3a15346eea9ec34d405f2b64c6e81283574465158ed9f05e6dd7527` (slot 425139356, 1,050,104 B) |
| wrapper | `aeyakovenko/percolator-prog @ fe69816` (line numbers below) |
| engine | `aeyakovenko/percolator @ 58dc118` |
| Market | `4AXbMuJzrUv5KtVs6zc5jDtXTB4XKhKtEBGU6BmkVut4` |

## Mechanism

1. **The shared clock is written by any asset's mark push.** `handle_push_auth_mark` (`src/v16_program.rs:10188`) sets `cfg.last_good_oracle_slot = core::cmp::max(cfg.last_good_oracle_slot, authenticated_slot)` at `:10231` — **before** and outside the `if asset_index_usize == 0` gate at `:10234` that scopes the asset-0-only profile copy. Authorization is per-asset: `expect_live_authority(authorities.oracle_authority, authority.key)` (`:10220`). (The EWMA/configure/crank siblings write the same shared field the same way: `:9868`, `:9977`, `:10081`, `:10172`, `:10499`.)
[...]

PR #236, opened 2026-07-15
[HARDENING DoS] Prevent permissionless asset from forcing global recovery

aeyakovenko:
## What changed

- pin the wrapper to engine commit `48c2cfc45c20ddc6d8c7d8996e1f10409a7ec586` from engine PR #126
- route the new bounded `AdvanceRecoveryClose` auto-crank plan through the wrapper
- avoid requiring an oracle observation for an already-pending recovery close
- add a public LiteSVM regression covering an attacker-controlled permissionless asset and an unrelated healthy base market

## Security impact

A permissionless asset authority could create an insolvent position on its own asset, begin a chunked bankrupt close, shut down the asset, forfeit the leg, and abandon the remaining close work until expiry. An unsigned `PermissionlessCrank` then selected the successful market-wide `DeclareRecovery` route. That committed global Recovery and prevented unrelated base users from closing otherwise healthy positions.

The regression uses only public program instructions and real LiteSVM transaction rollback. It does not inject account state.

## Root cause and fix

The engine selector treated every expired close ledger as global recovery work, including a close for a frozen permissionless asset already in asset-local Recovery. Engine PR #126 adds a bounded `AdvanceRecoveryClose` continuation for that case and preserves global recovery for genuinely drifting active-market closes. This wrapper update dispatches that plan without adding oracle/account requirements.

The test fails against the previous engine pin by observing market-wide Recovery and passes against `48c2cfc`: the asset-local close completes, the market remains Live, and the unrelated base position can close.

## Validation

- SBF builds for both matcher fixtures
- `cargo test --all-targets`: 5 unit tests and 564/564 LiteSVM/CU tests pass
- `cargo check --all-targets`
- wrapper Kani `kani_v16_permissionless_crank_decode_preserves_wire_fields`: pass, 864 checks
- wrapper Kani `kani_v16_legacy_permissionless_crank_size_payload_is_rejected`: pass, 827 checks and both covers
- engine suite: 50 unit, 9 backing-fuzz, 2 resolved-insolvent, and 70 spec tests pass
- engine Kani selector contract and both asset-local/global expiry boundary proofs pass

The repository-wide `cargo kani --tests` command was stopped after more than 30 minutes in the pre-existing `kani_v16_unknown_or_truncated_tags_reject` decoder harness. That harness lacks an explicit unwind bound and was still symbolically unwinding the observation loop; it reported no property failure before termination.

Engine dependency: aeyakovenko/percolator#126.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=48c2cfc45c20ddc6d8c7d8996e1f10409a7ec586#48c2cfc45c20ddc6d8c7d8996e1f10409a7ec586"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #237, opened 2026-07-15
[PRIVILEGED LoF] Lock live insurance while users remain exposed

aeyakovenko:
## Summary
- Reject live `WithdrawInsuranceAsset` while the target asset has any position or unresolved loss state.
- Add a public-only LiteSVM regression for a fully accrued adverse mark with an unrefreshed losing portfolio.
- Keep flat live withdrawal, resolved withdrawal, backing withdrawal, and empty shutdown-drain behavior unchanged.

## Exploit proof
On unpatched `main`, a separate public cranker fully applies marks `100 -> 200 -> 400 -> 800`, clearing market-level stale/lag gates while the short portfolio certificate remains on the old oracle epoch. The insurance operator then publicly withdraws all 500 atoms. Refresh and liquidation subsequently create 450 atoms of winning-side B social loss. No state injection is used.

With this change, that withdrawal rejects atomically. The same refresh/liquidation consumes 450 insurance atoms, leaves 50 insurance, and creates zero B or explicit unallocated loss.

## Validation
- `cargo test --all-targets --no-default-features` (564 LiteSVM + 5 unit tests pass)
- `cargo kani --tests --harness kani_v16_domain_topup_and_asset_insurance_decode_preserves_wire_fields` (821/821 checks pass)
- A direct full `cargo kani --tests` run completed the 823-check active-payload truncation harness, then spent more than 15 minutes expanding the pre-existing monolithic unknown/truncated-tag decoder harness past 140 unwind iterations; that unrelated full run was stopped and is not claimed green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -669,7 +669,7 @@ There are two different insurance withdrawal surfaces:
 - live `WithdrawInsuranceAsset`, which is a per-asset operator path bounded by that asset's funded
   long+short insurance budget
 
-Live insurance withdrawal is intentionally stricter. It is expected to be allowed only when the live market is flat or loss-current, target/effective-lag-free, stress-free, h-lock-free, and has non-negative senior residual. In other words, live insurance can be withdrawn from an empty or fully healthy market, but not while the insurance fund is still protecting unresolved loss or bankruptcy work.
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_privileged_policy_boundary_matrix_rejects_untrusted_callers`. Only the configured insurance authority can invoke the alleged withdrawal; an untrusted public caller rejects with exact market and custody rollback. The executable disposition remains in PR #135; no separate production fix is required.

PR #244, opened 2026-07-15
Block backing escape after committed mark

aeyakovenko:
## Impact

After an AuthMark or EWMA authority committed an adverse mark, the backing authority could withdraw the entire live backing bucket before the first crank published that target into the engine. Public cranks then realized a winner claim that exceeded the losing trader's principal, leaving the claim underbacked.

The public LiteSVM regression demonstrates the old behavior with both price-managed market types: 1,000,000 backing escaped, then bounded public cranks produced a 137,161 claim against 100,000 of losing principal.

## Root cause

The live backing-withdrawal gate checked only the engine's `raw_oracle_target_price` against its effective price. `PushAuthMark` and `PushEwmaMark` first commit the target in the wrapper oracle profile, so there is a public pre-crank interval where the economically committed target is absent from that engine-only check.

## Fix

For exposed price-managed assets, include the wrapper profile's committed mark and oracle target in the existing target/effective lag gate. This keeps the reserve in place until bounded accrual catches up. It does not publish a new engine raw target during mark submission, which would invalidate account certificates and force an extra crank between EWMA-moving trades.

## Verification

- Regression failed against the parent program and passes with the fix for AuthMark and EWMA.
- `cargo test --lib` (5/5)
- `cargo test --test v16_cu` (564/564, including CU assertions)
- `cargo test --all-targets`
- SBF fixture builds for `auth_matcher` and `hostile_matcher`
- `cargo fmt --check`

Full Kani was also started; the payload-truncation proof passed and the unrelated high-cost decoder proof was still progressing when this PR was published.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4438,7 +4438,7 @@ pub mod processor {
         if asset_local_loss_stale_view(group, asset_index) {
             return Err(PercolatorError::EngineLockActive.into());
         }
-        reject_exposed_target_effective_lag_view(group, asset_index)?;
+        reject_exposed_target_effective_lag_view(cfg, group, asset_index)?;
         Ok(false)
     }
 
@@ -4460,6 +4460,7 @@ pub mod processor {
     }
 
     fn asset_has_exposed_target_effective_lag_view(
+        cfg: &WrapperConfigV16,
         group: &state::MarketViewMutV16<'_>,
         asset_index: usize,
     ) -> Result<bool, ProgramError> {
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Clean-room downstream Meta integration found a second public interval that this patch does not close.

After a normal Meta controller handoff preserving an external backing authority, the probe:

- tops up short-loss domain 1 with 1,000,000 atoms;
- opens balanced OI at 1,000,000 with 1,100,000 short capital;
- commits AuthMark 3,000,000;
- uses an unrelated empty portfolio to permissionlessly crank slots 101 and 102 until both `raw_oracle_target_price` and `effective_price` equal 3,000,000;
- withdraws all 1,000,000 backing atoms through public `WithdrawBackingBucket`;
- permissionlessly refreshes the two stale portfolios.

The withdrawal succeeds against pinned `da64be639168b496833dd4c701607a94fcb06b6c`. The later source claim is under-backed by exactly 900,300 atoms. No state bytes are injected; the real Percolator and Meta controller SBF binaries are used.

The profile target/effective check added here is clear by the time of withdrawal, so this PR remains vulnerable after another account publishes the mark but before the winner absorbs its private K snapshot. The market slab has no aggregate that proves every exposed portfolio is current. Principal withdrawal therefore needs the same fail-closed position/loss-state lock proposed for live insurance in #237 (or a new exact aggregate), plus a regression for this post-effective interval. Backing earnings are a separate class; this evidence concerns principal only.

aeyakovenko:
The exact downstream red LiteSVM repro is preserved at aeyakovenko/percolator-meta@6ee5403.

aeyakovenko:
Closed as a privileged authority-compromise/self-rug scenario rather than a public untrusted-caller exploit. PR135 proves role authentication and explicitly scopes the remaining economic trust assumption; operational threshold control belongs in the authority setup.

PR #245, opened 2026-07-15
[REAL LoF] Harden authority marks against signer compromise

aeyakovenko:
## Finding

A single compromised AuthMark signer could drain a permissionless LP through public instructions.

The pre-fix LiteSVM reproduction used no protocol-state injection:

- attacker deposited 60,000; LP deposited 2,000,000 and opted into the public matcher
- attacker opened a one-unit long without an LP signature after opt-in
- the compromised oracle moved the authority mark from 1,000,000 toward 2,000,000
- repeated public cranks converged the production 24 bps per-slot cap in about 290 slots
- attacker closed and withdrew 1,060,000; LP capital fell to 1,000,000

The movement cap delayed the forged target but did not bound cumulative loss.

## Root cause

`PushAuthMark` and `PushEwmaMark` trusted one operational oracle key. That key also authorized all three oracle configuration handlers while an asset was flat, so compromise exposed both live reporting and mode/feed/anchor governance.

## Fix

- require distinct configured `oracle_authority` and `backing_bucket_authority` signatures for every AuthMark/EwmaMark push
- move Hybrid/AuthMark/EwmaMark configuration to the cold per-asset `asset_admin`
- migrate all push call sites to the three-account ABI
- document the trust boundary and deployment migration
- add public LiteSVM coverage for the economic drain, equal-key pseudo-quorum, all three configuration handlers, per-asset isolation, authority rotation, and positive liveness controls

## Compatibility

This intentionally changes authority-mark push accounts to:

`[oracle_authority signer, market writable, backing_bucket_authority signer]`

Deployments with equal oracle/backing keys must rotate one role with `UpdateAssetAuthority` before the next mark push.

Compromise of `asset_admin`, or compromise/collusion of both quorum authorities, remains a privileged trust assumption.

## Verification

- `cargo test --all-targets`: 5 unit + 566 LiteSVM tests passed
- targeted Kani oracle-mark wire proof: 837 checks passed
- `cargo clippy --all-targets`
- `git diff --check`
- 10 MiB CU control: PushEwmaMark 4,538 CU; liquidation crank 229,338 CU

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,8 +446,8 @@ This section describes intent and operational ordering, not argument-by-argument
 - External-oracle markets authenticate configured oracle account(s) in the oracle configuration/crank
   paths; `TradeCpi` / `TradeNoCpi` use the already-stored effective mark for fee and settlement
   accounting.
-- AuthMark markets use **ConfigureAuthMark** (tag 62) and **PushAuthMark** (tag 63), signed by the configured mark authority, to store a direct authority mark without EWMA smoothing.
-- EwmaMark markets use **ConfigureEwmaMark** (tag 35) and **PushEwmaMark** (tag 36), signed by the configured mark authority, to update a smoothed EWMA mark input.
+- AuthMark markets use **ConfigureAuthMark** (tag 62) and **PushAuthMark** (tag 63) to store a direct authority mark without EWMA smoothing. Configuration requires the asset's `asset_admin`; a push requires the distinct configured `oracle_authority` and `backing_bucket_authority` to co-sign.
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as out of scope for the program threat model. The demonstrated loss requires compromise of the configured AuthMark signer, which is already a trusted pricing authority rather than an unauthorized public-interface exploit. Deployments that require multi-party control of this authority should configure it behind a Squads multisig instead of adding a second bespoke signer/quorum mechanism to the program.

PR #246, opened 2026-07-15
[BLOCKER DoS] Prevent permissionless micro-price crank pinning

aeyakovenko:
## Impact

A permissionless caller could indefinitely pin a live exposed market away from its authenticated oracle target when the percentage movement cap rounded a short interval below one price atom.

At price 100 with a 24 bps/slot cap and target 200, one public `PermissionlessCrank` per slot previously succeeded and advanced engine `slot_last` while leaving `effective_price` at 100. Repeating this erased the elapsed interval needed for a representable move, freezing price-dependent PnL, liquidation, and funding progress despite real long and short OI.

The fail-first LiteSVM probe used only public instructions on a normally initialized AuthMark market. Against engine `143e68c4917ed0400a27b952f036a5677047cd84`, the first attacker crank incorrectly succeeded at 69,481 CU instead of returning `NonProgress`.

## Fix

- Pin engine `a8b8877e784e34fbef7bfd16ce7228130219904b` from [percolator#128](https://github.com/aeyakovenko/percolator/pull/128). The engine rejects exposed, lagged, positive-time accrual segments whose capped effective price does not change, before any mutation.
- Mark observations as `observations_preaccrued` after the wrapper authenticates and applies them, so auto-crank dispatches the selected account action without replaying a second market segment.
- Propagate every auto-crank engine error to the instruction boundary. No `NonProgress` result is converted into a successful partial commit; SVM rollback preserves the elapsed interval.
- Keep the 10 MiB/highest-asset liquidation liveness check exact: the bounded price/refresh/liquidation sequence completes in three calls, with peak liquidation work at 245,179 CU.

## Coverage

`v16_attack_permissionless_micro_price_cranks_cannot_pin_oracle_progress` creates opposing positions through the public trade API, commits target 200, and submits attacker cranks at every slot. Slots 1-4 and 6-9 must return engine `NonProgress` with byte-identical market state; slots 5 and 10 must move one atom toward the target within the crank CU budget.

This is net-new coverage of target-distance liveness under integer cap quantization. Existing liveness tests treated successful `slot_last` advancement as progress and did not exercise a successful zero-price-delta state transition.

## Verification

- `cargo test --all-targets` (6 unit + 565 LiteSVM tests)
- `cargo test --test v16_cu v16_attack_permissionless_micro_price_cranks_cannot_pin_oracle_progress -- --exact --nocapture`
- `cargo test --test v16_cu v16_bpf_permissionless_liquidation_is_bounded -- --exact --nocapture` (242,236 CU)
- `cargo test --test v16_cu v16_bpf_10m_market_liquidation_high_asset_stays_bounded -- --exact --nocapture` (245,179 CU)
- `cargo kani --tests --harness kani_v16_permissionless_crank_decode_preserves_wire_fields`
- Engine Kani: `proof_v16_lagged_target_zero_delta_cannot_consume_accrual_segment` (0/3,329 failed; cover satisfied)
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=a8b8877e784e34fbef7bfd16ce7228130219904b#a8b8877e784e34fbef7bfd16ce7228130219904b"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "a8b8877e784e34fbef7bfd16ce7228130219904b" }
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #247, opened 2026-07-16
Prevent permissionless crank-cadence rounding loss

aeyakovenko:
## Impact

A permissionless caller could settle a victim after every fractional K/F movement. Each refresh floored PnL against the account and advanced its snapshots without retaining the sub-atom remainder.

The public LiteSVM regressions use fresh, valid markets and only public instructions:

- K settlement: the same 0.001-lot short and final effective mark leaves `9,999,998` capital atoms after one delayed refresh but only `9,999,989` after ten fragmented permissionless refreshes.
- F settlement: both runs use the exact same per-slot market-accrual sequence. At a 1 bps mark cap, every K step settles integrally; only F is fractional. Refreshing the 0.1-lot victim each slot leaves PnL `100` versus `109` when refreshed once.

No account bytes or injected engine state are used. An attacker can therefore force conservative rounding repeatedly and take value from another live user.

## Fix

This PR pins engine commit [a74b81b](https://github.com/aeyakovenko/percolator/commit/a74b81b2112aba670334f7d1d5af58a2a233f910), proposed in [engine PR #129](https://github.com/aeyakovenko/percolator/pull/129). The engine:

- persists Euclidean K and F settlement remainders per leg
- preserves those remainders across same-side leg transformations
- applies funding precision before settlement rounding
- bumps the account-layout discriminator

## TDD

Both new LiteSVM tests fail against engine `143e68c` with the value divergences above and pass against `a74b81b`:

- `v16_attack_permissionless_refresh_cadence_cannot_magnify_fractional_price_loss`
- `v16_attack_permissionless_crank_cadence_cannot_erase_funding`

The existing funding-index expectation is updated for the exact, non-truncated index delta.

## Verification

- SBF fixture builds for auth and hostile matchers
- `cargo test --all-targets`: 566/566 LiteSVM/CU tests plus all host targets
- Kani: 38/38 wrapper harnesses, with unwinding checks enabled and exact per-loop bounds for the bounded decoder/feed loops

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=a74b81b2112aba670334f7d1d5af58a2a233f910#a74b81b2112aba670334f7d1d5af58a2a233f910"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Superseded by PR135 and consolidated engine PR195. The canonical accrual implementation and history-partition invariant cover K/F cadence, with public terminal entitlement evidence retained.

PR #248, opened 2026-07-16
Collect maintenance fees before live withdrawal

aeyakovenko:
## Impact

A live user could wait while account-level maintenance fees accrued, flatten both sides through the public `TradeNoCpi` interface, withdraw each stale pre-fee capital balance, and close both portfolios in one transaction. The transaction transferred all collateral, collected no maintenance fee, and removed the accounts against which the fee could later be realized.

## Fix

- Settle the engine's account maintenance fee at the authenticated market slot before a live withdrawal.
- Credit the charged amount to the wrapper's canonical asset-0 insurance budget.
- Preserve an atomic withdraw-all path by treating a request for the pre-fee full balance as a request for the resulting post-fee balance.
- Keep partial withdrawals exact and preserve the existing over-`u64` validation behavior.

## Public TDD regression

`v16_attack_atomic_live_exit_cannot_erase_elapsed_maintenance_fee` creates a normal permissionless asset and two funded portfolios, opens a matched position, advances the market through public cranks, then atomically closes the trade, withdraws both portfolios, and closes both accounts. Before the fix, both users received 10,000 atoms and insurance received zero. With the fix, both receive 9,910 atoms, insurance receives the 180-atom elapsed obligation, both accounts close, and wrapper/SPL custody remains exact under the 1.4M CU limit.

## Verification

- auth and hostile matcher SBF fixture builds
- focused regression
- all 58 withdrawal-filtered tests
- all 16 maintenance-filtered tests
- `cargo test --all-targets --no-default-features` (6 unit + 565 LiteSVM tests)
- `cargo check --all-targets --no-default-features`
- `cargo kani --tests --exact --harness kani_v16_unknown_or_truncated_tags_reject --unwind 18` (814 checks, 0 failures)
- `git diff --check`

Engine pin remains `143e68c4917ed0400a27b952f036a5677047cd84`.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -5718,11 +5718,12 @@ pub mod processor {
             &vault_authority,
             &cfg,
         )?;
-        let amount_u64 = amount_to_u64(amount)?;
-        require_token_balance(vault_token, amount_u64)?;
-
+        if amount == 0 {
+            return Ok(());
+        }
+        amount_to_u64(amount)?;
         ensure_portfolio_storage_for_market_slots(portfolio_ai, max_market_slots)?;
-        {
+        let withdrawn_amount = {
             let mut market_data = market_ai.try_borrow_mut_data()?;
             let (cfg, mut group) = state::market_view_mut(&mut market_data)?;
             if group.header.mode != 0 {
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closed during consolidation. The observed effect was avoidable maintenance revenue rather than theft of another user principal; PR135 nevertheless centralizes pre-debit maintenance collection and retains fee-entitlement coverage.

PR #249, opened 2026-07-16
Preserve Hybrid oracle progress under omitted observations

aeyakovenko:
## Public reproduction

A stale permissionless auto-crank with an empty selected-asset observation can land before a valid Hybrid oracle crank. Before the engine fix, that empty crank consumed the selected asset's positive-time accrual slot using the committed old price, so the legitimate observation could no longer move the price. Repeating this ordering can suppress Hybrid oracle progress while the attacker continues.

The LiteSVM TDD builds a live market and positions through public instructions. Both the victim and epoch-staling assets use external Pyth observations. After normal market initialization, every protocol transition in the attack is permissionless; the test does not inject or mutate protocol state.

On the pre-fix engine (`143e68c4917ed0400a27b952f036a5677047cd84`), the public reproduction fails with:

`omitted Hybrid observation consumed slot 2 at stale price 1000000`

## Invariants tested

- Omitted observation first: account progress succeeds, market state is byte-stable, then the valid Hybrid observation consumes the slot and moves price.
- Valid observation first: a late empty stale crank still progresses the account without accruing the selected asset again.
- The public path remains best-effort; an omitted observation is not rejected.

## Fix

Pins engine commit `00edc666def4be2cda07ece580ae3e53251e26b0` from [aeyakovenko/percolator#133](https://github.com/aeyakovenko/percolator/pull/133). The engine now distinguishes a supplied selected-asset observation from the committed fallback price: account action may use the fallback, but missing evidence cannot authorize market accrual or consume the elapsed slot.

## Validation

- Pre-fix public LiteSVM TDD: fails on the slot-consumption assertion above.
- This PR: `cargo test --all-targets` passes 6 unit tests and 567/567 LiteSVM/CU tests.
- Wrapper contracts are unchanged by this pin/test PR. The direct `cargo kani --tests` run on [PR135](https://github.com/aeyakovenko/percolator-prog/pull/135) at `fa0dd62f` passes 40/40 harnesses, and its runtime suite passes 6 unit tests plus 606/606 LiteSVM/CU tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=a8b8877e784e34fbef7bfd16ce7228130219904b#a8b8877e784e34fbef7bfd16ce7228130219904b"
+source = "git+https://github.com/aeyakovenko/percolator?rev=00edc666def4be2cda07ece580ae3e53251e26b0#00edc666def4be2cda07ece580ae3e53251e26b0"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Superseded by PR135 and consolidated engine PR195. Observation-free calls no longer consume authenticated accrual progress, and PR135 retains both landing orders and the public oracle-progress liveness oracle.

PR #250, opened 2026-07-17
[REAL LoF] Prove ADL split-transfer backing drain is closed

aeyakovenko:
## Impact

A public quantity-ADL split trade could copy already-haircut raw basis into a fresh portfolio. On the vulnerable engine pin, the end-to-end LiteSVM sequence converted 13,585 atoms against 12,441 atoms of counterparty loss, creating a 1,144-atom claim on honest backing.

The reproduction uses only public instructions: deposits, a normal trade, maintenance-driven permissionless liquidation, `TradeNoCpi`, an authority-mark update, permissionless cranks, and owner conversion. It does not inject or mutate account state.

## Change

- Pin engine commit `54d21977695d24351be5b2ed919ff83fc33a78c5` from engine PR aeyakovenko/percolator#134.
- Add a public LiteSVM regression proving post-ADL basis cannot be reissued through a split transfer.
- Prove the rejected trade propagates `LockActive` and SVM rollback leaves market, both portfolios, and vault byte-for-byte unchanged.
- Prove subsequent mark settlement is zero-sum within one atom of conservative rounding and does not move vault custody.
- Prove the existing owner can still reduce the position within the custody CU guardrail.

## Validation

- `cargo test --all-targets`: 6 unit tests and 566 LiteSVM tests passed, including all CU tests.
- Engine `cargo test --all-targets`: 131 passed.
- Engine `cargo test --features fuzz`: 160 passed.
- Wrapper Kani: 39/40 harnesses passed with checked unwind bounds. The unchanged `configure_hybrid_oracle` ABI harness still expands the monolithic decoder past 18 GB at its exact 97-step bound; no semantic counterexample was produced.
- `git diff --check`

Engine issue: aeyakovenko/percolator#131
Engine fix: aeyakovenko/percolator#134

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=54d21977695d24351be5b2ed919ff83fc33a78c5#54d21977695d24351be5b2ed919ff83fc33a78c5"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-061 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #251, opened 2026-07-17
[BLOCKER LoF] Prevent delayed asset-authority revival

aeyakovenko:
## What changed

- Require the current holder of every `UpdateAssetAuthority` role to sign its handoff.
- Keep the incoming-key co-signature and the rule that only `asset_admin` may be burned.
- Add a public LiteSVM regression for a delayed, previously valid admin-signed operator assignment that lands after a newer holder-signed correction.
- Retain coverage for the separate compromised-admin provider takeover and reconcile authority lifecycle fixtures and the README.

## Root cause and public exploit

`UpdateAssetAuthority` let an unchanged `asset_admin` replace a distinct operational authority without the current holder. A stale but honestly signed admin assignment therefore remained valid after a newer authority correction.

The exact-parent LiteSVM reproduction uses only public instructions and normally initialized accounts:

1. A cold asset admin signs an operator assignment to key B but it is delayed.
2. The current operator lands a newer co-signed handoff to key C.
3. An independent insurance provider funds 50,000 atoms after C is visible.
4. An unprivileged relayer lands the retained admin+B assignment under the same recent blockhash.
5. B is revived as operator and withdraws all 50,000 atoms of the provider's fresh reserve.

No key currently authorized at step 4 is compromised, no protocol bytes are injected, and the victim provider did not sign the harmful transition. This is `[BLOCKER LoF]` under `scripts/loop.md`'s honest-signature replay rule. The older compromised-admin takeover remains a separate `[PRIVILEGED LoF]` consequence of the same root cause.

## Fix

Every selected role now owns its own handoff: the current holder must sign and a nonzero replacement must co-sign. `asset_admin` may rotate or burn only itself. After a role is handed to another key, stale admin assignments and unilateral admin takeovers reject before mutation.

## TDD evidence

- Exact parent `36fa587e1424a8c7fdde2c701c10938188a51876` with RED test commit `be3f713b`: the retained handoff lands, B drains 50,000 atoms, and the test panics.
- Fixed head `255e56ee`: the stale handoff rejects; market and vault remain byte-identical; B remains unauthorized; the provider reserve remains intact.

## Verification

- Built real `auth_matcher` and `hostile_matcher` SBF fixtures
- `cargo test --all-targets -- --test-threads=1`: 567/567 LiteSVM/CU + 6/6 unit tests passed
- `cargo check --all-targets`
- `cargo fmt` and `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,12 +159,12 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
   gone). Asset 0 is **not** special for authorities — its only special properties are **fee capture**
   (it's the insurance-redirect target) and that it **cannot be permissionlessly created** (it's created
   at `InitMarket`, not via `UpdateAssetLifecycle`).
-- **Each asset (0..N) has a cold-storage admin** that can **rotate that asset's other keys**
-  (insurance/operator/backing/oracle), **shut down/restart that asset after the exit+empty gates**,
-  and **can be burned (set to 0)** — a credibly admin-free asset that can't be locally revived after
-  shutdown. **✅** For asset 0 this means the market admin starts as the asset-0 admin and can
-  force-replace the shared insurance operator/authority via `UpdateAssetAuthority`, while required
-  domain authorities themselves cannot be burned to zero.
+- **Each asset (0..N) has a cold-storage admin** that controls its own handoff and can
+  **restart that asset after the exit+empty gates**, and **can be burned (set to 0)** - a credibly
+  admin-free asset that cannot be locally revived after shutdown. **✅** Operational authorities
+  (insurance/operator/backing/oracle) each control their own co-signed handoff. An `asset_admin`
+  cannot replace a distinct holder or seize its funded domain; required operational authorities
+  cannot be burned to zero.
 - **One market-level key: `marketauth`.** **✅** All market-level governance collapses into a single
   `WrapperConfigV16.marketauth` key (it replaced the former separate `admin` / `asset_authority` /
   `base_unit_authority`). `marketauth` is the only key that can: **create market 0** (`InitMarket`),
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Blocking Meta compatibility finding at head `3988a0723ee44b7b2b4d5253d496296f0cc9bfd5` against percolator-meta master `2cc8269931b99ac4351b70705236bbae6fd7ef55`. I built the PR's real SBF and loaded it into the unchanged Meta LiteSVM chain.

Two existing public-lifecycle regressions fail at the role rotation with `Custom(8)`:

```text
cargo test -p twap-program --test chain e2e_resolved_asset0_backing_is_returned_only_to_its_recorded_provider -- --nocapture
# fails: any cranker returns backing through the canonical pool admin

cargo test -p twap-program --test chain e2e_controller_freezes_shutdown_returns_when_stale_resolution_matures -- --nocapture
# fails: resolved cleanup returns the provider's insurance
```

The asset-0 case is current-design state, not a predecessor fixture: the independent provider deposits real backing, the market resolves and empties, and the provider is absent. Meta's amountless resolved cleanup currently uses the constrained `asset_admin` to rotate only the value role, derives both amounts and the recipient from the slab, and pays the recorded provider. Requiring the absent current holder to sign makes that cleanup impossible. Whole-market stale resolution also freezes the shutdown-return path, so the provider atom can remain a permanent `CloseSlab` blocker.

The live takeover fixed by this PR is REAL LoF and should remain blocked. The upstream API also needs a terminal `resolved && empty` marketauth/controller exception (or an equivalent fixed provider-bound withdrawal) that cannot select an amount or beneficiary. Without that terminal path, this PR trades the live LoF for a public terminal DoS and is not safe to merge into the Meta pin.

aeyakovenko:
Strict Meta/public classification update: the live takeover requires the target asset's `asset_admin`. `percolator-meta` never exposes raw tag 65 through its governance proxy; its only controller-signed role rotations are fixed shutdown/resolved-empty cleanup paths with slab-derived amounts and recorded-provider-bound destinations. An unaffiliated caller, external backing provider, or governance signer cannot select a live successor through Meta.

So the underlying raw-Percolator defect is privileged LoF, while this head's absent-holder terminal regression is a confirmed public DoS for Meta (see the earlier exact-SBF failures). A mergeable version needs holder consent during live custody plus a narrowly constrained `resolved && empty` terminal rotation/return path; this PR remains blocked until both properties hold.

aeyakovenko:

PR #252, opened 2026-07-18
[BLOCKER DoS] Restore terminal exit after partial Recovery forfeit (#236)

aeyakovenko:
## Classification: [BLOCKER DoS]

A public losing trader can strand every user after a normal secondary-asset shutdown. The reproduction uses only program instructions and the default certified engine profile. It does not mutate account bytes, inject a close ledger, forge an oracle, or require a malicious market authority.

## Public attack

1. Two funded users open opposite positions on a secondary asset.
2. Authenticated marks advance within the configured per-slot price cap until one side is bankrupt.
3. The market authority performs the normal secondary-asset shutdown transition.
4. The losing owner calls `ForfeitRecoveryLeg` once. The default bounded bankruptcy chunk commits, leaving a real `100_000_000` residual and an active close ledger, then the owner disappears.
5. `ForceCloseAbandonedAsset`, the opposite owner's `ForfeitRecoveryLeg` and `RebalanceReduce`, and pre-expiry auto-crank all reject.
6. After `max_bankrupt_close_lifetime_slots`, public auto-crank escalates the global market to `Recovery`.
7. Before this fix, both wrapper resolve routes reject `Recovery`, while withdrawal requires `Live` and terminal close requires `Resolved`. An independent funded winner has no terminal route.

This disproves the earlier state-injection-only premise. The liquidation preflight is not the only public close-ledger producer; `ForfeitRecoveryLeg` can organically commit a partial bankruptcy ledger.

## Fix

- Pin engine commit `493fcacc55e236af36e0c8c6e683923ee50a2365`, which permits `Recovery -> Resolved`.
- Let admin `ResolveMarket` and mature `ResolveStalePermissionless` reject only an already-`Resolved` market.
- Keep stale maturity, authenticated clock, owner, and engine validation unchanged.

Once resolved, the regression proves bounded public progress through the abandoned loser, two bounded winner B-loss steps, and a real SPL-vault payout. The winner recovers at least principal; the vault falls by exactly the payout and senior conservation still holds.

## TDD evidence

Exact parent:

- wrapper `36fa587e1424a8c7fdde2c701c10938188a51876`
- engine `143e68c4917ed0400a27b952f036a5677047cd84`
- `v16_attack_partial_recovery_forfeit_has_public_terminal_escape`
- RED: mature `ResolveStalePermissionless` returns `Custom(21)` after the public sequence reaches global `Recovery`

Fixed branch:

- wrapper `631abdec`
- engine `493fcacc55e236af36e0c8c6e683923ee50a2365`
- GREEN: the same public setup reaches `Resolved`, clears both users in bounded calls, and pays the independent winner through the canonical vault

## Verification

- focused public RED/GREEN regression
- Recovery filter: 12/12
- resolve filter: 65/65
- full LiteSVM/CU suite: 566/566
- wrapper unit tests: 6/6
- `v16_kani` target compiles cleanly (proof harnesses are Kani-gated)

Paired engine PR: aeyakovenko/percolator#136.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=493fcacc55e236af36e0c8c6e683923ee50a2365#493fcacc55e236af36e0c8c6e683923ee50a2365"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "493fcacc55e236af36e0c8c6e683923ee50a2365" }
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #253, opened 2026-07-18
[BLOCKER LoF] Preserve rounded premium funding on auto-crank

aeyakovenko:
## Impact

A public cranker could omit the engine-selected asset observation when the per-slot price cap rounded the effective-price movement to zero. On the negative-premium branch, conservative signed rounding still produced nonzero funding. Landing the omission first advanced the asset slot with a zero funding rate, permanently erasing funding owed by an attacker-controlled short to an independent long.

The public LiteSVM reproduction uses a normally initialized live two-asset market, honest AuthMark updates, signed trades, and permissionless cranks only. It does not inject or mutate account state.

On exact pre-fix `origin/main` (`da64be639168b496833dd4c701607a94fcb06b6c`), test commit `d90ec2dc` fails with:

```
control: long=100100000 short=99900000 f_long=1000000000000000 f_short=-1000000000000000
attack:  long=100000000 short=100000000 f_long=0                f_short=0
```

The attacker retains `100,000` collateral atoms and the independent victim loses the same `100,000` atoms.

## Fix

The wrapper now accepts an omitted selected-asset observation only when it changes neither effective price nor computed premium funding. Otherwise the crank rejects atomically, and the same bounded permissionless crank succeeds when resubmitted with the selected observation.

## Verification

- Exact-parent public LiteSVM red reproduction
- Fixed public LiteSVM green reproduction with identical control/attack funding indexes and resolved token payouts
- `cargo test --all-targets`: 566/566 LiteSVM/CU, 6/6 host, all examples green

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10663,7 +10663,7 @@ pub mod processor {
             if let Some(asset_index) =
                 auto_crank_selected_asset_that_accrues_view(&portfolio, &summary)?
             {
-                reject_missing_pending_selected_observation_view(
+                reject_missing_selected_observation_that_changes_accrual_view(
                     &cfg,
                     &group,
                     asset_index,
@@ -10885,7 +10885,7 @@ pub mod processor {
         Ok(None)
     }
 
-    fn reject_missing_pending_selected_observation_view(
+    fn reject_missing_selected_observation_that_changes_accrual_view(
         cfg: &WrapperConfigV16,
         group: &state::MarketViewMutV16<'_>,
         asset_index: usize,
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent clean-room real-SBF confirmation on the funding payer's own refresh path.

Distinct case from the PR test: price `200`, honest AuthMark `199`, attacker-controlled short, independent victim long, and the attacker cranks its own stale short while omitting asset 0. A second normally traded asset advances the oracle epoch; all portfolios are independently funded and every state transition uses public instructions.

Exact parent `da64be639168b496833dd4c701607a94fcb06b6c`:
- observed control: `f_long_num = 1e15`, `f_short_num = -1e15`, attacker/victim payouts `99,900,000 / 100,100,000`
- omitted own-refresh lands: both indexes zero, payouts `100,000,000 / 100,000,000`
- exact victim-to-attacker transfer: `100,000` atoms

Exact head `e718d09b892d4d3e09ad1828bfa2d0a3259a20ab`: the omitted refresh rejects, the bounded observed retry lands, and indexes plus terminal token payouts exactly equal control. This confirms #253 as a real public LoF, including the stronger case where the funding payer erases its own debt.

aeyakovenko:

PR #254, opened 2026-07-18
[PRIVILEGED LoF] Accrue committed funding before asset shutdown

aeyakovenko:
## Finding

A compromised non-oracle asset admin could call the public shutdown instruction after an honest AuthMark had committed a premium but before the next permissionless accrual. Shutdown entered Recovery and reset the oracle profile to the frozen effective price without booking that deterministic funding segment.

The exact-parent LiteSVM repro uses independent long and short owners. The control books `f_long_num = 1e15`, `f_short_num = -1e15` and pays `100,100,000` / `99,900,000`; shutdown-first zeroed both funding indexes and paid both sides `100,000,000`, transferring 100,000 atoms away from the owed long. No account bytes are injected and the oracle signer remains honest and distinct from the shutdown admin.

## Fix

- derive the next effective price from the already committed internal mark
- book a value-active price/funding segment before entering Recovery
- require ordinary bounded permissionless catch-up when a value-active asset is more than one engine accrual segment behind
- retain the prior no-op shutdown behavior when no exposed price or funding change is pending

## Verification

- exact-parent public LiteSVM test fails with the payout/index divergence above
- fixed public LiteSVM test matches the control exactly
- `cargo test --all-targets` (566/566 LiteSVM/CU plus host/example targets)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9291,6 +9291,54 @@ pub mod processor {
                 }
                 match group.markets[asset_index].engine.asset.lifecycle {
                     ASSET_LIFECYCLE_ACTIVE | ASSET_LIFECYCLE_DRAIN_ONLY => {
+                        let asset_slot = group.markets[asset_index].engine.asset.slot_last.get();
+                        let shutdown_dt = authenticated_slot
+                            .checked_sub(asset_slot)
+                            .ok_or(PercolatorError::EngineStale)?;
+                        let shutdown_price = committed_effective_price_for_accrual_view(
+                            &profile,
+                            &group,
+                            asset_index,
+                            authenticated_slot,
+                        )?;
+                        let funding_rate_e9 = permissionless_funding_rate_e9_view(
+                            &profile,
+                            &group,
+                            asset_index,
+                            shutdown_price,
+                        )?;
+                        let asset = group.markets[asset_index].engine.asset;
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Meta reachability review: the payout divergence is a real accounting defect, but it is not a public LoF path in `percolator-meta`'s managed lifecycle. Controller-approved secondary activation is signed by the controller PDA, so that PDA becomes `asset_admin`; market donation rejects any live creator-administered secondary asset and requires direct permissionless append to be disabled. Therefore an unaffiliated caller cannot invoke this shutdown after handoff. Retitled as privileged LoF. The upstream fix remains worthwhile as defense in depth and for raw Percolator deployments that intentionally delegate asset admin.

aeyakovenko:
Added commit `34e40355` with a second public-interface TDD consequence of the same shutdown omission: an independent honest oracle authenticates a 100 -> 200 target, then a compromised non-oracle admin shuts down before crank. Exact-parent behavior freezes at 100 and pays 10,000/10,000; this PR freezes at the circuit-breaker-bounded 105 and matches the crank-first 10,050/9,950 payouts.

Reverified the combined branch with its exact historical matcher artifacts: `cargo test --all-targets -- --test-threads=1` passes 6 unit + 567 LiteSVM/BPF/CU tests; formatting and diff checks pass.

aeyakovenko:

PR #255, opened 2026-07-18
[BLOCKER LoF] Require committed accrual before market resolve

aeyakovenko:
## Finding

Terminal resolution could freeze value before the public accrual route committed it.

1. A compromised non-oracle market authority could call public `ResolveMarket` after an honest AuthMark checkpoint had activated but before its next deterministic funding segment accrued.
2. Any caller could invoke `ResolveStalePermissionless` while an authenticated mark target was pending, causing resolution at the old effective mark and permanently discarding the move.

Both exact-parent LiteSVM repros use independent long and short owners plus a separate honest oracle. The funding repro transferred 100,000 atoms away from the owed long. The pending-mark repro moves an AuthMark from 1,000,000 to 1,010,000: the control pays 2,010,000,000 / 1,990,000,000, while resolve-first paid 2,000,000,000 to each side, transferring 10,000,000 atoms from the victim long to the attacker short. Both attacks use only public instructions and no injected protocol state.

## Fix

- Use one shared predicate for value-bearing accrual that must precede privileged or stale permissionless resolution.
- Require stale permissionless resolution to commit a pending authenticated mark before terminalizing the asset.
- Preserve current privileged-resolution mark policy here; PR #262 owns that separate policy change.
- At stale-resolution maturity, allow only zero-oracle-account catch-up of the exact stored transition required by the predicate.
- Skip account action/liquidation dispatch on this settlement-only route, and never reset the stale clock.
- Keep the scan cheap for empty assets by filtering exposure before decoding an oracle profile.
- Short-circuit an active mark equal to the engine price: both clamped price movement and premium funding are exactly zero, so no wide arithmetic is needed.

## TDD and liveness

- On exact parent `da64be63`, stale resolution lands before committed funding and before the pending authenticated mark, proving both value transfers.
- On the initial guard-only fix, stale resolution rejects with `EngineStale` but the catch-up crank rejects with `OracleStale`; the test exposes that persistent DoS.
- At fixed head, each unsafe resolve rolls back, one public stored-state crank commits the pending transition, retry resolves, and payouts match the control.
- A public CU regression pre-sizes a normal 4,000-asset market, activates/configures every asset, and opens balanced OI through `BatchTradeNoCpi`. At stale maturity the victim exit is `OracleStale`; the sole resolver remains below the transaction limit.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 570/570 LiteSVM/CU tests and 6/6 host tests green
- 10MiB public resolve reaches exposed asset 5,833 in a 10,484,888-byte market at 79,692 CU

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8683,6 +8683,12 @@ pub mod processor {
         {
             return Ok(None);
         }
+        // With an active mark equal to the engine price, the clamped next price is unchanged and
+        // premium funding is exactly zero. Avoid two wide divisions per exposed asset; resolution
+        // must remain bounded for dense, normally constructed markets.
+        if profile.mark_ewma_e6 == asset.effective_price.get() {
+            return Ok(None);
+        }
         let next_price =
             committed_effective_price_for_accrual_view(profile, group, asset_index, resolved_slot)?;
         let funding_rate_e9 =
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8661,6 +8661,39 @@ pub mod processor {
         Ok(())
     }
 
+    fn committed_market_resolve_accrual_for_profile_view(
+        profile: &state::AssetOracleProfileV16,
+        group: &state::MarketViewMutV16<'_>,
+        asset_index: usize,
+        resolved_slot: u64,
+    ) -> Result<Option<(u64, i128)>, ProgramError> {
+        if asset_index >= group.header.config.max_market_slots.get() as usize
+            || asset_index >= group.markets.len()
+        {
+            return Err(PercolatorError::InvalidInstruction.into());
+        }
+        let asset = group.markets[asset_index].engine.asset;
+        let exposed = asset.oi_eff_long_q.get() != 0 || asset.oi_eff_short_q.get() != 0;
+        // A mark newer than the asset cursor remains prospective. Resolution and its catch-up
+        // crank must agree on this boundary so a stale caller cannot activate a fresh report.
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent public-interface red/green confirms the stale pending-mark LoF on the exact Meta production pin.

Red: wrapper `da64be639168b496833dd4c701607a94fcb06b6c` / engine `143e68c4917ed0400a27b952f036a5677047cd84`. Two independent owners deposited 2,000,000,000 atoms each and opened balanced OI at 1,000,000. A separate honest oracle committed an AuthMark at 1,010,000. The control path cranked before stale maturity and terminal payouts were 2,010,000,000 / 1,990,000,000. An unaffiliated caller instead invoked public `ResolveStalePermissionless` before the crank; it succeeded and terminal payouts were 2,000,000,000 / 2,000,000, transferring 10,000,000 atoms from the long to the short. The sequence uses only normal public initialization/instructions and zero protocol-state injection.

Green: built wrapper head `5f1789a157dd06a27ecc4c966dcfa6a97d164d6a` (PR #269, containing this PR) against the same engine and ran the identical test. Premature resolution rejects, one bounded stored-state crank commits the mark, retry resolves, and payouts match the 2,010,000,000 / 1,990,000,000 control.

Static review found no added signer/custody authority: the change preserves the existing public crank and resolve routes while requiring committed value-bearing accrual. This is REAL public LoF on the current pin and should remain blocked until this stack merges and Meta repins.

aeyakovenko:

PR #256, opened 2026-07-18
[HARDENING LoF] Prevent live fee hikes from taxing user capital

aeyakovenko:
## Finding

`UpdateTradeFeePolicy` allowed the asset-0 insurance authority, which defaults to `marketauth`, to raise the mandatory base fee while unrelated users already had capital and open positions in a live market. A compromised non-oracle admin could set the fee to 10,000 bps, force both users to surrender principal merely to flatten, and withdraw those fees through the public live insurance interface.

The exact-parent LiteSVM repro on `da64be639168b496833dd4c701607a94fcb06b6c` opens a balanced position between two independently owned portfolios, raises the live fee from 0 to 10,000 bps, closes the position, and extracts 2,000 collateral atoms as the configured operator.

## Fix

- Reject base-fee increases while `c_tot != 0`, keeping the terms no worse for as long as any user capital is entrusted to the market.
- Continue to permit fee reductions while capital is deposited.
- Permit increases again after all capital is withdrawn.
- Use the engine-maintained constant-time `c_tot` aggregate instead of scanning assets.

## Verification

- Exact-parent public LiteSVM test fails with a 2,000-atom extraction.
- Fixed public LiteSVM test passes and verifies atomic rejection, unchanged user principal, live fee reductions, and post-withdraw reconfiguration.
- Maximal 5,834-asset / near-10 MiB market update consumes 2,005 CU.
- `cargo test --all-targets`: 567/567 LiteSVM tests plus all host/example targets pass after building the authenticated and hostile matcher fixtures.

No engine change is required.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9657,21 +9657,22 @@ pub mod processor {
         expect_signer(authority)?;
         expect_writable(market_ai)?;
         expect_owner(market_ai, program_id)?;
-        let (mut cfg, asset0_insurance_authority, max_trading_fee_bps) = {
-            let market_data = market_ai.try_borrow_data()?;
-            let (cfg, _, _, _, max_trading_fee_bps) =
-                state::read_market_trade_preflight(&market_data, 0)?;
-            let profile0 = read_oracle_profile_for_asset(&market_data, &cfg, 0)?;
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #257, opened 2026-07-18
[HARDENING DoS] Keep funded Recovery deadlines monotonic

aeyakovenko:
## Impact

A non-oracle market authority could extend `force_close_delay_slots` after an asset entered Recovery and postpone an already-funded winner claim for up to 10,000,000 slots.

The exact-parent LiteSVM reproduction uses only public instructions and separate oracle/admin keys:

- honest oracle moves the mark from 100 to 110;
- two funded users open the position;
- the admin shuts the asset down under a 5-slot force-close window;
- immediately before maturity, the admin extends the live policy to the maximum;
- an honest cranker can no longer force-close, and `ConvertReleasedPnl` cannot release the winner's 10,000-atom claim.

The owner can forfeit the Recovery leg and withdraw principal, so this is a partial payout DoS rather than a total account freeze or principal LoF.

## Fix

Once a nonzero force-close policy exists, reject delay increases while `c_tot != 0`. First-time configuration, shorter windows while capital is live, and longer windows after all capital exits remain supported.

The check reads the constant-time market aggregate and does not scan assets.

## Validation

- exact-parent public LiteSVM red test, fixed green
- controls for first configuration, delay reduction, and post-exit increase
- `cargo test --all-targets`: 567/567 runtime tests plus all host/example targets
- 10 MiB / 5,834-asset `ConfigurePermissionlessResolve`: 1,940 CU

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9733,18 +9733,25 @@ pub mod processor {
         {
             return Err(PercolatorError::InvalidInstruction.into());
         }
-        let (mut cfg, mode, _, _) =
-            state::read_market_config_mode_and_capacity(&market_ai.try_borrow_data()?)?;
+        let mut market_data = market_ai.try_borrow_mut_data()?;
+        let (mut cfg, group) = state::market_view_mut(&mut market_data)?;
         expect_live_authority(&cfg.marketauth, admin.key)?;
-        if mode != MarketModeV16::Live {
+        if group.header.mode != 0 {
             return Err(PercolatorError::EngineLockActive.into());
         }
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_privileged_policy_boundary_matrix_rejects_untrusted_callers`. Only the configured market authority can rewrite the recovery delay; an untrusted public caller rejects with exact rollback. The executable disposition remains in PR #135; no separate production fix is required.

PR #258, opened 2026-07-18
[PRIVILEGED LoF] Freeze funded hard-stale deadlines

aeyakovenko:
## Impact

A compromised non-oracle market authority could raise `permissionless_resolve_stale_slots` immediately before the configured hard-stale boundary and keep stale price-taking live for up to 6,480,000 slots.

The exact-parent LiteSVM reproduction uses only public instructions and distinct roles:

- users deposit under a five-slot hard-stale policy;
- an independent LP owner authorizes a matcher that fills at the wrapper-supplied oracle mark;
- the separate honest oracle stops updating after price 100;
- at slot 5, the non-oracle admin raises the hard-stale policy to its maximum;
- at the original slot-6 deadline, an attacker opens an unsigned `TradeCpi` long against the passive LP at stale price 100;
- the honest oracle resumes at 200, the attacker flattens, converts the PnL, and withdraws 1,001,000 atoms while the LP can withdraw only 999,000.

Without the live extension, the same fill rejects atomically at slot 6 and `ResolveStalePermissionless` pays both users their original deposits.

## Fix

Once a nonzero hard-stale policy exists, reject deadline increases while `c_tot != 0`. First configuration, shorter deadlines while capital is live, idempotent writes, and longer deadlines on an empty live market remain supported.

The guard reads the constant-time market aggregate and does not scan assets or block trading/crank progress.

Deadline shortening is intentionally not claimed as a separate finding here. Its earlier payout divergence depended on the pending-mark omission already owned by PR #255; composing #255 with this PR leaves shortening enabled, forces stale resolution to reject at the old price, and permits one bounded public catch-up crank before resolution.

## TDD evidence

- exact-parent public LiteSVM RED: `7aad0c9d` completes the 1,000-atom passive-LP extraction
- fixed head: `191efc10`
- current focused recheck: exploit regression and stale non-base/global-resolution regression green against the real SBF
- patch-equivalent full verification: 6 wrapper unit tests and 566 serialized LiteSVM/CU tests green
- `cargo fmt --all -- --check` and `git diff --check` green

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9733,18 +9733,25 @@ pub mod processor {
         {
             return Err(PercolatorError::InvalidInstruction.into());
         }
-        let (mut cfg, mode, _, _) =
-            state::read_market_config_mode_and_capacity(&market_ai.try_borrow_data()?)?;
+        let mut market_data = market_ai.try_borrow_mut_data()?;
+        let (mut cfg, group) = state::market_view_mut(&mut market_data)?;
         expect_live_authority(&cfg.marketauth, admin.key)?;
-        if mode != MarketModeV16::Live {
+        if group.header.mode != 0 {
             return Err(PercolatorError::EngineLockActive.into());
         }
         if oracle_v16::permissionless_stale_matured(&cfg, authenticated_slot_or_fallback(0)) {
             return Err(PercolatorError::OracleStale.into());
         }
+        if cfg.permissionless_resolve_stale_slots != 0
+            && stale_slots > cfg.permissionless_resolve_stale_slots
+            && group.header.c_tot.get() != 0
+        {
+            return Err(PercolatorError::EngineLockActive.into());
+        }
         cfg.permissionless_resolve_stale_slots = stale_slots;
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Meta reachability review: this is a real bounded-admin extraction in raw Percolator, but not a public LoF through `percolator-meta`. The Meta controller's `validate_permissionless_resolve_update` rejects any increase after the first nonzero value for both `stale_slots` and `force_close_delay_slots`, and direct calls cannot produce the controller PDA signature. The existing real-SBF `e2e_controller_cannot_extend_funded_resolution_deadlines` covers funded users, atomic rejection, shorter-policy progress, and direct-forgery failure. Retitled privileged; no duplicate probe retained.

aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_privileged_policy_boundary_matrix_rejects_untrusted_callers`. Only the configured market authority can rewrite the hard-stale deadline; an untrusted public caller rejects with exact rollback. The executable disposition remains in PR #135; no separate production fix is required.

PR #259, opened 2026-07-18
[PRIVILEGED LoF] Bind direct trades to backing-fee consent

aeyakovenko:
## Impact

A permissionless backing authority could raise a live source-backing fee after both owners signed a `TradeNoCpi` transaction. Landing the unchanged transaction then charged an independent LP at the new rate, credited the attacker-controlled backing ledger, and made the victim-funded fee immediately withdrawable.

The public LiteSVM regression uses normal market, portfolio, position, policy-update, trade, and withdrawal instructions. It does not mutate account state directly. The exploit extracts exactly 70 base units from the independent LP in the test geometry.

## Root cause

`TradeNoCpi` authorized the trade fee but did not bind either signer to the mutable source-backing fee policy evaluated at execution time. This made an already-signed transaction vulnerable to an out-of-order policy update.

## Fix

- Add a signer-authorized `max_backing_fee_bps` cap to `TradeNoCpi`.
- Decode legacy 35-byte payloads with an implicit zero cap, so existing signatures cannot acquire new fee consent.
- Reject a positive source-backing charge above the signed cap while preserving zero-fee and risk-reducing progress.
- Prove both legacy and extended wire decodes with Kani.

## TDD evidence

Exact parent `da64be639168b496833dd4c701607a94fcb06b6c` fails `v16_attack_presigned_trade_nocpi_cannot_be_rugged_by_backing_fee_update` because the zero-cap pre-signed transaction succeeds after the fee hike. Commit `2b317e55` rejects that stale transaction with byte-identical rollback, accepts a newly signed capped transaction, charges exactly 70 units, and lets the attacker withdraw the fee.

## Verification

- `cargo test --lib` (6/6)
- `cargo test --test v16_cu -- --test-threads=1` (566/566)
- `cargo build --examples`
- targeted Kani wire proofs (all green)
- `cargo fmt --check`
- `git diff --check`

`TradeCpi` has a distinct matcher-consent surface tracked separately; batch direct trading currently rejects configured backing fees.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2490,6 +2490,10 @@ pub mod ix {
             size_q: i128,
             exec_price: u64,
             fee_bps: u64,
+            /// Highest source-backing policy rate both owners authorize. Legacy payloads omit
+            /// this field and decode to zero, so a pre-existing signed transaction cannot be
+            /// charged a backing fee that was installed after authorization.
+            max_backing_fee_bps: u16,
         },
         TradeCpi {
             asset_index: u16,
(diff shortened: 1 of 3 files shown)


aeyakovenko:

On exact parent `da64be639168b496833dd4c701607a94fcb06b6c`, I constructed a normal two-asset market through public instructions, let an external provider fund source backing, and had both owners pre-sign the legacy `TradeNoCpi` wire while its backing fee was zero. The provider then changed only its own live domain policy to 5,000 bps and landed the unchanged transaction. Result:

- victim capital debit: exactly `70` atoms
- provider backing earnings credit: exactly `70` atoms
- provider subsequently withdrew all `70` through the real backing ledger

No protocol state was injected; only initially zero program-owned market/portfolio/ledger accounts and external token fixtures were supplied before normal initialization.

I then built immutable head `2b317e55a9c9ae7a928d1c9285c4b6e2d4afc613` and ran the unchanged probe with fixed expectations. The stale legacy signature rejects with byte-identical market and both portfolio accounts. A fresh raw wire capped at 5,000 bps remains live, charges exactly `70`, and the provider can withdraw exactly `70`. The diff adds signer fee consent only and no authority/custody surface. Confirmed **REAL public LoF**.

aeyakovenko:

PR #260, opened 2026-07-18
[BLOCKER LoF] Prevent pending EWMA PnL inheritance

aeyakovenko:
## Impact

A public trade can pay a small fee to queue a trade-driven EWMA move while open interest is tiny. Before any crank applies that move, a much larger trade can enter at the old engine effective price. The next crank applies the already-paid pending mark to the new open interest, transferring PnL from an independent counterparty to the attacker.

On exact parent `da64be639168b496833dd4c701607a94fcb06b6c`, the public LiteSVM reproducer pays 10,100 atoms to seed the move, then creates and withdraws a 5,000,000 atom gain funded by the victim LP. The exploit reproduces through `TradeNoCpi`, `BatchTradeNoCpi`, `TradeCpi`, and `BatchTradeCpi` with normal account construction and no state injection.

## Root cause

The wrapper updated `mark_ewma_e6` after a trade, while the engine effective price remained unchanged until a later crank. New exposure was admitted between those two transitions at the stale effective price. The reported-price path also synthesized `dt = 1` after a same-slot crank had consumed the real engine dt, allowing a trade to queue mark movement for a future slot.

## Fix

- Apply a pending trade-driven EWMA or stale-hybrid mark through the normal bounded engine accrual before admitting new exposure.
- Perform that accrual before CPI matcher quote and limit binding, so the matcher and user limit see the post-accrual effective price.
- Use the exact per-asset engine dt when clamping reported trade prices. Same-slot trades still execute, but cannot consume a synthetic future slot.
- Keep every trade route live rather than rejecting stale pre-signed transactions.

## TDD coverage

- Exact-parent red and fixed-head green extraction tests for direct and matcher routes, single and batch.
- CPI stale-limit rollback and post-mark limit acceptance.
- Same-slot zero-engine-dt trade execution without future mark movement.
- Public max-shape 14-asset pending-mark batches: 1,124,115 CU direct and 1,155,718 CU through the matcher, both below 1.4M.

## Verification

- `cargo test --test v16_cu -- --test-threads=1`: 570 passed
- `cargo build --examples`
- `cargo fmt --check`
- `git diff --check`

No instruction encoding, account layout, or engine dependency changes.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -5828,6 +5828,11 @@ pub mod processor {
             } else {
                 size_q.unsigned_abs()
             };
+            apply_pending_trade_mark_before_trade_view(
+                &oracle_profile,
+                &mut group,
+                asset_index as usize,
+            )?;
             // F-TRADENOCPI-FEE / F-NOCPI-MARK-FEE: request.exec_price is used by the engine only as
             // fee notional. In price-managed EWMA/stale-hybrid modes, the caller's reported print is
             // also the mark-discovery input, so first normalize it to the same per-asset dt price
@@ -6083,6 +6088,11 @@ pub mod processor {
                     return Err(PercolatorError::InvalidInstruction.into());
                 }
                 let abs_size = leg.size_q.unsigned_abs();
+                apply_pending_trade_mark_before_trade_view(
+                    &oracle_profile,
+                    &mut group,
+                    asset_index,
+                )?;
                 let fee_basis_price = accepted_reported_trade_price_view(
                     &oracle_profile,
                     &group,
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent clean-room confirmation from `percolator-meta` against real SBF in LiteSVM for the public direct `TradeNoCpi` route.

On exact parent `da64be639168b496833dd4c701607a94fcb06b6c`, a one-unit seed pair paid `10,100` atoms to queue a trade-driven EWMA move while engine effective price remained `1,000,000`. An already-signed 1,000-unit attacker/victim trade then landed at that stale cursor. Public cranks applied `1,005,000`; closing and converting produced:

- independent victim capital loss: `5,000,000`
- attacker converted gain: `5,000,000`
- attacker net withdrawal after seed cost: `4,989,900`

The probe used only normal public init, backing top-up, portfolio, deposit, trade, crank, conversion, and withdrawal instructions. No protocol state was injected; only initially zero program-owned market/portfolio accounts and external token fixtures preceded normal initialization.

I built immutable head `ba63f0e16b0ab61943f14bed0c77d9c14fda4037` and ran the unchanged probe with fixed expectations. The same pre-signed large trade remains live, but the pending `1,005,000` mark applies before its OI enters; victim loss, attacker gain, and net withdrawal are all exactly zero. No instruction/account layout or authority/custody surface is added. Confirmed **REAL public LoF** for `TradeNoCpi`; I did not independently reproduce the PR's matcher/batch variants.

aeyakovenko:

PR #261, opened 2026-07-18
[HARDENING DoS] Reject expired Hybrid terminal snapshots

aeyakovenko:
## What changed

- reject privileged `ResolveMarket` while any exposed Active/DrainOnly Hybrid asset has neither a current accepted external report nor a same-slot fee-backed after-hours EWMA
- apply the same asset-scoped gate before `UpdateAssetLifecycle::SHUTDOWN` freezes a Hybrid asset in Recovery
- require every configured composite-oracle leg to be current; a newer component cannot mask an expired component through the aggregate max publish time
- keep `ResolveStalePermissionless` unchanged as the configured outage escape
- add a public LiteSVM TDD regression with independent long and short portfolios and an independently controlled oracle authority

## Root cause and impact

The privileged market-resolution and asset-shutdown paths called engine terminal-snapshot transitions without enforcing the Hybrid profile's configured external-report freshness. A compromised non-oracle market authority could therefore resolve a market, or freeze an asset in Recovery, against an expired stored mark after an honest current report had been published but before a keeper ingested it.

On the exact pre-fix parent, each probe settles the victim/attacker at `(100_000_000, 100_000_000)` instead of the honest-report control `(110_000_000, 90_000_000)`, depriving the independent long of `10_000_000` units. The shutdown probe completes through delayed permissionless force-close, public PnL conversion, and ordinary owner withdrawals. The attacks use only public instructions and `marketauth`; neither signs an oracle instruction nor mutates protocol state directly.

## TDD evidence

- exact-parent resolve red: `bd39a4c0` test commit over `da64be63`; the regression fails with the payout divergence above
- post-first-fix shutdown red: `77ef3d69`; the independent shutdown regression fails with the same payout divergence
- composite-leg red: `8a75895d`; one current leg masks one expired leg and reproduces the same payout divergence against the first fix
- fixed green: `398c5cac`

## Validation

- `cargo test --test v16_cu -- --nocapture` (`567 passed`)
- `cargo test --lib --examples` (`6 passed`)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8687,10 +8687,16 @@ pub mod processor {
             return Ok(());
         }
 
-        let external_fresh = profile.oracle_target_publish_time > 0
-            && now_unix_ts >= profile.oracle_target_publish_time
-            && (now_unix_ts - profile.oracle_target_publish_time) as u64
-                <= profile.max_staleness_secs;
+        let oracle_leg_count = profile.oracle_leg_count as usize;
+        let external_fresh = oracle_leg_count != 0
+            && oracle_leg_count <= constants::ORACLE_LEG_CAP
+            && profile.oracle_leg_publish_times[..oracle_leg_count]
+                .iter()
+                .all(|publish_time| {
+                    *publish_time > 0
+                        && now_unix_ts >= *publish_time
+                        && (now_unix_ts - *publish_time) as u64 <= profile.max_staleness_secs
+                });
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-020 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #262, opened 2026-07-18
[REAL LoF] Commit authenticated marks before terminal resolve

aeyakovenko:
## Finding

A compromised non-oracle `marketauth` could call public `ResolveMarket` after an independent honest oracle had pushed a new AuthMark or EWMA target, but before a keeper advanced the engine effective-price cursor. Resolution terminally discarded the authenticated target and settled every open position at the older price.

The exact-parent public LiteSVM repro on `da64be639168b496833dd4c701607a94fcb06b6c` opens balanced positions between independent owners, separates the oracle authority from `marketauth`, pushes an honest target from 1,000,000 to 2,000,000, and resolves immediately. The honest crank-first control advances the circuit-breaker-bounded price to 1,002,400 and pays long/short `(102,400,000, 97,600,000)`. The attack freezes 1,000,000 and pays `(100,000,000, 100,000,000)`, depriving the independent long of 2,400,000 atoms while the attacker avoids the matching loss.

This also exposed a pre-existing test contradiction: `v16_regression_resolve_before_settlement_uses_official_price` described stale pre-move settlement as the attack in its headline, then asserted that exact stale settlement as correct in its body.

The signerless stale-resolution sibling and its required post-deadline rescue path are already owned by PR #255; they are intentionally not duplicated here.

## Fix

- Before privileged resolution, scan exposed Active/DrainOnly assets for a price-managed target with a real engine-bounded price step available at the resolve slot.
- Reject with `EngineStale` until one ordinary permissionless crank commits that step.
- Leave flat assets, zero-delta targets, same-slot circuit-breaker fixed points, and permissionless stale resolution unchanged.
- Reconcile the contradictory older regression with the no-omission invariant.

## TDD evidence

- Exact-parent test commit `ccbb94fb` is red with the payout divergence above.
- Fixed head covers both AuthMark and EWMA and proves rejected resolution rolls back, one public crank succeeds, retry resolution succeeds, and both payouts equal the crank-first control.
- A 5,834-asset / 10 MiB market with exposure at the highest slot resolves in 91,030 CU with the new scan.

## Verification

- `cargo test --test v16_cu`: 566/566
- `cargo test --lib --examples`: 6/6 library tests plus examples green


aeyakovenko:
Closing after invariant audit: this reverses the merged anti-retroactivity requirement. An unaccrued pending mark is not the official effective price and must not be retroactively applied by a terminal route; tests/v16_cu.rs::v16_regression_resolve_before_settlement_uses_official_price is the standing regression.

PR #263, opened 2026-07-18
[REAL LoF] Settle adverse marks before unilateral reduction

aeyakovenko:
## Finding

`RebalanceReduce` is an owner-callable unilateral exit at the engine's current effective price. The wrapper did not bind that operation to an authenticated AuthMark/EWMA target queued in the wrapper profile. A losing side could therefore reduce to flat and withdraw before the bounded mark step was committed, permanently removing the open interest that should have received the corresponding PnL transfer.

The exact-parent LiteSVM reproduction at `da64be63` uses a separately authorized honest oracle and only public instructions. With a 1,000,000 -> 2,000,000 authenticated mark:

- crank first: effective price 1,002,400, short withdraws 97,600,000, long receives 102,400,000
- reduce first: short withdrew 100,000,000, then the same honest mark could no longer transfer PnL; long received 100,000,000

This moved 2,400,000 from an independent victim to the reducing attacker. Commit `e9d34a54` is the exact-parent red test.

## Fix

Before unilateral `RebalanceReduce`, derive the next engine-admissible bounded price from the authenticated wrapper target (or the committed Hybrid target). Reject only when that currently available step is adverse to the reducing leg. The owner can permissionlessly crank once and retry; bilateral trade exits and favorable unilateral reductions remain available.

The regression asserts rejected-call rollback byte-for-byte, then takes the bounded crank continuation and proves payouts equal the crank-first control. A separate positive test proves that a long can still reduce immediately against a queued favorable upward mark.

## Verification

- `cargo test --lib --examples`: 6 passed, examples passed
- Existing `RebalanceReduce` custody CU budget test passes

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4555,6 +4555,72 @@ pub mod processor {
         Ok(())
     }
 
+    fn reject_rebalance_reduce_with_adverse_pending_mark_view(
+        cfg: &WrapperConfigV16,
+        group: &state::MarketViewMutV16<'_>,
+        portfolio: &percolator::PortfolioV16ViewMut<'_>,
+        asset_index: usize,
+    ) -> Result<(), V16Error> {
+        if asset_index >= group.header.config.max_market_slots.get() as usize
+            || asset_index >= group.markets.len()
+        {
+            return Err(V16Error::InvalidConfig);
+        }
+        let leg =
+            active_leg_for_asset_view(portfolio, asset_index).map_err(|_| V16Error::InvalidLeg)?;
+        let asset = group.markets[asset_index].engine.asset;
+        let profile = read_oracle_profile_from_view(group, cfg, asset_index)
+            .map_err(|_| V16Error::InvalidConfig)?;
+        if !oracle_v16::profile_is_price_managed(&profile) {
+            return Ok(());
+        }
+
+        let now_slot = authenticated_market_slot_or_fallback_view(group);
+        let slot_last = asset.slot_last.get();
+        if now_slot < slot_last {
+            return Err(V16Error::Stale);
+        }
+        let dt = core::cmp::min(
+            now_slot - slot_last,
+            group.header.config.max_accrual_dt_slots.get(),
+        );
+        if dt == 0 {
+            return Ok(());
+        }
+
+        // Auth/EWMA profiles own their queued target. During normal Hybrid operation the
+        // committed engine target is the latest authenticated external observation; after the
+        // soft-stale boundary the profile EWMA becomes the target used by the crank instead.
+        let target = if oracle_v16::profile_is_auth_mark(&profile)
+            || oracle_v16::profile_is_ewma_mark(&profile)
+            || (oracle_v16::profile_is_hybrid(&profile)
+                && oracle_v16::profile_hybrid_soft_stale_matured(&profile, now_slot))
+        {
+            profile.mark_ewma_e6
+        } else {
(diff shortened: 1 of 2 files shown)


aeyakovenko:

Exact parent `da64be639168b496833dd4c701607a94fcb06b6c`, separate honest AuthMark oracle, normal public init/portfolio/deposit/trade/push/reduce/withdraw/crank/resolve/payout instructions:

- crank-first control: effective `1,002,400`, short withdrawal `97,600,000`, independent long payout `102,400,000`
- reduce-first attack: stale unilateral reduction lands, short withdrawal `100,000,000`, independent long payout `100,000,000`
- exact attacker gain / victim loss: `2,400,000` atoms

After the short disappears, the unopposed mark reaches the full `2,000,000` target, but no longer transfers the omitted PnL. No protocol state was injected; only initially zero program-owned market/portfolio accounts and external token fixtures preceded normal initialization.

I then built immutable head `90219b488597c6c5971c6071cfd4158fe240f965` and ran the unchanged probe with fixed expectations. The first adverse reduction rejects with byte-identical market, attacker, and victim accounts; one bounded permissionless crank makes retry succeed; effective price and both payouts exactly equal control. A separate fresh-market case proves the favorable long still reduces immediately against the same queued upward mark. No authority or custody surface is added. Confirmed **REAL public LoF**.

aeyakovenko:
Independent clean-room LiteSVM found a missing public reduction surface. The matcher/direct-trade fix is correct for the submitted routes, but `RebalanceReduce` can still erase the same deterministic zero-move funding interval.

Fresh probe: `v16_probe_rebalance_partial_reduce_cannot_erase_elapsed_premium_funding`.

Public sequence:
- effective price `2`, honest AuthMark target `1`, 24 bps/slot cap, one-slot accrual;
- attacker opens a 100-unit short through the standard pre-authorized matcher; the independent long LP is passive and only the attacker signs the trade;
- slot 1 cranks publish the honest target without retroactive funding;
- at slot 2 the attacker alone calls `RebalanceReduce` for half its short before the honest crank, then the crank lands and the attacker reduces the remainder;
- attacker withdraws normally; the passive LP is paid through normal market resolution.

Exact value result (2,000,000 total atoms conserved):
- crank first: attacker `999900`, LP `1000100`;
- half-reduce first: attacker `999950`, LP `1000050`.

The attacker extracts 50 atoms from the independent LP. This reproduces on:
- exact parent `da64be639168b496833dd4c701607a94fcb06b6c`;
- exact #271 head `f52b7fbcbc42219cd7dc19a1c21b9499d6fd47a8`;
- a clean SBF composition of #263 production commit `90219b488597c6c5971c6071cfd4158fe240f965` plus #271 production commit `f52b7fbcbc42219cd7dc19a1c21b9499d6fd47a8`.

The composed result matters: #263's bounded adverse-price check sees an unchanged effective price, while #271 calls `accrue_zero_move_funding_before_trade_view` only from Trade/BatchTrade CPI and no-CPI paths. `handle_rebalance_reduce` still mutates OI directly without committing the funding segment.

Please extend #271's invariant to `RebalanceReduce` (and retain a public half-reduction regression). This is the same root cause and should be fixed in this PR before merge rather than split into another overlapping branch.

aeyakovenko:
Closing after invariant audit: this reintroduces retroactive pricing by forcing an unaccrued adverse mark into a unilateral reduction. Pending publication is not official accrued price; applying it here contradicts the merged anti-retroactivity invariant rather than fixing a public LoF.

PR #264, opened 2026-07-18
[BLOCKER LoF] Bind AuthMark targets before exposure migration

aeyakovenko:
## Finding

An authenticated `PushAuthMark` updated only the wrapper oracle profile. Until a later crank copied that target into the engine, engine health checks still certified accounts at the old target.

A losing trader could use only public instructions to move an adverse short from a well-capitalized portfolio into a thin sacrificial portfolio after the honest mark was signed, withdraw the original portfolio's capital, and leave an independent long underpaid when the mark later converged.

On the exact pre-fix parent, the LiteSVM repro produces:

| path | attacker recovery | independent victim payout |
| --- | ---: | ---: |
| no migration | 1,060,038,000 | 1,199,962,000 |
| migrated exposure | 1,100,000,000 | 1,160,000,000 |

That is an exact 39,962,000-atom transfer from the victim to the attacker. The market uses production risk parameters, a separate honest oracle signer, public portfolio/deposit/trade/withdraw/crank/resolve/close instructions, and no state injection.

## Fix

`PushAuthMark` now publishes the authenticated mark to the engine raw target in the same instruction. The target update increments the oracle epoch, invalidates stale certificates, and makes the engine's full adverse-target margin penalty apply before any subsequent exposure change.

The migration then fails final IM and LiteSVM proves byte-exact rollback for the market, both attacker portfolios, and the independent victim. Adequately collateralized trades during target convergence remain live through the dependent engine change in [percolator#138](https://github.com/aeyakovenko/percolator/pull/138); target lag is priced by final margin rather than blanket-blocked.

## TDD history

- `a36c7530`: exact-parent red public LiteSVM repro
- `f7a64eb0`: wrapper fix plus engine dependency; regression green

## Validation

- targeted public repro: green, attack and control payouts identical
- both target-lag positive-path regressions: green
- `cargo test --test v16_cu`: 566/566 green, including all CU tests
- `cargo test --all-targets`: 572/572 green (6 host + 566 LiteSVM)
- engine dependency: 130 native/spec/fuzz tests and four directly affected Kani proofs/contracts green in percolator#138

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=2af64e487f3dec88b18aac634071115e588d2af1#2af64e487f3dec88b18aac634071115e588d2af1"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "2af64e487f3dec88b18aac634071115e588d2af1" }
 anchor-lang-v2 = { git = "https://github.com/solana-foundation/anchor", rev = "e3cf760826fa0a60f247635ea0572e59861ec9b5", package = "anchor-lang-v2", default-features = false, features = ["alloc", "guardrails"], optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
 # Host tests use the same zero-copy account/view API as the SBF/program target.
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Independent public LiteSVM red/green confirms this as REAL LoF on Meta's exact production pin.

Red: wrapper `da64be639168b496833dd4c701607a94fcb06b6c` / engine `143e68c4917ed0400a27b952f036a5677047cd84`. A separately authorized honest oracle committed AuthMark 1,000,000 -> 1,100,000 after an independent victim long opened against the attacker's funded short. The attacker then used public `TradeNoCpi` between two attacker-owned portfolios to move the short into a normally initialized 60,000,000-atom portfolio and withdrew all 1,100,000,000 atoms from the funded portfolio. Bounded fresh-slot cranks converged the mark, followed by public resolution, closes, and payout topups.

Control (no migration): victim 1,200,000,000; attacker 1,060,000,000; canonical insurance 100,000,000.
Attack: victim 1,160,000,000; attacker 1,100,000,000; canonical insurance 60,000,000.

The attacker extracts exactly 40,000,000 atoms that the independent victim loses. The path uses normal public initialization/instructions and no injected protocol state.

Green: built this exact wrapper head `f7a64eb0759b3ec58396cadfdc6793ff2c896ed2` with engine `2af64e487f3dec88b18aac634071115e588d2af1` and ran the identical probe. Migration rejects byte-atomically across market, both attacker portfolios, and the token vault; terminal outcomes exactly match control: victim 1,200,000,000, attacker 1,060,000,000, insurance 100,000,000.

Static review found no new signer/custody surface. `PushAuthMark` publishes the already authenticated target through the existing engine setter, which increments `oracle_epoch`; final margin then prices the complete adverse target/effective lag.

aeyakovenko:

PR #265, opened 2026-07-18
[BLOCKER LoF] Publish EWMA oracle targets before exposure changes

aeyakovenko:
## Impact

`PushEwmaMark` authenticated and smoothed a new oracle mark into the wrapper profile but did not publish that committed target to the engine. Before the first crank, a losing trader could migrate its exposure from a funded portfolio into a thin sacrificial portfolio at the stale engine target, withdraw the funded portfolio, and leave an independent winner underpaid.

On exact parent `da64be639168b496833dd4c701607a94fcb06b6c`, the public LiteSVM reproducer uses a separately authorized honest oracle and produces:

| path | attacker recovery | independent victim payout |
| --- | ---: | ---: |
| no migration | 1,060,038,000 | 1,199,962,000 |
| migrate after EWMA push | 1,100,000,000 | 1,160,000,000 |

That is an exact 39,962,000-atom transfer from the victim to the attacker. The test uses only public portfolio, deposit, trade, oracle push, withdrawal, crank, resolve, and payout instructions with normal account construction.

## Fix

Publish the smoothed EWMA mark through `set_asset_raw_oracle_target_not_atomic` in the same successful `PushEwmaMark` transition. This invalidates stale certificates and makes the engine's target-lag margin account for the committed move before later exposure changes.

The wrapper pins engine commit `2af64e487f3dec88b18aac634071115e588d2af1` from percolator PR #138 so adequately collateralized trades remain live while target and effective price converge; the lag is priced at final margin instead of blanket-blocking trades.

This is the `PushEwmaMark` sibling of wrapper PR #264. PR #264 changes only `PushAuthMark`; it does not cover this public entrypoint.

## TDD history

- `5d1b37d2`: exact-parent public red reproduction
- `1dafd7de`: target publication fix and engine dependency; regression green

## Verification

- auth matcher and hostile matcher SBF fixture builds
- focused control/attack regression: green, recoveries and payouts identical
- `cargo test --all-targets -- --test-threads=1`: 6 host + 566 LiteSVM/CU tests green
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=2af64e487f3dec88b18aac634071115e588d2af1#2af64e487f3dec88b18aac634071115e588d2af1"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "2af64e487f3dec88b18aac634071115e588d2af1" }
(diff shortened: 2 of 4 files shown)


aeyakovenko:

Exact current pin `da64be639168b496833dd4c701607a94fcb06b6c` (engine `143e68c4917ed0400a27b952f036a5677047cd84`), using only public init/portfolio/deposit/trade/oracle-push/withdraw/crank/resolve/payout instructions and a separately authorized honest EWMA oracle:

- control: attacker `1,060,038,000`, victim `1,199,962,000`
- migrate after `PushEwmaMark`: attacker `1,100,000,000`, victim `1,160,000,000`
- exact attacker gain / victim loss: `39,962,000` atoms

I then built immutable head `1dafd7de1967ab53f171221dae23ba245774918a` (engine `2af64e487f3dec88b18aac634071115e588d2af1`) and ran the unchanged probe with fixed expectations. The migration rejects and both paths produce the control payouts exactly. The wrapper diff only publishes the already-authenticated smoothed target; it adds no authority or custody surface. Confirmed **REAL public LoF**.

aeyakovenko:

PR #267, opened 2026-07-18
[BLOCKER LoF] Keep source backing domain-local

aeyakovenko:
## Impact

A public trader could create positive PnL claims in two asset domains, leave the lower-index claim temporarily unbacked, and fund only the higher-index source domain. Backing consumption selected a funded source domain, but the later aggregate PnL setter retired claim face in portfolio slot order.

Two independent public sequences exploit the mismatch:

- `ConvertReleasedPnl` consumed higher-domain backing while burning lower-domain face, allowing the same provider backing to be consumed twice.
- An adverse mark on the higher-domain position consumed that domain's backing while burning the unrelated lower-domain face, preserving the funded claim for a later capital conversion.

In each sequence, a provider backing a 100-atom domain-local claim can be charged 200 atoms.

## Public TDD

Both LiteSVM regressions use only public instructions: portfolio creation, deposits, trades, authorized mark pushes, permissionless cranks, backing top-up, flattening trades, and `ConvertReleasedPnl`. They do not mutate account state directly.

- `5c70462d` is exact-parent red on wrapper main / engine `143e68c` for the conversion path.
- The adverse-PnL regression was independently run red on the same exact parent: asset 1 settlement erased asset 0's claim while consuming asset 1 backing.
- `7511cb75` pins engine PR #140.
- `81dfda28` adds the adverse-PnL public regression; both paths are fixed-green.
- Successful conversion remains below the existing 300k custody CU limit.

## Dependency

Depends on aeyakovenko/percolator#140.

## Verification

- both focused public LiteSVM regressions
- `cargo test --all-targets`: 6 unit + 567 LiteSVM/CU tests passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=f0e189a149e428b7647a9cc0909315cd80161376#f0e189a149e428b7647a9cc0909315cd80161376"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Independent public-SBF reproduction confirms this wrapper regression's severity.

On wrapper `da64be6` / engine `143e68c`, a winner created 100-atom claims in domains 1 and 3 while only domain 3 held 200 atoms of external backing. After flattening against a fresh replacement and leaving the original loser stale:

- conversion 1 consumed 100 from domain 3 but retired domain 1's unbacked claim
- a same-price open/close recertified the flat account
- conversion 2 consumed another 100 from domain 3
- domain 3 ended with 200 consumed/provider-receivable atoms against its original 100-atom claim
- `Withdraw` transferred the full 200 atoms to the attacker's SPL token account

No protocol bytes were injected or mutated; every state transition used the production public interface. This is a confirmed **REAL LoF**.

aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-031 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #266, opened 2026-07-18
[BLOCKER DoS] Keep exact-OI ADL survivors exit-live

aeyakovenko:
## Summary

- Add an exact-parent public LiteSVM reproduction for the equality boundary left by engine #109's preexisting-OI gate.
- Pin the isolated fix from engine PR #139, stacked on the generic reset cleanup from engine PR #99 / wrapper #201.
- Prove one bounded permissionless crank removes the residual leg and the owner withdraws all remaining principal.

## Public exploit

The test uses only ordinary public instructions and valid initialized accounts:

1. Two users deposit SPL collateral and open opposing positions.
2. Public maintenance and crank calls partially liquidate one side, leaving survivor basis larger than effective OI.
3. A crossed trade reduces effective OI by exactly the remaining amount.
4. Both OI counters become zero, but the survivor retains `1_000_000` basis units and deposited capital in an active leg.
5. Owner reduction returns `CounterUnderflow`, a fresh counterparty cannot fund an exit from zero preexisting OI, and ten honest permissionless cranks return success without changing state.

This is persistent user DoS, not a malformed-state probe, initialization footgun, stale caller hint, or transient failure. It contradicts the selector claim directly: `NoAction` was a successful no-op witness for an account whose funds had no bounded public exit.

## Exact-parent TDD

- Historical exact `main` parent: wrapper `da64be639168b496833dd4c701607a94fcb06b6c`, engine `143e68c4917ed0400a27b952f036a5677047cd84`
- Rebased red commit: `bc790f69`
- Red dependency: engine #99 at `28a7338bd8c0dff7134ea0aa06404795bbb8557b`, before the equality fix
- Red result: OI `(0,0)`, modes `(Normal,Normal)`, residual basis `1_000_000`; owner exit `Custom(25)`; 10/10 cranks return `Ok(58_771)` with identical state
- Fixed commit: `2fe02df1`
- Fixed engine: `619f3fec9a2f6c0cf2462414c47a65eeefdde81c`
- Fixed result: side enters `ResetPending`; one public crank clears the leg at 100,151 CU; owner withdraws all remaining principal

## Scope and dependencies

- Engine fix: https://github.com/aeyakovenko/percolator/pull/139
- Base wrapper: #201, providing engine #99's generic prior-reset cleanup
- Engine #118 / wrapper #219 are not duplicates: they cover bilateral zero-OI after unilateral bankruptcy close; this PR covers the distinct crossed-trade equality route

## Validation

[...]

PR #268, opened 2026-07-18
[BLOCKER DoS] Keep unilateral exact-OI ADL survivors exit-live

aeyakovenko:
## What changed

Adds a public LiteSVM regression for a unilateral exact-effective-OI reduction after ADL has left stored basis above effective OI, and pins the engine fix from aeyakovenko/percolator#141.

## Public exploit and impact

The test uses only normal public instructions on a live initialized market:

1. Deposit and trade to establish open interest.
2. Use public cranks and maintenance sync to reach an ADL survivor whose stored basis exceeds effective OI.
3. The owner calls `RebalanceReduce` for exactly the remaining effective OI.
4. On the exact pre-fix engine, the survivor remains `Normal` with zero OI and nonzero basis.
5. A fresh willing counterparty cannot close it, repeated honest auto-cranks do not clear it, and full withdrawal rejects, permanently stranding the owner's deposited principal.

No account bytes or engine state are injected. The test also cleans the opposite side before retrying with a fresh public counterparty, excluding another account's pending reset as the cause.

## Fix

Engine PR aeyakovenko/percolator#141 starts the existing bounded reset flow for either zero-effective-OI side retaining stored basis after unilateral reductions. An honest public auto-crank can then detach the survivor and restore withdrawal liveness.

This PR is intentionally stacked on #266 because the engine fix reuses the generic exact-OI residue detector introduced by its engine dependency. The diff against that base contains only this public regression and the engine pin.

## TDD evidence

- Exact pre-fix wrapper/engine parent: regression fails after public cranks, fresh trade, and withdrawal all leave the funded leg stuck.
- Fixed engine `b3886dcf6927c8e1cb2b822f0819451971a4b3a5`: regression passes and the owner can withdraw after honest crank cleanup.
- Engine suite: 133/133 passed.
- Wrapper LiteSVM/CU suite: 569/569 passed with the auth and hostile matcher SBF fixtures built.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=619f3fec9a2f6c0cf2462414c47a65eeefdde81c#619f3fec9a2f6c0cf2462414c47a65eeefdde81c"
+source = "git+https://github.com/aeyakovenko/percolator?rev=b3886dcf6927c8e1cb2b822f0819451971a4b3a5#b3886dcf6927c8e1cb2b822f0819451971a4b3a5"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "619f3fec9a2f6c0cf2462414c47a65eeefdde81c" }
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Independent real-SBF validation passes on this exact wrapper head.

Red: Meta's pinned wrapper `da64be639168b496833dd4c701607a94fcb06b6c` / engine `143e68c4917ed0400a27b952f036a5677047cd84` reaches the public exact-OI residue and leaves the funded survivor stuck after fresh-slot cranks, owner reduce, bilateral close, full withdraw, and forfeit probes.

Green: I built this head `b9c6b64e08e4904a64286c1e46bc4eae913add22` with engine `b3886dcf6927c8e1cb2b822f0819451971a4b3a5` and ran the same independently written public sequence. Both zero-OI residue sides enter `ResetPending`, bounded cranks remove both active legs, and both owners withdraw their complete remaining capital.

The pin and public behavior match engine #141. No Meta-side workaround or additional custody authority is needed.

aeyakovenko:

PR #269, opened 2026-07-19
[BLOCKER DoS] Keep max-shape zero-delta stale resolution live

aeyakovenko:
## What changed

- Add a public LiteSVM regression that fills all 5,834 supported market slots with balanced OI, stages a distinct authenticated mark on every asset, and proves stale resolution remains bounded when every configured price move rounds to zero.
- Precompute the zero-delta anchor ceiling once per resolve scan instead of evaluating wide clamp arithmetic per asset.
- Decode and fail-closed validate only the oracle-profile fields consumed by terminal accrual preflight. Full profile structure remains validated on every public profile write.

## Root cause

PR #255 added a full exposed-asset scan before stale resolution so pending committed mark/funding work cannot be discarded. Its existing 10MiB benchmark exposed only one late asset, while the public dense test stopped at 4,000 equal-mark assets and already consumed about 1.30M CU.

A normally configured low-price market can have a distinct authenticated mark while `anchor * max_price_move_bps_per_slot * max_accrual_dt_slots < 10_000`; the effective-price move is then exactly zero. With funding disabled, there is no work for the account crank, but the pre-fix resolver still fully validates and evaluates every exposed profile. At 2,000 public assets it exhausts the 1.4M transaction budget.

That is a persistent public liveness failure after stale maturity: ordinary exits are `OracleStale`, the honest permissionless crank returns `EngineNonProgress`, mature-state recovery mutations are blocked, and the sole permissionless resolver cannot complete.

## TDD evidence

- Base: exact PR #255 head `eab3c4f66e45381860541a04701e7f0438a0ffe3`
- Red: test commit `086ed2c6` reproduces `ProgramFailedToComplete` at 1,400,000 CU using only normal account construction and public wrapper instructions; no market or portfolio bytes are injected.
- Green: the fixed head resolves the fully exposed 5,834-asset market in 1,140,523 CU and enters `Resolved` mode.
- Real pending price and funding tests still force bounded public catch-up before resolution.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 570/570 LiteSVM/CU tests and 6/6 host tests pass
- `cargo test --test v16_cu v16_dense_zero_delta_price_managed_stale_market_remains_resolvable -- --nocapture`: 1/1 pass, 1,140,523 CU

This PR is intentionally stacked on #255 because the vulnerable accrual preflight does not exist on current `main`.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4583,6 +4583,45 @@ pub mod processor {
         Ok(profile)
     }
 
+    fn read_oracle_profile_for_resolve_scan_view(
+        group: &state::MarketViewMutV16<'_>,
+        asset_index: usize,
+    ) -> Result<state::AssetOracleProfileV16, ProgramError> {
+        // Public profile writes perform the full structural validation. Resolution only consumes
+        // this price summary, so re-walking every cold oracle-feed field here would make the sole
+        // terminal path exceed max-shape CU. Revalidate every field that this scan reads.
+        let market = group
+            .markets
+            .get(asset_index)
+            .ok_or(PercolatorError::InvalidInstruction)?;
+        let bytes = market
+            .wrapper
+            .get(..constants::ASSET_ORACLE_PROFILE_LEN)
+            .ok_or(PercolatorError::InvalidAccountLen)?;
+        let profile: state::AssetOracleProfileV16 = bytemuck::pod_read_unaligned(bytes);
+        match profile.oracle_mode {
+            constants::ORACLE_MODE_MANUAL
+            | constants::ORACLE_MODE_HYBRID_AFTER_HOURS
+            | constants::ORACLE_MODE_EWMA_MARK
+            | constants::ORACLE_MODE_AUTH_MARK => {}
+            _ => return Err(ProgramError::InvalidAccountData),
+        }
+        if oracle_v16::profile_is_price_managed(&profile)
+            && (profile.mark_ewma_e6 == 0
+                || profile.mark_ewma_e6 > percolator::MAX_ORACLE_PRICE
+                || profile.oracle_target_price_e6 == 0
+                || profile.oracle_target_price_e6 > percolator::MAX_ORACLE_PRICE)
+        {
+            return Err(ProgramError::InvalidAccountData);
+        }
+        if oracle_v16::profile_is_auth_mark(&profile)
+            && profile.mark_ewma_e6 != profile.oracle_target_price_e6
+        {
+            return Err(ProgramError::InvalidAccountData);
+        }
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent composition check on this exact head `5f1789a157dd06a27ecc4c966dcfa6a97d164d6a` passes.

I built its real SBF and ran a separately written public stale-resolution value test: two independent funded owners, balanced OI, a separate honest AuthMark authority, pending 1,000,000 -> 1,010,000 mark, stale maturity, terminal closes, and payout topups. The premature resolver rejects, one bounded stored-state crank commits the mark, retry resolves, and terminal payouts exactly match the crank-first control at 2,010,000,000 / 1,990,000,000.

This validates that the max-shape optimization preserves PR #255's value-safety and catch-up liveness on the ordinary one-asset path. It is not an independent reproduction of this PR's 5,834-asset CU red case. Static review of the optimization found no new signer/custody surface: the reduced profile decoder fail-closes every field consumed by the scan, public profile writes still perform full validation, and the zero-delta ceiling is the exact condition `anchor * max_bps * max_dt < 10_000`.

aeyakovenko:
Independent clean-room red/green confirmation against real SBF.

I used a distinct valid zero-delta configuration from the PR test: 2,250 assets, initial/effective price `33`, `max_price_move_bps_per_slot = 3`, funding disabled, and authenticated targets pushed to `99`. Every asset was activated and exposed through ordinary public instructions; no slab or portfolio bytes were injected.

On exact stacked base `eab3c4f66e45381860541a04701e7f0438a0ffe3`, ordinary exits reject as `OracleStale`, the stored-state catch-up rejects as `EngineNonProgress`, and `ResolveStalePermissionless` consumes exactly 1,400,000 CU with `ProgramFailedToComplete`, leaving mode `Live`. (At 2,100 assets the same resolver still barely lands at 1,355,113 CU, so this also measures the boundary rather than assuming it.)

On exact head `5f1789a157dd06a27ecc4c966dcfa6a97d164d6a`, the byte-equivalent state resolves in 439,284 CU and enters `Resolved`. This confirms #269 as a real persistent public DoS for a valid dense deployment. Meta governance must configure the asset set, but unaffiliated users can publicly create the balanced exposure that makes the sole terminal path hit this scan.

aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_dense_zero_delta_resolution_shape_matrix_keeps_terminal_exit_bounded`. Public max-shape stale resolution and funded terminal payout both fit on the pinned program. The executable disposition remains in PR #135; no separate production fix is required.

PR #270, opened 2026-07-19
[BLOCKER DoS] Keep source-heavy portfolios settlement-live

aeyakovenko:
## What changed

Add a public LiteSVM regression for a persistent source-domain capacity DoS and pin the paired engine fix from aeyakovenko/percolator#142.

The test fills all 32 sparse portfolio source-domain entries through ordinary trades and settlement across 16 assets, retains one old exposure, and then attempts to admit a leg whose favorable K/F domain is absent. On the pre-fix engine that admission succeeds. After an honest AuthMark move from 100 to 101, every bounded owner or keeper continuation fails and rolls back:

- `PermissionlessCrank`
- `ConvertReleasedPnl`
- `TradeNoCpi` closing the newly admitted leg
- `TradeNoCpi` closing the old leg
- `RebalanceReduce`

Only privileged shutdown followed by forfeiture can recover the portfolio, so this is a real public-interface DoS even with an honest cranker.

The engine fix reserves sparse source capacity before admitting new exposure. The fixed behavior rejects the unsafe admission atomically, leaves the old position closeable, and lets the owner withdraw the original principal.

## TDD evidence

- Exact wrapper parent: `da64be639168b496833dd4c701607a94fcb06b6c`
- Exact vulnerable engine: `143e68c4917ed0400a27b952f036a5677047cd84`
- Test-only commit: `b3e8d2df`
- RED: `v16_attack_source_domain_capacity_reserves_settlement_for_admitted_legs` panics at `an admitted live leg has no bounded owner/keeper continuation after honest mark movement`
- GREEN engine: `c1a058f3f6057838e78a6ada51509af336fd4ed6`

The test intentionally does not convert all 32 historical claims after the safe rejection. That max-source conversion CU path is already tracked separately; including it here would conflate two findings. This regression proves that the newly admitted position cannot create the permanently stuck state and that existing principal remains exit-live.

## Verification

- `cargo test --all-targets -- --test-threads=1`
  - 6 unit tests passed
  - 566 LiteSVM/CU tests passed
- `v16_bpf_batch_trade_14_legs_under_tx_limit`: 1,308,173 CU
- `v16_bpf_batch_trade_cpi_14_legs_under_tx_limit`: 1,339,517 CU
- `v16_attack_10m_batch_tradecpi_max_tail_rejects_before_cu_exhaustion`: 1,343,687 CU

The engine branch also passes its full fuzz/conformance/spec suite and the final attach-body Kani composition (6,137 checks, 1/1 cover, zero failures).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=c1a058f3f6057838e78a6ada51509af336fd4ed6#c1a058f3f6057838e78a6ada51509af336fd4ed6"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "c1a058f3f6057838e78a6ada51509af336fd4ed6" }
 anchor-lang-v2 = { git = "https://github.com/solana-foundation/anchor", rev = "e3cf760826fa0a60f247635ea0572e59861ec9b5", package = "anchor-lang-v2", default-features = false, features = ["alloc", "guardrails"], optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
 # Host tests use the same zero-copy account/view API as the SBF/program target.
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "c1a058f3f6057838e78a6ada51509af336fd4ed6" }
 
 [dev-dependencies]
 bincode = "1"
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -59717,3 +59717,304 @@ fn v16_attack_underfunded_exit_cannot_move_ewma_with_uncollectible_fee() {
         assert_underfunded_ewma_exit_uses_collected_fee(path);
     }
 }
+
+// Public-interface DoS regression: retain every sparse source domain, keep an old exposure active,
+// then try to admit a leg whose first favorable settlement needs a 33rd domain. The vulnerable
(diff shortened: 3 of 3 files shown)


aeyakovenko:
Independent clean-room LiteSVM confirmation against the exact binaries, extended to remove the owner-footgun ambiguity.

I changed only a temporary harness at wrapper head `cf69b40e095bf4ee6b750af56c68f1c3c05b9c0a`. The victim LP fills all 32 source-domain entries and retains the old exposure through public instructions, then authorizes the standard matcher. A separate unaffiliated attacker is the **only transaction signer** on the final `TradeCpi`; the full-table victim is the passive LP.

RED, exact Meta pin `da64be639168b496833dd4c701607a94fcb06b6c` / engine `143e68c4917ed0400a27b952f036a5677047cd84`:
- attacker-only `TradeCpi` admits the unsafe victim leg;
- honest AuthMark 100 -> 101 and losing-side crank succeed;
- victim `PermissionlessCrank`, `ConvertReleasedPnl`, a new-leg close with a fresh fully funded rescue counterparty, old-leg close, and `RebalanceReduce` all fail;
- failed continuations leave market and victim account bytes unchanged.

GREEN, exact wrapper `cf69b40e095bf4ee6b750af56c68f1c3c05b9c0a` / engine `c1a058f3f6057838e78a6ada51509af336fd4ed6`:
- the attacker-only admission rejects atomically with market, victim, historical counterparty, attacker, and matcher snapshots unchanged;
- the old exposure still closes and the victim withdraws its full 1,000,000-unit principal.

Command used for both binaries:
`cargo test --test v16_cu v16_probe_matcher_cannot_admit_an_unsettleable_lp_leg -- --nocapture --test-threads=1`

This is a real public-interface DoS: an unaffiliated taker can trap an already-authorized passive LP; it does not require the victim owner to submit the unsafe trade.

aeyakovenko:
Independent Meta LiteSVM follow-up found the max-source conversion path that this PR explicitly leaves separate.

Public red test: https://github.com/aeyakovenko/percolator-meta/commit/1e1a33f

The fixture creates every claim through signed `TradeNoCpi`, authenticated mark pushes, and permissionless settlement. No account-state injection is used.

Results at the chain's 1,400,000-CU ceiling:

- This PR's exact fixed pin (`cf69b40e` / engine `c1a058f3`): 16 domains convert in 1,017,938 CU; 32 domains on a minimum capacity-17 slab consume exactly 1,400,000 CU and fail at the BPF meter.
- Domain-local conversion head (`81dfda28` / engine `f0e189a1`): 16 domains convert in 1,023,549 CU; the same 32-domain minimum slab also consumes exactly 1,400,000 CU and fails.

The failed conversion rolls market and portfolio bytes back. The owner can withdraw the original 1,000,000-unit principal, but cannot withdraw any of the 32 earned, source-backed units, and the remaining positive-claim portfolio cannot close. There is no domain-selective public conversion wire, so the supported state has no bounded live owner path to recover earned value.

This is a separate real public-interface LoF, not a reason to weaken this admission fix. A fix needs either a transaction-bounded domain-local conversion cursor/chunk or a proven full conversion below the ceiling at all supported source shapes, while preserving the caller cap and PR #267/#140's domain-local burn semantics. I left the Meta branch red and opened no Meta PR pending that upstream fix.

aeyakovenko:

PR #271, opened 2026-07-19
[BLOCKER LoF] Settle zero-move funding before trades

aeyakovenko:
## Impact

A short could close through a pre-authorized permissionless CPI LP immediately before the honest crank and erase an already-deterministic premium-funding interval. The edge is reachable when the bounded price step rounds to zero but signed premium funding does not.

The exact-parent LiteSVM reproduction uses an honest AuthMark target of `1` against effective price `2`, a 24-bps per-slot cap, a fair authenticated matcher, and normal public instructions. Only the attacker signs the close; the independent LP had previously authorized its matcher.

On `da64be639168b496833dd4c701607a94fcb06b6c` with engine `143e68c4917ed0400a27b952f036a5677047cd84`:

```text
crank first: attacker=999900  LP=1000100  F=(1000000000000000,-1000000000000000)
close first: attacker=1000000 LP=1000000  F=(0,0)
```

The attacker retains 100 collateral atoms and the independent LP loses the same 100 atoms. Effective price remains exactly `2` in both paths, so this is not pending-price PnL, dishonest-oracle behavior, cadence rounding, or an omitted-observation crank.

## Root cause

Trade mutated balanced OI before committing the elapsed funding segment implied by the stored authenticated mark. Once both positions were flat, the later honest crank correctly found no balanced exposure and accrued zero funding.

## Fix

- Before matcher invocation and before every direct position delta, detect a managed-price interval where the engine-bounded price is unchanged but premium funding is nonzero.
- Commit exactly one normal `max_accrual_dt_slots`-bounded engine accrual segment at the unchanged effective price.
- Let the existing engine trade settlement apply the new F index before changing OI.
- Cover single and batch CPI paths; direct paths share the same pre-execution helpers.

The fix does not require an oracle account, reject the trade, grant a synthetic time step, or retroactively settle more than the engine's normal bounded segment. Nonzero pending price movement remains separately scoped to #260.

## TDD

- `e145af9d`: exact-parent public red test, covering both `TradeCpi` and one-leg `BatchTradeCpi` with an unsigned independent LP.
- `f52b7fbc`: production fix; close-first and crank-first balances and funding indexes are identical.
- Conservation is asserted through final SPL withdrawals in both orderings.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 6 host + 566 LiteSVM/CU tests passed
- existing 14-leg direct and CPI batch CU tests passed
- three focused funding/mark Kani harnesses passed with all covers reached
- full `cargo kani --tests` was stopped on the unchanged main-branch `kani_v16_unknown_or_truncated_tags_reject` CBMC blowup; this PR does not claim that pre-existing all-harness gate
- `cargo clippy --all-targets --no-deps` passed with existing warnings only
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -5828,6 +5828,11 @@ pub mod processor {
             } else {
                 size_q.unsigned_abs()
             };
+            accrue_zero_move_funding_before_trade_view(
+                &oracle_profile,
+                &mut group,
+                asset_index as usize,
+            )?;
             // F-TRADENOCPI-FEE / F-NOCPI-MARK-FEE: request.exec_price is used by the engine only as
             // fee notional. In price-managed EWMA/stale-hybrid modes, the caller's reported print is
             // also the mark-discovery input, so first normalize it to the same per-asset dt price
@@ -6083,6 +6088,11 @@ pub mod processor {
                     return Err(PercolatorError::InvalidInstruction.into());
                 }
                 let abs_size = leg.size_q.unsigned_abs();
+                accrue_zero_move_funding_before_trade_view(
+                    &oracle_profile,
+                    &mut group,
+                    asset_index,
+                )?;
                 let fee_basis_price = accepted_reported_trade_price_view(
                     &oracle_profile,
                     &group,
@@ -6554,6 +6564,8 @@ pub mod processor {
             return Err(PercolatorError::InvalidInstruction.into());
         }
 
+        accrue_zero_move_funding_before_matcher_view(market_ai, &[asset_index])?;
+
         let (cfg_pre, mode_pre, current_slot_pre, oracle_price, max_trading_fee_bps) =
             state::read_market_trade_preflight(
                 &market_ai.try_borrow_data()?,
@@ -6867,6 +6879,7 @@ pub mod processor {
             }
             asset_indices.push(leg.asset_index);
         }
+        accrue_zero_move_funding_before_matcher_view(market_ai, &asset_indices)?;
         // One header/config parse for mode + slot + every leg's oracle price (avoids O(N^2)
         // re-parsing the market once per leg).
         let (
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent clean-room LiteSVM found a missing public reduction surface. The matcher/direct-trade fix is correct for the submitted routes, but `RebalanceReduce` can still erase the same deterministic zero-move funding interval.

Fresh probe: `v16_probe_rebalance_partial_reduce_cannot_erase_elapsed_premium_funding`.

Public sequence:
- effective price `2`, honest AuthMark target `1`, 24 bps/slot cap, one-slot accrual;
- attacker opens a 100-unit short through the standard pre-authorized matcher; the independent long LP is passive and only the attacker signs the trade;
- slot 1 cranks publish the honest target without retroactive funding;
- at slot 2 the attacker alone calls `RebalanceReduce` for half its short before the honest crank, then the crank lands and the attacker reduces the remainder;
- attacker withdraws normally; the passive LP is paid through normal market resolution.

Exact value result (2,000,000 total atoms conserved):
- crank first: attacker `999900`, LP `1000100`;
- half-reduce first: attacker `999950`, LP `1000050`.

The attacker extracts 50 atoms from the independent LP. This reproduces on:
- exact parent `da64be639168b496833dd4c701607a94fcb06b6c`;
- exact #271 head `f52b7fbcbc42219cd7dc19a1c21b9499d6fd47a8`;
- a clean SBF composition of #263 production commit `90219b488597c6c5971c6071cfd4158fe240f965` plus #271 production commit `f52b7fbcbc42219cd7dc19a1c21b9499d6fd47a8`.

The composed result matters: #263's bounded adverse-price check sees an unchanged effective price, while #271 calls `accrue_zero_move_funding_before_trade_view` only from Trade/BatchTrade CPI and no-CPI paths. `handle_rebalance_reduce` still mutates OI directly without committing the funding segment.

Please extend #271's invariant to `RebalanceReduce` (and retain a public half-reduction regression). This is the same root cause and should be fixed in this PR before merge rather than split into another overlapping branch.

aeyakovenko:
Independent LiteSVM confirmation: https://github.com/aeyakovenko/percolator-meta/pull/280

On the current Meta pins (`cf69b40e` program / `c1a058f3` engine), a retained slot-1 `TradeNoCpi` close submitted before the slot-2 crank lets the paying short withdraw 1,000,000 instead of 999,900 and reduces the independent long claim from 1,000,100 to 1,000,000. Funding counters fall from 100/100 to 0/0.

A clean composition of only this PR's `src/v16_program.rs` diff onto `cf69b40e` produces SBF SHA-256 `f1d9b0a7d16a59d45aeb8ef76af5ad20e6033860e5b469c9b6710e3fd164a48b` and makes the exact test pass. This is a real public LoF, but the PR head is still based on the older `da64be63` / `143e...` line and should be rebased onto the current pins before Meta repins.

aeyakovenko:

PR #272, opened 2026-07-19
[BLOCKER LoF] Settle zero-move funding before unilateral reduction

aeyakovenko:
## What changed

- Accrue one normal engine-clamped funding segment before `RebalanceReduce` when the bounded mark step rounds to the current effective price but premium funding is nonzero.
- Let the engine settle that funding onto the reducing account before unilateral OI reduction.
- Add a public LiteSVM red/green exploit test and enforce the existing custody CU limit.

## Root cause and impact

`RebalanceReduce` settled the caller at the current K/F cursor and then reduced matching effective OI. A short could therefore reduce to flat before an honest crank during an elapsed premium interval. Full side reset then removed the funding opportunity for the independent long.

The public repro uses AuthMark target `1`, effective price `2`, a 24 bps price cap, and balanced `100 * POS_SCALE` exposure. The bounded price step rounds to zero while premium funding is 100 atoms:

- honest crank first: attacker withdraws `999900`; LP claim is `1000100`; paid/received counters are `(100, 100)`
- attacker reduces first: attacker withdraws `1000000`; LP claim is `1000000`; counters are `(0, 0)`

Only the attacker signs the attack reduction. The LP is independently owned, and its post-reduction claim is settled by a permissionless crank. The test uses only public instructions and no state injection.

The finalized test-only commit `619a46b3` fails on exact main parent `da64be63`; fix commit `90eaf762` passes it. Nonzero pending price moves are intentionally unchanged here so this composes with #263.

## Verification

- exact-parent red LiteSVM test
- `cargo test --all-targets -- --test-threads=1` (6 host + 566 LiteSVM/CU tests)
- `cargo clippy --all-targets` (passes with existing warnings)
- `cargo kani --tests --harness kani_v16_premium_funding_rate_is_clamped_and_signed`
- `cargo kani --tests --harness kani_v16_collected_base_fee_cannot_fund_mark_movement`
- `cargo kani --tests --harness kani_v16_fee_supported_mark_clamp_is_directional_and_zero_support_is_noop`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4555,6 +4555,75 @@ pub mod processor {
         Ok(())
     }
 
+    fn accrue_zero_move_funding_before_rebalance_reduce_view(
+        cfg: &WrapperConfigV16,
+        group: &mut state::MarketViewMutV16<'_>,
+        asset_index: usize,
+    ) -> Result<(), V16Error> {
+        if asset_index >= group.header.config.max_market_slots.get() as usize
+            || asset_index >= group.markets.len()
+        {
+            return Err(V16Error::InvalidConfig);
+        }
+        let profile = read_oracle_profile_from_view(group, cfg, asset_index)
+            .map_err(|_| V16Error::InvalidConfig)?;
+        if !oracle_v16::profile_is_price_managed(&profile) {
+            return Ok(());
+        }
+
+        let now_slot = authenticated_market_slot_or_fallback_view(group);
+        let asset = group.markets[asset_index].engine.asset;
+        let slot_last = asset.slot_last.get();
+        if now_slot < slot_last {
+            return Err(V16Error::Stale);
+        }
+        let dt = core::cmp::min(
+            now_slot - slot_last,
+            group.header.config.max_accrual_dt_slots.get(),
+        );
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent clean-room Meta LiteSVM confirms this is a **REAL public LoF**, not only ordering hardening.

Meta proof: https://github.com/aeyakovenko/percolator-meta/pull/279 (`168c204`).

Against the exact current Meta pin (`cf69b40e`, engine `c1a058f3`, SBF `15e43cb3939f365fdc7f5f3e3b4c9a6c0046db4306a48aa226af78da847b396d`):
- crank first: attacker SPL withdrawal `999900`, independent long claim `1000100`, funding counters `100/100`;
- owner-only reduce first: attacker SPL withdrawal `1000000`, independent long claim `1000000`, counters `0/0`.

I also applied only this PR's production `src/v16_program.rs` diff onto `cf69/c1a`, built SBF `04928fa53022c7862b981407cb2798e12b4ad2848b26ef8c6980714e4ce41618`, and the identical test passed. Please rebase the fix onto the current source-capacity program/engine line before merge; the submitted head is still based on `da64/143e`.

aeyakovenko:

PR #273, opened 2026-07-19
[BLOCKER LoF] Accrue zero-move funding before recovery-leg forfeit

aeyakovenko:
## [REAL LoF] Accrue zero-move funding before recovery-leg forfeit

`ForfeitRecoveryLeg` settles a dead leg via `record_account_funding_flow` at the **stale** funding index and then detaches the leg. When an elapsed premium-funding segment has not yet been booked because the oracle price move rounded to zero (a "zero-move segment"), the forfeit advances past that segment without charging it, and detaching the leg zeroes the side's open interest — so a later keeper crank can no longer book the segment for the surviving counterparty. The forfeiting short escapes the premium it owes; its independent long counterparty is denied the credit it earned. This is the same `record_account_funding_flow`-at-stale-index primitive proven real on the rebalance-reduce (#272) and trade (#271) paths, reached here through a **distinct instruction** (`ForfeitRecoveryLeg`) and a **distinct settle fn** (`settle_forfeited_leg_kf_effects`).

### Measured terminal values (test isolates the attacker↔LP matched pair)

Public setup: live market, honest AuthMark, funding rate ≠ 0, price move rounds to zero over slot 1→2. Attacker holds a small short (5) matched to an independent LP long (5); a separate long unilaterally socializes >90% of matching short OI via public `RebalanceReduce`, dropping `a_short` below `MIN_A_SIDE` so `mode_short = DrainOnly` (leg is forfeitable) while the market stays Live and the asset Active. No `mutate_market`, no injection.

| party / field | (A) forfeit **before** crank — pre-fix | (B) crank first (honest) |
|---|---|---|
| attacker withdrawn | **1,000,000** (pays 0) | 999,995 (pays 5) |
| attacker `funding_short_paid` | **0** | 5 |
| independent LP pnl | **0** | 5 |
| independent LP `funding_long_received` | **0** | 5 |
| insurance | 0 | 0 |
| vault | 4,000,000 | 4,000,000 |

Pre-fix, ordering A lets the attacker keep the 5 it owes and denies the still-open independent LP the 5 it earns (vault/insurance untouched — a funding-transfer leak, exactly as in #271/#272). Post-fix, ordering A matches the honest ordering B: attacker pays 5, LP receives 5.

### Fix

Wrapper-only (engine pin `143e68c4` unchanged, like #271/#272). New helper `accrue_zero_move_funding_before_recovery_forfeit_view` accrues the elapsed zero-move segment onto the asset funding accumulators **before** `forfeit_recovery_leg_not_atomic`, so the forfeit settle observes the correct funding index and the surviving side keeps its claim. Guarded on a **true zero-move** interval (bounded `next` price `== current` effective price) and on nonzero two-sided OI / nonzero premium rate, so it is a strict no-op whenever a crank would not have accrued — no legitimate price-moving interval is double-counted. Follows the #271/#272 accrue-before-act pattern.

This corrects a prior mis-analysis that classed the forfeit path DEFENDED on the premise that a stale-index forfeit settle is a non-extractable lazy deferral; the terminal-value measurement above disproves that premise for the zero-move segment.

### Tests
- Red test `v16_attack_forfeit_recovery_leg_cannot_erase_elapsed_premium_funding` fails on the pre-fix parent (attacker pays 0 / LP receives 0) and passes on the fix (5 / 5).
- Full `cargo test --test v16_cu` green: **566 passed, 0 failed**.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4555,6 +4555,84 @@ pub mod processor {
         Ok(())
     }
 
+    // Accrue an elapsed but as-yet-unbooked zero-move premium-funding segment onto the asset's
+    // funding accumulators BEFORE a dead-leg recovery forfeit settles/clears the leg. Without this,
+    // `ForfeitRecoveryLeg` settles the forfeited leg at the STALE funding index (f_delta over the
+    // zero-move segment is 0) and then detaches the leg — zeroing the side's open interest so a
+    // later keeper crank can no longer book the segment for the surviving counterparty. The premium
+    // the forfeiting owner owed is destroyed and its independent counterparty is denied the credit
+    // it earned. Mirrors the unilateral rebalance-reduce accrue-before-act ordering fix: only fires
+    // on a true zero-move segment (bounded next price == current effective price) so no legitimate
+    // price-moving interval is double-counted; a no-op whenever a crank would not have accrued.
+    fn accrue_zero_move_funding_before_recovery_forfeit_view(
+        cfg: &WrapperConfigV16,
+        group: &mut state::MarketViewMutV16<'_>,
+        asset_index: usize,
+    ) -> Result<(), V16Error> {
+        if asset_index >= group.header.config.max_market_slots.get() as usize
+            || asset_index >= group.markets.len()
+        {
+            return Err(V16Error::InvalidConfig);
+        }
+        let profile = read_oracle_profile_from_view(group, cfg, asset_index)
+            .map_err(|_| V16Error::InvalidConfig)?;
+        if !oracle_v16::profile_is_price_managed(&profile) {
+            return Ok(());
+        }
+
+        let now_slot = authenticated_market_slot_or_fallback_view(group);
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent clean-room LiteSVM confirmation: https://github.com/aeyakovenko/percolator-meta/pull/281

The Meta probe puts the short side into `DrainOnly` at slot 0, before the measured premium interval, so `RebalanceReduce` cannot confound the result. On the current `cf69b40e` / `c1a058f3` pins, crank-first gives attacker/LP `999995 / 1000005` with funding counters `5/5`; forfeit-first gives `1000000 / 1000000` with counters `0/0`.

Applying only this PR's 83-line wrapper change to `cf69b40e` produces SBF SHA-256 `774a5f8020d061d4dc678bdab2d009d62a1a5112134eb09f454d477c6b6de0b0` and makes the exact test pass. This independently confirms the classification as real public LoF.

aeyakovenko:

PR #275, opened 2026-07-19
[BLOCKER LoF] Bind signed mark pushes to asset generations

aeyakovenko:
## What changed

- add the current asset `market_id` to signed `PushAuthMark` and `PushEwmaMark` payloads
- reject a retained report when its generation differs from the live asset occupying that slot
- add public LiteSVM lifecycle coverage for direct and smoothed authority-mark paths
- extend the mark wire-field Kani harness over the generation field

## Impact and root cause

The mark authority signed a slot index, timestamp, and price, but not the asset incarnation. A signed report retained from an empty old asset could therefore land after permissionless shutdown/restart of that slot and move the mark against positions opened only in the replacement generation.

On the vulnerable parent, the public LiteSVM lifecycle uses ordinary configure, shutdown, restart, deposit, trade, crank, PnL conversion, and withdrawal instructions. The stale report produced extractable transfers from fresh independent victim capital:

- AuthMark: victim equity 1,000,000 -> 750,000; beneficiary withdrawal 1,250,000
- EwmaMark: victim equity 1,000,000 -> 875,000; beneficiary withdrawal 1,125,000

The signed asset `market_id` now identifies the intended incarnation, and mismatch rejection occurs before market or portfolio mutation. Current-generation pushes remain valid.

This PR is stacked on #231 because that PR supplies the persistent market-generation boundary used by the fix.

## TDD commits

- RED: `f0135bd1` (`test: reproduce stale oracle report generation replay`)
- GREEN: `22bc6e78` (`Bind mark pushes to asset generations`)

## Verification

- `cargo fmt --check`
- `cargo check --tests`
- focused AuthMark/EwmaMark public LiteSVM reproduction passed on the fixed program
- `cargo test --all-targets`: 5 host + 569 LiteSVM/CU tests passed
- `cargo kani --tests --exact --harness kani_v16_ewma_mark_decode_preserves_wire_fields`: 1/1 harness, 847 checks passed

The aggregate Kani run reached the repository's known variable-length trailing-byte decoder expansion wall and was stopped after more than 20 minutes. The changed mark wire surface is covered by the exact focused harness above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,8 +468,8 @@ This section describes intent and operational ordering, not argument-by-argument
 - External-oracle markets authenticate configured oracle account(s) in the oracle configuration/crank
   paths; `TradeCpi` / `TradeNoCpi` use the already-stored effective mark for fee and settlement
   accounting.
-- AuthMark markets use **ConfigureAuthMark** (tag 62) and **PushAuthMark** (tag 63), signed by the configured mark authority, to store a direct authority mark without EWMA smoothing.
-- EwmaMark markets use **ConfigureEwmaMark** (tag 35) and **PushEwmaMark** (tag 36), signed by the configured mark authority, to update a smoothed EWMA mark input.
+- AuthMark markets use **ConfigureAuthMark** (tag 62) and **PushAuthMark** (tag 63), signed by the configured mark authority, to store a direct authority mark without EWMA smoothing. Pushes include the current asset `market_id` so a signed report cannot cross an asset restart.
+- EwmaMark markets use **ConfigureEwmaMark** (tag 35) and **PushEwmaMark** (tag 36), signed by the configured mark authority, to update a smoothed EWMA mark input. Pushes carry the same generation binding.
 - The per-slot effective-price movement cap is a risk parameter set at init; there is no standalone `SetOraclePriceCap` instruction in the current ABI.
 
 ### Insurance management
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Clean-room Meta integration independently confirms this as a real live-instance LoF through the supported controller lifecycle.\n\nReproduction: https://github.com/aeyakovenko/percolator-meta/commit/0e2edaf\n\nThe retained transaction is signed by an independent external oracle before handoff. A permissionless creator donates lifecycle control to the constrained Meta controller, governance shuts down and restarts the empty slot, fresh users each deposit 1,000,000 atoms and open balanced exposure, and the original transaction is submitted without re-signing. A same-blockhash control transaction succeeds immediately before replay. On the current Meta pin (percolator-prog 5f1789a), replay reduces fresh victim equity to 750,000 and the beneficiary converts and publicly withdraws 1,250,000.\n\nThis PR's current stack is not safe for Meta to pin yet. It diverges from 5f1789a and is 12 commits behind that line; in particular it predates merged monotonic portfolio IDs (567c76c) and the collected-fee EWMA protection (#198 / c8cf4de). Pinning 22bc6e7 would remove those protections. Please rebase/stack the generation binding onto the current security line without dropping those changes. Once there is one compatible immutable program commit, the Meta reproduction can become the green pin regression.

aeyakovenko:

PR #276, opened 2026-07-19
[HARDENING LoF] Bind signed trades to portfolio incarnations

aeyakovenko:
## Finding

A counterparty can retain two fully signed public `TradeNoCpi` transactions for an old portfolio incarnation, wait for that portfolio address to be closed and publicly reinitialized, then order the retained open and close around an honest mark move. The signatures bind only account pubkeys, so before this fix they authorize risk against fresh replacement capital.

The exact-parent LiteSVM RED repro uses ordinary instructions and retained transaction objects, with no state injection:

- victim replacement equity falls from 1,000,000 to 915,000 atoms
- attacker exits custody with 1,105,000 atoms from a 1,000,000-atom deposit
- the mark move is an honest AuthMark update from 100 to 50

RED: `b304de33`
GREEN: `4505f934`

## Fix

All four signed trade payloads now bind both program-assigned portfolio incarnation IDs and validate them before matcher CPI or engine mutation: `TradeNoCpi`, `TradeCpi`, `BatchTradeNoCpi`, and `BatchTradeCpi`. Existing pre-ID portfolios receive a reserved legacy incarnation on first trade; already-grown zero-filled legacy tails are handled without reallocation.

The GREEN test checks exact `EngineProvenanceMismatch`, byte-for-byte rollback, all four routes, matcher-context non-execution on stale IDs, and a current-incarnation open/close control.

## Verification

- `cargo test --all-targets`: 6 host + 566 LiteSVM/CU tests pass
- `kani_v16_tradenocpi_decode_preserves_wire_fields`: 845 checks pass
- `kani_v16_tradecpi_decode_preserves_wire_fields`: 845 checks pass
- batch no-CPI/CPI wire proof: 1,645 checks pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,10 +266,11 @@ Benefits:
 Positions and PnL use native `i128`/`u128` (`POS_SCALE = 1_000_000`, `ADL_ONE = 1_000_000_000_000_000`). There are no I256/U256 wrapper types for positions or PnL. Positions use the ADL A/K coefficient mechanism defined in the spec.
 
 ### Two trade paths
-- **TradeNoCpi**: no external matcher; used for baseline integration, local testing, and deterministic program-test scenarios.
+- **TradeNoCpi**: no external matcher; used for baseline integration, local testing, and deterministic program-test scenarios. Its signed payload binds both program-assigned portfolio incarnation IDs.
 - **TradeCpi**: production matcher path; the LP owner configures a matcher program/context once on
   the LP portfolio with `SetMatcherConfig` (tag 68). Fills then run without the LP owner
-  signing each transaction. Direct LP-signed bilateral trading is `TradeNoCpi`.
+  signing each transaction. The taker's signed payload binds both portfolio incarnation IDs. Direct
+  LP-signed bilateral trading is `TradeNoCpi`.
 
 ### MatchingEngine trait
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Clean-room Meta LiteSVM confirmation: this is a **real public LoF**, not hardening.

Probe: [`f58af24`](https://github.com/aeyakovenko/percolator-meta/commit/f58af24) / [Meta draft PR #276](https://github.com/aeyakovenko/percolator-meta/pull/276).

- RED on the exact currently pinned program `cf69b40e` / engine `c1a058f3` SBF (`sha256 15e43cb3...`): retained old-incarnation open and close signatures execute after the victim publicly closes, reinitializes the same portfolio address, and deposits fresh capital. An honest AuthMark move then leaves the replacement victim at `750,000` from `1,000,000`, while the counterparty withdraws `1,250,000` from an original `1,000,000`.
- GREEN on a clean-room legacy-entrypoint SBF built from this PR's exact head `4505f934` (`sha256 1f29b2df...`): the retained old wire is rejected before mutation, and the replacement portfolio plus market remain byte-identical.

The current head pins engine `143e68c`, while Meta is already on `c1a058f3`; the branches diverged before later source-domain capacity fixes. Please rebase this fix onto the current `cf69b40e` / `c1a058f3` program-engine stack before merge. Meta should not repin to this head as-is because doing so would reopen those later fixes. After that rebase, the Meta test can switch from the deliberately retained old wire to the generation-bound trade wire and assert the stale IDs are rejected directly.

aeyakovenko:

PR #277, opened 2026-07-19
[BLOCKER LoF] Bind oracle configuration to asset generations

aeyakovenko:
## Finding

A fully signed `ConfigureAuthMark` transaction retained from an old, empty asset generation remained valid after public shutdown/restart reused the same asset slot. An attacker can land it immediately before a fresh-generation trade signed against the visible replacement anchor. The stale configuration resets the entry anchor, and the honest current-generation mark update realizes the artificial PnL against an independent victim.

The exact-parent LiteSVM RED repro uses only ordinary public instructions and retained `Transaction` objects. With deposits of 1,000,000 each, the victim finishes at 750,000 equity and the beneficiary exits and withdraws 1,250,000. No account bytes are injected.

## Fix

Bind `ConfigureHybridOracle`, `ConfigureEwmaMark`, `ConfigureAuthMark`, and `RestartAssetOracle` to the current asset `market_id`, rejecting stale generations before reading feeds or mutating state. Legacy unbound payloads fail closed. Current-generation configuration remains functional.

This PR is intentionally stacked on #275 (and transitively #231) because it extends the same generation primitive to a distinct signed authorization surface.

## TDD

- RED `3bcac346`: exact public LiteSVM replay; victim 1,000,000 -> 750,000, beneficiary withdraws 1,250,000
- GREEN `c5ca015b`: stale Auth/EWMA/Hybrid/Restart rejected with byte-identical rollback; current-generation Auth succeeds

## Verification

- `cargo test --all-targets`: 5 host + 571 LiteSVM/CU tests pass
- Kani wire proofs: Restart 1,517 checks; EWMA/Auth 853 checks; Hybrid 878 checks
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
   `UpdateAssetLifecycle { action: RETIRE, asset_index: 0 }` rejects, so asset 0 is never returned to
   the permissionless reusable-slot pool. Once any asset 0..N is in RECOVERY and every position/loss plus
   value-bearing source/backing/reservation state for that asset is gone,
-  `RestartAssetOracle { asset_index, ... }` signed by that asset's `asset_admin` atomically retires the old market id,
+  `RestartAssetOracle { asset_index, market_id, ... }` signed by that asset's `asset_admin` atomically retires the old market id,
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #278, opened 2026-07-19
[HARDENING LoF] Bind recovery forfeits to portfolio incarnations

aeyakovenko:
## Root cause

`ForfeitRecoveryLeg` authorized an irreversible value-discarding operation using only the owner signature, portfolio address, and asset slot. It did not bind that consent to the program-assigned portfolio incarnation ID. A signed transaction retained from an old incarnation could therefore execute against a newly initialized portfolio at the same address while its recent blockhash remained valid.

## Public LiteSVM reproduction

The RED commit `2f2b0252` uses only ordinary wrapper and System Program instructions:

1. Initialize a market and two funded portfolios, then open a position.
2. Enter asset recovery and retain an owner-authorized forfeit transaction for portfolio ID 1.
3. Clear, withdraw, and close that portfolio.
4. Re-fund the same address through the System Program and initialize portfolio ID 3.
5. Open a fresh position, move the honest authenticated mark from 100 to its bounded effective value, and enter recovery again.
6. Rebroadcast the retained transaction.

Before the fix, the stale consent detached the replacement winner, discarded 100,000 units of positive PnL, paid the victim only its 1,000,000 principal, and left 100,000 units orphaned in a resolved slab. After both portfolios were paid and closed, `c_tot`, insurance, and the portfolio count were zero but `vault` remained 100,000; public `CloseSlab` failed with `EngineLockActive`.

The final fixture makes the retained transaction fee-paid and signed only by the victim, so rebroadcast does not require a privileged key. Market authority is used only for the documented bounded asset shutdown/restart lifecycle; account data is never injected or rewritten.

## Fix

- Add the monotonic `portfolio_id` to the signed `ForfeitRecoveryLeg` wire payload.
- Reject an ID mismatch before entering the engine mutation.
- Update all positive callers to read and supply the current incarnation.
- Prove symbolic wire round-trip preservation and universal rejection of the legacy unbound payload.

## Verification

- Exact public LiteSVM replay: PASS after failing on the RED parent
- `cargo test --all-targets`: 6 host + 566 LiteSVM/CU tests PASS
- `cargo kani --tests --harness kani_v16_recovery_close_progress_decode_preserves_wire_fields`: PASS
- `cargo kani --tests --harness kani_v16_legacy_unbound_forfeit_payload_is_rejected`: PASS
- `cargo fmt --check`
- `git diff --check`

This PR intentionally fixes only cross-incarnation consent replay. Whether a current-incarnation voluntary forfeit can create a terminal residual is a separate engine-level question and is not masked here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,8 @@ This section describes intent and operational ordering, not argument-by-argument
 - **CureAndCancelClose** (tag 42)
   - owner-signed close recovery path; optional deposit is transferred first, then the engine cancels the pending close if the cure succeeds
 - **ForfeitRecoveryLeg** (tag 43)
-  - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget
+  - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget; the signed
+    payload includes the program-assigned `portfolio_id`, so consent cannot cross account reuse
 - **RebalanceReduce** (tag 44)
   - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector
 - **ClaimResolvedPayoutTopup** (tag 46)
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2651,6 +2651,7 @@ pub mod ix {
             optional_deposit: u128,
         },
         ForfeitRecoveryLeg {
+            portfolio_id: u64,
             asset_index: u16,
             b_delta_budget: u128,
         },
@@ -2924,6 +2925,7 @@ pub mod ix {
                     optional_deposit: read_u128(&mut rest)?,
                 },
                 43 => Self::ForfeitRecoveryLeg {
+                    portfolio_id: read_u64(&mut rest)?,
                     asset_index: read_u16(&mut rest)?,
                     b_delta_budget: read_u128(&mut rest)?,
                 },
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Independent Meta LiteSVM confirmation: https://github.com/aeyakovenko/percolator-meta/pull/282

This is **real public LoF plus terminal DoS**, not merely hardening. The stronger probe atomically combines `ClosePortfolio` with a System transfer, so it does not depend on zero-lamport revival. The retained transaction is fee-paid and signed only by the victim; rebroadcast needs no key.

On `cf69b40e` / `c1a058f3`, consent for portfolio ID 1 detaches replacement ID 3, destroys 250,000 of honest marked profit, pays the victim only 1,000,000 principal, and leaves 250,000 in both the engine/token vault after `c_tot`, insurance, and materialized count reach zero. `CloseSlab` then permanently fails.

Applying only this PR's `src/v16_program.rs` change to `cf69b40e` produces SBF SHA-256 `f6153c5746e7d14f2afaa193cfff1b05222c62b21a6763208565ecab703176db` and passes the same dual-wire test using valid current-ID setup calls and the retained old-ID replay. I recommend retitling this PR `[REAL LoF/DoS]`.

aeyakovenko:

PR #279, opened 2026-07-19
[BLOCKER LoF] Bind collateral top-ups to asset generations

aeyakovenko:
## Summary

- add the current asset `market_id` to `TopUpInsurance`, `TopUpInsuranceDomain`, and `TopUpBackingBucket`
- reject a signed collateral contribution when its generation does not match the live asset slot
- prove the new wire fields round-trip and reject every legacy generation-unbound top-up payload

## Impact

This closes a public stale-signature loss-of-funds path. A victim could sign an insurance contribution for one asset generation without submitting it. After a compromised market authority retired the empty asset and activated a replacement in the same slot, it could retain the victim as `insurance_authority`, assign itself as `insurance_operator`, replay the victim's old transaction, and withdraw the credited collateral.

The LiteSVM reproduction uses only public wrapper instructions, normal market/asset initialization, and ordinary SPL token accounts. On the vulnerable parent it debits 250,000 atoms from the victim and lets the replacement operator withdraw all 250,000 atoms.

## Root cause

Collateral contribution instructions authenticated the signer and logical domain, but the signed payload did not identify the asset incarnation. Slot reuse therefore changed the beneficiary and withdrawal authority without invalidating retained signatures.

This PR is intentionally based on #231 because that PR supplies the persistent monotonic generation source required to keep the binding sound across full market-account recreation.

## TDD evidence

- RED: `c8d1f8aa` on exact parent `4c5753f0` reproduces the 250,000-atom victim loss.
- GREEN: `1517bfe2` rejects the identical retained transaction with `AssetGenerationMismatch`, preserves the victim source and market byte-for-byte, and accepts a current-generation positive control.

## Verification

- `cargo test --all-targets` (5 library + 569 LiteSVM/CU tests passed)
- `cargo kani --tests --exact --harness kani_v16_domain_topup_and_asset_insurance_decode_preserves_wire_fields`
- `cargo kani --tests --exact --harness kani_v16_top_up_backing_bucket_decode_preserves_wire_fields`
- `cargo kani --tests --exact --harness kani_v16_legacy_unbound_collateral_topup_payloads_are_rejected`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ Percolator enforces three layers with distinct responsibilities:
 - **Layout**: header + wrapper config + `MarketGroupV16Account`
 - Holds market-level totals, insurance, oracle/asset state, source-domain credit state, and asset lifecycle state.
 
-The v16 asset index ABI is `u16`. The current persisted layout is still a fixed-capacity Pod market-group layout, but asset indices are treated as reusable logical slots. A retired asset slot can only be reactivated after the configured shutdown/activation timeout, and reactivation assigns a new monotonic `u64` `market_id` from the market group. `market_id` values are never reused, including when a closed market account is recreated at the same address. Trade instructions, co-signed authority rotations, portfolio legs, and close-progress ledgers carry that id, so neither stale signed intent nor stale state from an old shutdown market can bind to a reused slot.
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #281, opened 2026-07-19
[BLOCKER LoF] Keep B settlement source-domain local

aeyakovenko:
## Finding

An asset-1 B-settlement loss could retire a source-backing claim from asset 0 before consuming the liability's own asset-1 source domain. A public terminal settlement could therefore transfer value between unrelated asset domains and leave the asset-1/provider-backed claim available for later withdrawal.

The regression uses only wrapper instructions under normal initialization. It creates independent source claims in two assets, realizes an asset-1 B loss, and checks that settlement preserves the unrelated asset-0 claim while retiring the asset-1 source claim first.

## TDD history

- `831a69d6`: test-only commit on wrapper `main`, pinned to engine `143e68c4917...`; the public LiteSVM regression fails because the asset-0 claim is consumed.
- `5b182282`: advances only the engine pin to `74e9bc5e8f41...`, where [engine PR #144](https://github.com/aeyakovenko/percolator/pull/144) fixes source-domain selection and sparse physical-slot traversal.

## Verification

- auth-matcher and hostile-matcher fixture SBF builds
- engine `cargo test --all-targets --features fuzz`: 164 tests passed
- focused engine source-lien Kani contracts passed

Draft because this PR depends on engine PR #144 and its base dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=74e9bc5e8f41dbd929ad6866fb0f638d856c421d#74e9bc5e8f41dbd929ad6866fb0f638d856c421d"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "74e9bc5e8f41dbd929ad6866fb0f638d856c421d" }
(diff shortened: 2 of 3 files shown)


aeyakovenko:
The source-domain accounting defect is reproducible, but the current retained regression does not yet prove the PR title's terminal LoF classification.

On the current Meta integration pins (`percolator c1a058f`, `percolator-prog cf69b40`), an asset-1 B settlement does incorrectly retire asset-0 claim face first. Your fix makes that accounting assertion pass. However, the public terminal extraction attempt is blocked by a distinct confirmed DoS: after bankruptcy reduces effective OI and a later bilateral close reaches zero OI, a normal-epoch residual basis leg cannot be cleared by live, recovery, or resolved paths. The clean LiteSVM reproduction is `aeyakovenko/percolator-meta@9c75c9d`; details are on engine #118 and program #219.

That means the current test proves cross-domain liability corruption, not yet an independently backed provider's realized token loss. In addition, the SBF I built from this PR stack (SHA-256 `e5edc9df8b2dbed6ec9fc67cb26d5ee54cbd51d331c5ad68befd549d122d8495`) drops the source-capacity admission fix in `c1a058f` and reopens the public terminal DoS covered by Meta PR #275.

Please stack/rebase on the source-capacity fix and the bilateral-zero-OI repair, then retain a public LiteSVM test that reaches provider-backing consumption or another independent victim net loss. Until that terminal proof exists, this is a confirmed accounting bug / latent LoF, not a demonstrated live custody LoF.

aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-035 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #282, opened 2026-07-19
[BLOCKER LoF] Price pending EWMA overrides at target notional

aeyakovenko:
## Impact

A public trader could cheaply overwrite a legitimate pending EWMA rebound while the engine circuit-bounded effective price still lagged far below the stored target. The attacker could then realize the displaced PnL against independent open interest.

The exact-parent LiteSVM regression uses only normal public instructions, ordinary SPL accounts, a separate honest oracle authority, and the repository authorized echo matcher. It reproduces through all four trade routes.

On `TradeNoCpi` and one-leg `BatchTradeNoCpi`:

- attacker coalition deposits: `24,000,000,000`
- attacker coalition withdraws: `24,437,499,800`
- independent victim deposits: `20,000,000,000`
- independent victim withdraws: `19,500,000,000`
- movement fee paid: `62,500,200`

On `TradeCpi` and one-leg `BatchTradeCpi`, the normal matcher echoes the lagging engine effective price:

- attacker coalition withdraws: `24,874,999,800`
- independent victim withdraws: `19,000,000,000`
- movement fee paid: `125,000,200`

That is up to `874,999,800` net attacker profit and `1,000,000,000` independent-victim loss without a hostile matcher.

## Root Cause

The wrapper computed EWMA movement fees from existing OI valued at `asset.effective_price`. The movement itself changes `profile.mark_ewma_e6`. When an honest target rebound was pending, the lagging effective price understated the value affected by overwriting that target.

## Fix

Value existing OI at `max(asset.effective_price, profile.mark_ewma_e6)` when deriving the mark externality notional. Existing collected-fee logic then clamps the internal target move without rejecting the trade.

After the fix, all four public routes remain executable. The no-CPI coalition withdraws `23,968,999,800`, losing `31,000,200`; its `62,000,200` collected fee covers `31,000,000` displaced victim PnL. The CPI coalition withdraws `23,937,999,800`, losing `62,000,200`; its `124,000,200` fee covers `62,000,000` displaced victim PnL.

The original test-only commit `11d11c27` is red on its exact parent. Commit `6c302f3e` extends the same exact-parent reproduction through the normal CPI matcher; both CPI routes are red there and green at the fixed head. The attack also remains red when composed with wrapper PR #260 and engine PR #138, so it is not duplicate pending coverage.

## Verification

- built `tests/fixtures/auth_matcher` and `tests/fixtures/hostile_matcher`
- `cargo test --all-targets -- --test-threads=1`: 6 unit + 566 LiteSVM/CU tests pass before the CPI route extension
- focused four-route LiteSVM economic regression passes after the extension
- `cargo kani --tests --harness kani_v16_fee_supported_mark_clamp_is_directional_and_zero_support_is_noop`
- `cargo kani --tests --harness kani_v16_collected_base_fee_cannot_fund_mark_movement`
- `git diff --check`

`cargo kani --tests` was also attempted, but the pre-existing unconstrained `kani_v16_unknown_or_truncated_tags_reject` decoder harness remained in CBMC symbolic execution after 15 minutes and was stopped. The two pricing/fee harnesses relevant to this change verify successfully.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -11624,7 +11624,11 @@ pub mod processor {
         let effective_price = asset.effective_price.get();
         let trade_notional = trade_fee_notional_ceil(size_q_abs, accepted_exec_price)?;
         let max_side_oi_q = core::cmp::max(asset.oi_eff_long_q.get(), asset.oi_eff_short_q.get());
-        let max_side_notional = risk_notional_ceil(max_side_oi_q, effective_price)?;
+        // The trade moves the stored target, so value existing OI at the larger of that target
+        // and the lagging engine price. Otherwise a wash trade can cheaply overwrite an honest
+        // rebound while the circuit breaker still holds effective_price near the prior low.
+        let externality_price = core::cmp::max(effective_price, profile.mark_ewma_e6);
+        let max_side_notional = risk_notional_ceil(max_side_oi_q, externality_price)?;
         let mark_externality_notional = core::cmp::max(max_side_notional, trade_notional)
             .checked_mul(2)
             .ok_or(PercolatorError::EngineArithmeticOverflow)?;
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -59717,3 +59717,259 @@ fn v16_attack_underfunded_exit_cannot_move_ewma_with_uncollectible_fee() {
         assert_underfunded_ewma_exit_uses_collected_fee(path);
     }
 }
+
+// Public regression: a legitimate EWMA round trip can leave the authenticated target far above
+// the bounded engine effective price. A public wash trade must not be able to buy down that pending
+// target for less than the PnL it removes from unrelated open interest.
+#[test]
+fn v16_attack_trade_cannot_underpay_to_override_pending_ewma_target() {
+    #[derive(Debug)]
+    struct Outcome {
+        attacker_withdrawn: u128,
+        victim_withdrawn: u128,
+        movement_fee: u128,
+        low_price: u64,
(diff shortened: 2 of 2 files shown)


aeyakovenko:
Clean-room Meta LiteSVM confirmation: this is a **real public LoF**.

Probe: [`56c5519`](https://github.com/aeyakovenko/percolator-meta/commit/56c5519) / [Meta draft PR #278](https://github.com/aeyakovenko/percolator-meta/pull/278).

The independent probe uses one honest directional pair, one attacker-controlled wash pair, a separately keyed honest EWMA authority, bounded public cranks, and actual `ResolveMarket -> CloseResolved -> ClaimResolvedPayoutTopup` token payouts.

- RED on exact pinned program `cf69b40e` / engine `c1a058f3` SBF (`sha256 15e43cb3...`): attacker deposits `24,000,000,000` and withdraws `24,437,499,800`; victim deposits `20,000,000,000` and receives `19,500,000,000`; movement fee is `62,500,200`; honest target `10,000,000` is overwritten to `9,500,000`.
- GREEN on a clean-room legacy-entrypoint SBF built from this PR's exact head `c5b06fe5` (`sha256 5dcbb66f...`): the same trades and resolved payout lifecycle complete, but coalition withdrawals no longer exceed deposits.

This PR is based on program `da64be63` / engine `143e68c`, while Meta's audited branch is on `cf69b40e` / `c1a058f3`. Please rebase the fix onto that current stack before merge; repinning Meta to this head as-is would discard later source-domain capacity fixes.

aeyakovenko:

PR #284, opened 2026-07-20
[BLOCKER LoF] Route negative batch fees to the correct insurance side

aeyakovenko:
## What changed

- Route collected `BatchTradeNoCpi` fees by the signed leg's long/short orientation before crediting per-side insurance budgets.
- Add a public LiteSVM regression that compares identical negative single and batch trades through terminal settlement.

## Root cause and impact

The engine's batch outcome reports `fee_a` and `fee_b` by logical account identity. The wrapper treated those fields as `fee_long` and `fee_short`. For a negative leg, account A is the short side and account B is the long side, so the wrapper credited both fees to the opposite insurance domains.

On the exact pre-fix parent, the public regression produces:

- single route: victim payout 1,000,000; long/short budgets 1,000,000 / 1
- batch route: victim payout 0; long/short budgets 1 / 1,000,000

The sequence uses a normally initialized live market, public instructions only, and a separate honest oracle. The independent long winner loses its entire 1,000,000 capital because the batch route put the collected backstop in the wrong side.

## Stack

This PR targets `codex/fix-resolved-adl-winner-close` and is intentionally stacked on #216. That engine pin is required for the terminal winner to exercise its public resolved-close path; this PR's own diff is two commits and only changes signed batch fee routing plus its regression.

## Verification

- rebuilt all three matcher SBF fixtures
- `cargo test --all-targets`: 567/567 LiteSVM/CU tests pass
- targeted regression passes on fixed head
- `cargo kani --tests`: full run in progress; draft will be updated with the final result

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -6136,8 +6136,8 @@ pub mod processor {
             let mut remaining_fee_a = outcome.fee_a;
             let mut remaining_fee_b = outcome.fee_b;
             let mut cfg_dirty = false;
-            for (asset_index, oracle_profile, fee_basis_price, fee_quote, abs_size) in
-                leg_ctx.iter_mut()
+            for ((asset_index, oracle_profile, fee_basis_price, fee_quote, abs_size), request) in
+                leg_ctx.iter_mut().zip(requests.iter())
             {
                 let requested_fee_leg =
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #286, opened 2026-07-20
[BLOCKER LoF] Preserve payouts after post-snapshot backing expiry

aeyakovenko:
## What changed

- Pin the engine fix from `percolator#147`.
- Add a public LiteSVM lifecycle regression covering two independent zero-sum markets, out-of-order source-backing expiry, resolved close, receipt top-up, portfolio closure, and slab closure.
- Enforce the existing custody CU guardrail on every resolved-close continuation.

## Exact-parent RED

On the exact parent pin `5997d84eb00c21cb28c49f05e5e2786f5ee2e54e`, ordinary public trades create two 40-atom winner claims. One counterparty backing bucket expires before payout snapshot capture and the other expires afterward.

The parent freezes the ledger at `snapshot=40`, `rate=40/80`. Both losers receive 960, both independent winners receive only 1,020 instead of 1,040, and 40 atoms remain in the engine vault. No state bytes are injected and no dishonest oracle value is used.

## Fixed-head GREEN

The expiry now raises the common snapshot to 80 and the rate to 80/80. Both losers receive 960, both winners receive 1,040, and all 4,000 deposited atoms are conserved. The test then closes all five public portfolios and the market slab with a zero engine vault.

## Root cause

Fresh source backing is excluded from junior residual as a senior claim. When expiry releases that backing after payout-snapshot capture, the engine previously refreshed aggregate backing state but left the frozen payout ledger unchanged. Terminal payout therefore depended on which winner caused expiry first.

## Verification

- Exact-parent public LiteSVM RED and fixed-head GREEN
- `cargo build-sbf --tools-version v1.52` for the program and all three matcher fixtures
- `cargo test --all-targets`: 5 unit + 571 LiteSVM/CU tests pass
- Engine `cargo test --all-targets`: 50 unit + 10 backing fuzz + 2 insolvent payout fuzz + 75 spec tests pass
- Focused engine Kani: 99 checks, 2/2 covers

Engine fix: https://github.com/aeyakovenko/percolator/pull/147

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=5997d84eb00c21cb28c49f05e5e2786f5ee2e54e#5997d84eb00c21cb28c49f05e5e2786f5ee2e54e"
+source = "git+https://github.com/aeyakovenko/percolator?rev=090589679cad025c5e0a06c5aba61123d8eef395#090589679cad025c5e0a06c5aba61123d8eef395"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_post_snapshot_expiry_rejects_stale_trade_then_owner_progresses`. PR286 is stacked on the live source-backing expiry fix; on the pinned program its authenticated public close rejects and rolls back before the post-snapshot payout state can commit. The executable disposition remains in PR #135; no separate production fix is required.

PR #287, opened 2026-07-20
[BLOCKER LoF] Preserve funded shutdown reclaim windows

aeyakovenko:
## Impact

On the composed shutdown-escheat and live-policy stack (#226 plus #257), a non-oracle market authority could retroactively shorten a provider reclaim window after shutdown. The exact-parent public LiteSVM reproduction:

- creates a normal permissionless asset under a 100-slot shutdown delay;
- removes the base init-fee insurance so the backing aggregate is the only guard input;
- has an independent provider fund 500 backing atoms through `TopUpBackingBucket`;
- shuts the asset down, changes the live delay from 100 slots to 1, and invokes mature fallback cleanup;
- withdraws the escheated base insurance through the canonical asset-0 operator route.

The RED test completes the extraction and fails with the attacker destination holding exactly 500 atoms. It uses no state injection, dishonest oracle, malformed account, or initialization footgun.

## Fix

Reject retroactive `force_close_delay_slots` reductions while any engine-maintained provider-value aggregate remains nonzero:

- recoverable source backing principal;
- backing-provider earnings;
- remaining domain insurance budget.

The check is O(1), does not scan assets or domains, and leaves first configuration and delay increases unchanged. The regression proves rejected state rolls back byte-for-byte, the provider can reclaim all 500 atoms under the original window, and the same reduction succeeds after every protected residual is withdrawn.

## Stack

This draft is based on #226 because the exploit targets its mature-shutdown escheat fallback. It also contains the production change from #257, which is required to test the composed pending behavior. Rebase it after those dependencies land.

## Verification

- exact-parent public LiteSVM RED: attacker receives 500 atoms
- fixed public LiteSVM GREEN
- adjacent shutdown-escheat tests: green
- adjacent resolve-policy tests: green
- `cargo build-sbf --no-default-features`: green
- 6 wrapper unit tests: green
- 568 LiteSVM/CU tests: green
- full `cargo kani --tests`: did not complete; the generic `Instruction::decode` proof kept increasing its unwind bound through iteration 167 until manually terminated. No property failed, but this is not claimed as a green Kani run.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9683,6 +9683,16 @@ pub mod processor {
         {
             return Err(PercolatorError::EngineLockActive.into());
         }
+        if cfg.force_close_delay_slots != 0
+            && force_close_delay_slots < cfg.force_close_delay_slots
+            && (group.header.source_fresh_backing_total_num.get() != 0
+                || group.header.backing_provider_earnings_total.get() != 0
+                || group.header.insurance_domain_budget_remaining_total.get() != 0)
+        {
+            // Mature-shutdown fallback escheats abandoned provider value. Do not let marketauth
+            // retroactively shorten the reclaim window while any such value remains entrusted.
+            return Err(PercolatorError::EngineLockActive.into());
+        }
         cfg.permissionless_resolve_stale_slots = stale_slots;
         cfg.force_close_delay_slots = force_close_delay_slots;
         drop(group);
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_privileged_policy_boundary_matrix_rejects_untrusted_callers`. The reclaim-window rewrite requires the configured market authority; an untrusted public caller rejects with exact rollback. The executable disposition remains in PR #135; no separate production fix is required.

PR #288, opened 2026-07-20
[BLOCKER DoS] Keep claimant Recovery deadlines monotonic

aeyakovenko:
## Impact

A non-oracle market authority could extend an already-running Recovery force-close deadline after ordinary elapsed maintenance consumed all deposited principal, even while a user still held positive PnL behind unresolved positions. Raising a five-slot delay to ten million slots blocked the honest force-close path and left the winner's 100-atom claim unconvertible; owner forfeit removed the leg but did not release the payout.

The exact-parent LiteSVM reproduction uses only public instructions and separate roles: two users deposit and trade, an honest AuthMark signer moves the price, public maintenance syncs drain both principals, marketauth shuts down the asset and extends the deadline, and an independent cranker attempts the original force close. No state bytes or oracle fixtures are injected.

## Root cause

The funded-deadline guard used `c_tot != 0` as its only proxy for in-flight user value. Maintenance can legitimately make `c_tot == 0` while positive PnL and stored positions remain. The engine already tracks those payout barriers exactly in `resolved_payout_blocker_count`.

## Fix

Reject force-close deadline increases while either aggregate capital or the engine-maintained unresolved-payout blocker count is nonzero. The added check is O(1), permits shorter deadlines, and still permits longer delays after all user claims and blockers are settled.

## Validation

- exact-parent public LiteSVM RED: extension lands, original force close fails, conversion fails before and after owner forfeit
- fixed public LiteSVM GREEN: extension rolls back byte-for-byte, original deadline closes the pair, winner converts 100 atoms
- existing funded-deadline and post-settlement reconfiguration controls remain green
- `cargo test --all-targets`: 6 unit + 568 LiteSVM/CU tests green
- wrapper Kani suite: pending

Stacked on #257 so this PR contains only the zero-capital claimant refinement.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9744,7 +9744,8 @@ pub mod processor {
         }
         if cfg.force_close_delay_slots != 0
             && force_close_delay_slots > cfg.force_close_delay_slots
-            && group.header.c_tot.get() != 0
+            && (group.header.c_tot.get() != 0
+                || group.header.resolved_payout_blocker_count.get() != 0)
         {
             return Err(PercolatorError::EngineLockActive.into());
         }
--- a/tests/v16_cu.rs
+++ b/tests/v16_cu.rs
@@ -59912,6 +59912,193 @@ fn v16_attack_live_policy_cannot_extend_recovery_force_close_deadline() {
     );
 }
 
+// PARTIAL DoS: maintenance can legitimately consume every principal atom while a non-flat
+// portfolio still owns positive PnL. The Recovery deadline therefore has to remain immutable while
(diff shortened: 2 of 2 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_privileged_policy_boundary_matrix_rejects_untrusted_callers`. The claimant-deadline rewrite requires the configured market authority; an untrusted public caller rejects with exact rollback. The executable disposition remains in PR #135; no separate production fix is required.

PR #148, opened 2026-06-15
[codex] Reject stale matcher return replay

aeyakovenko:
## Summary

Fixes a stale matcher-context replay in the single `TradeCpi` path.

`TradeCpi` reads the matcher return from the persistent matcher context account. The old request id was `current_slot + 1`, so two same-slot requests with the same tuple could share the same id. If a matcher returned `Ok` without writing a fresh return prefix, the wrapper could accept the previous context return and apply a second fill.

This PR derives the matcher request id from the authenticated slot plus the pre-CPI matcher context return prefix and context key. A consumed return prefix therefore cannot validate as the next request unless the matcher writes a fresh response echoing the new request id.

## Tests

- `cargo build-sbf --no-default-features`
- `cd tests/fixtures/hostile_matcher && cargo build-sbf`
- `cargo test v16_attack_hostile_matcher_no_write_cannot_replay_stale_tradecpi_return -- --nocapture`
- `cargo test v16_attack_hostile_matcher -- --nocapture`
- `cargo test v16_bpf_tradecpi -- --nocapture`
- `cargo test v16_bpf_batch_trade_cpi_14_legs_under_tx_limit -- --nocapture`

Note: `cargo fmt --check` reports broad pre-existing formatting drift in `tests/v16_cu.rs`; I did not apply unrelated rustfmt churn.

PR #290, opened 2026-07-21
[BLOCKER LoF] Reject cross-domain bankruptcy insurance drain

aeyakovenko:
## Finding

A public cross-margin sequence could create debt on asset 1, flatten that loss-bearing leg through the normal risk-reducing route, and retain an unrelated asset-0 leg. The engine retained only aggregate negative PnL, so later permissionless liquidation inferred the debt domain from the surviving leg and spent an independent asset-0 insurance provider's funds.

On exact wrapper parent `da64be639168b496833dd4c701607a94fcb06b6c` with engine `143e68c4917ed0400a27b952f036a5677047cd84`, the retained LiteSVM RED test shows:

- unrelated asset-0 insurance spent: `100100`
- attacker-controlled payout: `115300`
- attacker net gain: `95100`
- bounded public progression: one liquidation call and resolved-close calls `(0, 220, 1)`

The reproduction uses only public initialization, SPL deposits, zero-sum trades, authenticated mark updates, maintenance sync, owner-authorized risk reduction, permissionless crank, resolve, and resolved close. It does not mutate account state or assume a dishonest oracle.

## Fix

This PR pins engine commit `ab111c7e746de307e02b1caefb99dd5de7611518` from engine PR #148 and retains the public regression. The engine persists negative-PnL provenance by insurance domain and settles principal, insurance, and terminal losses against that recorded source after the originating leg exits.

Two existing synthetic security fixtures now initialize the new provenance field so their injected negative-PnL state remains structurally valid. The resolved deep-insolvency lifecycle test performs one bounded public crank per recorded debt domain, matching the engine's one-action-per-call liveness contract.

## Verification

- `cargo build-sbf --tools-version v1.52`: passed against immutable engine pin `ab111c7e`.
- `cargo test --all-targets`: 6 unit tests and 566 LiteSVM/CU tests passed.
- The new public exploit regression and four cross-margin/liveness-sensitive targeted tests passed.
- Wrapper Kani compiled successfully. Three concrete decoder harnesses passed, including the all-active-payload truncation proof (`0/847` failed), and the symbolic one-byte unknown/truncated-tag harness passed separately with a checked unwind bound (`0/814` failed). The unchanged grouped oracle/asset trailing-byte proof did not terminate under bounded CaDiCaL, Kissat, or Bitwuzla attempts; no property failure was reported. This is not claimed as a full wrapper Kani pass, and neither wrapper production code nor `tests/v16_kani.rs` differs from `main`.
- Engine PR #148: 160 native/fuzz tests and eleven focused Kani harnesses passed on `ab111c7e`. The full 246-engine-harness roster was not rerun.

## Dependency

Draft until a final engine commit from aeyakovenko/percolator#148 is merged. The wrapper pin is immutable and can be updated to the merge commit if GitHub changes it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=ab111c7e746de307e02b1caefb99dd5de7611518#ab111c7e746de307e02b1caefb99dd5de7611518"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "ab111c7e746de307e02b1caefb99dd5de7611518" }
 anchor-lang-v2 = { git = "https://github.com/solana-foundation/anchor", rev = "e3cf760826fa0a60f247635ea0572e59861ec9b5", package = "anchor-lang-v2", default-features = false, features = ["alloc", "guardrails"], optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
 # Host tests use the same zero-copy account/view API as the SBF/program target.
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Strengthened the regression in `af5646c1` with a second public-interface exploit schedule for the same engine fix.

The new LiteSVM test creates unequal independent winners and an insolvent loser through deposits, trades, authenticated mark pushes, resolve, and `CloseResolved`, then exercises both winner close orders. On the exact pre-fix engine (`143e68c`), a winner could detach before the loser booked bankruptcy; the remaining winner became the sole B weight (small-first observed `b_long_num = 66_665_333_333_333_333_333`), shifting terminal loss by transaction order. With pinned fix `ab111c7e`, both schedules leave `(b_long_num, b_short_num) == (0, 0)` because Resolved bankruptcy is already represented by the payout haircut.

Exact-branch verification after rebuilding the program plus both matcher fixtures: `cargo test --all-targets -- --test-threads=1` passed 6 unit tests and all 567 LiteSVM/CU tests.

aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-034 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #292, opened 2026-07-22
[HARDENING LoF] Bind rebalance reductions to portfolio incarnations

aeyakovenko:
## What changed

- add a public LiteSVM regression that closes and recreates the same portfolio address, then replays an owner-signed `RebalanceReduce` from the old incarnation
- bind tag 44 to the program-assigned `portfolio_id`
- reject a mismatched incarnation before entering engine mutation
- reject the legacy unbound 19-byte payload and document the wire-format change

## Root cause and impact

`RebalanceReduce` authorized irreversible exposure removal but authenticated only the owner, portfolio address, asset, and amount. A transaction signed for an old portfolio incarnation remained valid after that address was publicly closed and reinitialized.

The exact-parent RED test uses only public program/System Program transitions. At a temporary authenticated mark move from 100 to 50, the stale consent closes the replacement victim position. When the mark returns to 100, terminal payouts are victim 950,000 and independent attacker 1,050,000. The replacement portfolio has a different program-assigned ID (1 -> 3), so binding that ID prevents the replay.

## Validation

- exact-parent RED at `408daae0` against parent `da64be63`
- `cargo test --all-targets`: 6 unit + 566 LiteSVM/CU tests pass
- `cargo kani --tests --harness kani_v16_recovery_close_progress_decode_preserves_wire_fields`
- `cargo kani --tests --harness kani_v16_legacy_unbound_rebalance_payload_is_rejected`
- `git diff --check`

This intentionally changes the tag-44 payload ABI; clients must include the current portfolio incarnation ID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,7 +481,8 @@ This section describes intent and operational ordering, not argument-by-argument
 - **ForfeitRecoveryLeg** (tag 43)
   - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget
 - **RebalanceReduce** (tag 44)
-  - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector
+  - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector;
+    the payload binds consent to the program-assigned `portfolio_id`
 - **ClaimResolvedPayoutTopup** (tag 46)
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #293, opened 2026-07-22
[HARDENING LoF] Bind RebalanceReduce to market generations

aeyakovenko:
## What changed

- add the current asset `market_id` to the owner-signed `RebalanceReduce` wire payload
- reject a retained tag-44 instruction with `AssetGenerationMismatch` before entering the engine
- prove wire round-trip and rejection of the legacy unbound payload with Kani
- retain an end-to-end LiteSVM RED/GREEN exploit test with exact error and rollback assertions

## Root cause and impact

`RebalanceReduce` was owner-signed but did not identify the market incarnation. After a market was closed and publicly reinitialized at the same address, its portfolio allocator restarted. A transaction retained from the old market could therefore execute against a replacement portfolio at the same address and reduce fresh exposure at an attacker-selected later mark.

The exact-parent RED commit is `ac577b59`. It recreates the market and both portfolios at the same public addresses, reopens fresh exposure, replays the old signed reduction at an adverse mark, and demonstrates an independent-victim transfer: victim payout 950,000; attacker payout 1,050,000.

## Dependency and interaction

This is intentionally based on #231, which supplies the persistent market-generation PDA and changing `market_id` across full market recreation.

#292 independently binds `RebalanceReduce` to `portfolio_id`. Neither boundary subsumes the other. After #231 lands, this PR and #292 must be rebased/combined so tag 44 binds both `market_id` and `portfolio_id`.

## Verification

- `cargo test --all-targets`: 5 unit + 569 LiteSVM/CU tests passed
- `cargo kani --tests --exact --harness kani_v16_recovery_close_progress_decode_preserves_wire_fields`
- `cargo kani --tests --exact --harness kani_v16_legacy_unbound_rebalance_market_generation_payload_is_rejected`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,6 +496,7 @@ This section describes intent and operational ordering, not argument-by-argument
   - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget
 - **RebalanceReduce** (tag 44)
   - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector
+  - binds the current asset `market_id`; retained consent from a retired market generation rejects
 - **ClaimResolvedPayoutTopup** (tag 46)
   - permissionless resolved-payout top-up claim; pays only the stored owner receipt token account
 
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2708,6 +2708,7 @@ pub mod ix {
         },
         RebalanceReduce {
             asset_index: u16,
+            market_id: u64,
             reduce_q: u128,
         },
         FinalizeResetSide {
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Independent clean-room Meta confirmation: aeyakovenko/percolator-meta#355 at commit 42127bf reproduces a 50,000-atom replacement-victim transfer on pinned program 19f3b494 using only public full-lifecycle instructions. The identical test passes on 8a0743ab with exact AssetGenerationMismatch, byte-for-byte rollback, and equal 1,000,000 payouts. This also confirms #231 is required to persist the namespace across CloseSlab.

aeyakovenko:

PR #295, opened 2026-07-23
[HARDENING LoF] Bind recovery forfeits to market generations

aeyakovenko:
## What changed

- Add the current asset `market_id` to the owner-signed `ForfeitRecoveryLeg` payload.
- Reject consent from a retired market generation before entering the engine mutation.
- Prove wire round-trip preservation and rejection of the legacy unbound payload with Kani.
- Retain a public LiteSVM RED/GREEN lifecycle with exact rollback assertions.

## Root cause and impact

`ForfeitRecoveryLeg` is irreversible consent to discard one portfolio's recovery claim, but its payload identified only the market address and asset slot. After `CloseSlab` and public `InitMarket` reuse of the same address, a transaction signed for the retired market could execute against a replacement portfolio and position.

The exact-parent RED commit is `e3310eca`. It closes and recreates the market and portfolios through public instructions, changes the persistent asset generation from 1 to 3, opens a fresh winning position, and replays the old victim-signed forfeit. The replay discards 100,000 of replacement-victim PnL, pays the victim only 1,000,000 principal, leaves exactly 100,000 in the canonical vault after all portfolios close, and makes public `CloseSlab` fail with `EngineLockActive`.

The fix is `cf15ff73`. A stale tag-43 request now fails with `AssetGenerationMismatch` and SVM rollback leaves the market and victim portfolio byte-identical.

## Dependency

This PR is intentionally based on #231, which supplies the persistent market-generation PDA and changing `market_id` across full market recreation. #278 independently binds the same instruction to `portfolio_id`; after both dependencies land, tag 43 must carry both boundaries.

## Verification

- Exact-parent public LiteSVM RED, then GREEN with exact error and rollback checks
- `cargo test --all-targets`: 5 unit and 569 LiteSVM/CU tests passed
- `cargo kani --tests --exact --harness kani_v16_recovery_close_progress_decode_preserves_wire_fields`: 1,524 checks, 0 failures
- `cargo kani --tests --exact --harness kani_v16_legacy_unbound_forfeit_market_generation_payload_is_rejected`: 831 checks, 0 failures
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,7 @@ This section describes intent and operational ordering, not argument-by-argument
   - owner-signed close recovery path; optional deposit is transferred first, then the engine cancels the pending close if the cure succeeds
 - **ForfeitRecoveryLeg** (tag 43)
   - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget
+  - binds the current asset `market_id`; consent from a retired market generation rejects
 - **RebalanceReduce** (tag 44)
   - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector
 - **ClaimResolvedPayoutTopup** (tag 46)
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2704,6 +2704,7 @@ pub mod ix {
         },
         ForfeitRecoveryLeg {
             asset_index: u16,
+            market_id: u64,
             b_delta_budget: u128,
         },
         RebalanceReduce {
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Independent Meta LiteSVM confirmation: aeyakovenko/percolator-meta#354 at commit f420787 reproduces the public replay on pinned program 19f3b494 (victim loses 250,000 atoms; 250,000 remains and blocks CloseSlab) and passes unchanged against cf15ff73 with exact AssetGenerationMismatch plus byte-for-byte rollback. The test also verifies that persistent generations from #231 are required for the bound field to distinguish full slab recreation.

aeyakovenko:

PR #296, opened 2026-07-23
[BLOCKER LoF] Bind trade-fee policy to market generations

aeyakovenko:
## Summary

- bind `UpdateTradeFeePolicy` to asset 0's persistent `market_id`
- reject retained policies from retired market incarnations before mutation
- fail closed on the legacy tag-55 payload
- add an exact-parent public LiteSVM RED/green exploit regression and focused Kani ABI proofs

## Impact

A fee policy signed for a retired market could be retained and replayed after the same market address was publicly closed and reinitialized. The attacker could land the old 10,000 bps policy while the replacement market was empty, bypassing the funded-market policy guard, then land already-signed deposits and a zero-fee trade. A permissionless replacement-asset operator could withdraw both resulting fee domains and gain exactly 1,000 units taken from an independent victim in the regression.

This PR is stacked on #231 because that PR supplies the persistent market-generation account and generation-bound transaction model. The bug is distinct from #256: the stale policy lands while `c_tot == 0`.

## Root cause

Tag 55 authenticated the current asset-0 insurance authority but did not bind the signed intent to the market incarnation for which it was authorized.

## Verification

- exact-parent RED on `4c5753f0`: public LiteSVM exploit extracts 1,000 victim units
- `cargo test --all-targets`: 5 unit + 569 LiteSVM/CU tests pass
- Kani field-preservation harness: 823 checks pass
- Kani legacy-payload rejection harness: 831 checks pass
- Kani admin-policy trailing-byte harness: 1,519 checks pass

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2612,6 +2612,7 @@ pub mod ix {
             insurance_share_bps: u16,
         },
         UpdateTradeFeePolicy {
+            market_id: u64,
             trade_fee_base_bps: u64,
         },
         UpdateFeeRedirectPolicy {
@@ -2880,6 +2881,7 @@ pub mod ix {
                     insurance_share_bps: read_u16(&mut rest)?,
                 },
                 55 => Self::UpdateTradeFeePolicy {
+                    market_id: read_u64(&mut rest)?,
                     trade_fee_base_bps: read_u64(&mut rest)?,
                 },
                 58 => Self::UpdateFeeRedirectPolicy {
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Independent clean-room Meta confirmation: https://github.com/aeyakovenko/percolator-meta/pull/357

The upstream regression's direct replacement-account injection was removed from the confirmation path. The Meta probe closes generation A through public Percolator/SPL lifecycle instructions, invokes same-address public initialization under LiteSVM's retained zero-lamport account model, enables the configured one-atom permissionless append path, and creates/funds both generation-B portfolios without injecting market or portfolio bytes.

Against pinned program `19f3b494049b2dfcbf8881366443c611c4e09290` / engine `4bf72ea3f9bea8682fe23b5c6fff9e04b5fb41d3` (SBF `e4948402cbd85b58d0de6ad57550da9ab20aed8ec5d494540cda26fb1d46f2ab`), the stale policy charges 1,000 atoms to each pre-signed side. The permissionless asset operator withdraws both 1,000-atom domains, ending 1,000 ahead while an independent victim ends 1,000 down; the remaining 18,001 vault atoms exactly cover both 9,000-atom portfolio claims plus the one-atom init fee.

The identical probe passes against this PR composed with persistent generation PR #231 at `e748ff18f113e7e39902378d8bd7d6e88731e923` (SBF `94a543366c3f7d5c9731b55db37205616e1c3912630444a08365dae1ba011178`): replay rejects with exact `AssetGenerationMismatch` / `Custom(30)`, market bytes roll back, both deposits and the generation-bound zero-fee trade remain executable, both users retain 10,000, and asset-1 insurance remains zero.

aeyakovenko:

PR #291, opened 2026-07-22
[REAL DoS] Pin lapsed backing settlement fix

aeyakovenko:
## Finding
A portfolio can remain permanently uncloseable after valid market resolution when its source backing expires at the resolution slot. The engine attempted K/F settlement before materializing the lapsed backing transition, so `CloseResolvedAccount` repeatedly returned stale/budget errors while the resolved lifecycle exposed no independent rescue path.

## Change
Pin both target and host builds to engine commit `e3ea7251ed75f0111ba3cf5a4b87961f6c7cca63` (aeyakovenko/percolator#149). No wrapper instruction, authority, custody, amount, or destination behavior changes.

## TDD and verification
- Engine RED: `b4b087025bd708d012e46bbf8ddd75ff995734ff`
- Engine GREEN: `e3ea7251ed75f0111ba3cf5a4b87961f6c7cca63`
- Wrapper: 6 unit + 567 `v16_cu` tests passed
- Rebuilt SBF SHA-256: `ba6b50a9b56a7ff678e05473ebf8ed4c01dc9c84cb541342d534e70ccb2c8fb6`
- Meta LiteSVM RED/GREEN covers public init, backing deposit and expiry, positions, crank, controller resolution, and bounded resolved close retries with the real wrapper binary.

This PR is intentionally pin-only and stacked on the current wrapper audit head.

PR #297, opened 2026-07-23
[BLOCKER DoS] Prove prospective K/F source expiry progress

aeyakovenko:
## Finding

This public LiteSVM TDD reproduction proves a persistent resolved-close DoS after the fix in #291: asynchronous K snapshots can leave positive prospective K/F settlement against a Fresh backing domain that lapses before either portfolio records that source domain.

The exact-parent RED uses only public instructions and normal initialized state: deposits, a trade, authenticated mark moves, permissionless cranks, resolution, and alternating honest `CloseResolved` calls. Every bounded close attempt reaches the same engine error and rolls back; one side retains its unresolved PnL and the other remains active, so neither honest caller can finish payout and withdrawal.

## Change

- Pin the stacked engine fix that expires one lapsed prospective K/F source before settlement.
- Keep the exact public lifecycle as the green regression.
- Add a 14-active-leg CU test that forces the preflight to inspect the full roster before finding the only positive source.
- Track every successful close in the lifecycle under the 1,000,000-CU instruction ceiling.

## Verification

- Exact-parent test is RED at `9cbae705` with engine `b6827a09`
- Engine production regression and liveness-rank Kani proof pass
- Full engine suite: 168 tests pass
- Full wrapper matrix: 5 unit + 575 LiteSVM/CU tests pass
- Existing 28-domain resolved-close CU tests pass after the flat-portfolio fast path

Stacked on #291; this is a distinct prospective-source blocker not handled by that PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=f87ca00a95b285aef5579639132991697fcbe7c5#f87ca00a95b285aef5579639132991697fcbe7c5"
+source = "git+https://github.com/aeyakovenko/percolator?rev=ca7f10c2c0e1cd93b3adb25cdd81973c26974cdf#ca7f10c2c0e1cd93b3adb25cdd81973c26974cdf"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_prospective_source_expiry_prerequisite_matrix_keeps_exit_live`. PR297 is stacked on the PR291 settlement-ordering fix; on the pinned predecessor the exact prospective-only expiry lifecycle closes both portfolios without rejection and returns all 200,000,000 deposited atoms. The executable disposition remains in PR #135; no separate production fix is required.

PR #298, opened 2026-07-23
[BLOCKER DoS] Expire prospective loss backing before resolved settlement

aeyakovenko:
## Finding

`CloseResolved` could settle a pending negative K/F delta into a new same-side capital-backed loss after that backing domain's `Fresh` term had already lapsed. The later backing add rejected with `LockActive`; SVM rollback erased every attempted transition, so one losing account could permanently hold the global resolved-payout gate closed for independent winners.

This is distinct from the existing-source ordering fix in #291 and the prospective positive-source fix in #297. The account has no persisted source claim for the preflight to inspect, and the pending delta is negative: settlement creates new same-side backing rather than a positive opposite-side claim.

## Public RED

The exact-parent LiteSVM regression uses only public program instructions and authenticated slots:

- two active assets and four independently owned portfolios
- ordinary deposits and `TradeNoCpi` fills
- authenticated mark movement that leaves one long with a prospective negative K delta
- a hedge that removes the matching positive-source continuation
- an ordinarily deposited source-backing bucket that lapses before resolution
- 16 alternating owner-signed `CloseResolved` attempts with fresh blockhashes

Against engine parent `b4901b93e72ba29dcedde7efd513b08eb43fb6ab`, every victim close returns engine `LockActive`, the state rolls back, and no bounded terminal route releases either the victim or the blocked payout. The new engine preflight expires exactly one lapsed prospective-loss domain and returns bounded progress before K/F settlement.

## Changes

- pin engine commit `68364eac72c47d73e07820608f08acb58e38b08f` from engine PR #151
- add the public two-market RED/GREEN lifecycle
- add a 14-leg last-slot scan proving the new preflight stays below the 1,000,000 CU budget

## Verification

- focused public LiteSVM regression: pass
- positive and negative 14-leg resolved-close CU regressions: pass
- `cargo test --all-targets --quiet`: 5 unit + 577 LiteSVM/CU tests pass
- `cargo fmt --check`
- engine PR #151: all 138 engine tests pass; focused Kani rank proof succeeds with 0/16 failures and all 3 covers reached

`cargo kani --tests` translated successfully and passed the exhaustive 823-case payload truncation harness. The unchanged aggregate unknown/truncated-tag decoder harness remains solver-bound, as already recorded on #291 and #297; this PR changes no wrapper decoder or production wrapper code.

Depends on engine PR #151 and wrapper PR #297.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=b4901b93e72ba29dcedde7efd513b08eb43fb6ab#b4901b93e72ba29dcedde7efd513b08eb43fb6ab"
+source = "git+https://github.com/aeyakovenko/percolator?rev=68364eac72c47d73e07820608f08acb58e38b08f#68364eac72c47d73e07820608f08acb58e38b08f"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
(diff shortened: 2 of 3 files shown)


aeyakovenko:

PR #299, opened 2026-07-23
[HARDENING LoF] Bind withdrawals to portfolio incarnations

aeyakovenko:
## Impact

A relayer can retain an owner-signed `Withdraw` for portfolio incarnation A, then submit it after the same owner closes A and initializes independently funded incarnation B at the same account address. The amount-only tag-4 payload cannot distinguish the two market-assigned portfolio IDs.

The clean-room Meta LiteSVM chain signs B's trade before replaying A's withdrawal. Against program `19f3b494049b2dfcbf8881366443c611c4e09290` (SBF `e4948402cbd85b58d0de6ad57550da9ab20aed8ec5d494540cda26fb1d46f2ab`), the replay removes 1,100,000 atoms of B margin, the already signed B trade still meets initial margin, and a later independent authenticated mark lets an unprivileged cranker extract and withdraw 498 atoms of victim-funded liquidation fee.

## Fix

- Add the expected monotonic `portfolio_id` to the signed `Withdraw` wire.
- Reject a mismatch with `EngineProvenanceMismatch` before engine accounting or token CPI.
- Preserve exit for pre-ID live portfolios through the existing reserved `u64::MAX` migration identity.
- Update wire proofs, BPF callers, migration coverage, and README.

No signer, authority, destination, token amount, custody path, or admin capability is added.

## TDD

RED on the exact parent:

```text
v16_attack_presigned_withdraw_rejects_reinitialized_portfolio
incarnation A's signed withdrawal must not debit incarnation B
```

GREEN on this commit:

```text
cargo test --test v16_cu
test result: ok. 566 passed; 0 failed
```

The rebuilt fixed SBF is `b955cdba9726e74efc838cfa10559309fd0f9b7b1f7499f5060fa94b6e1143db`. The identical Meta end-to-end test rejects the stale ID with byte-identical market, portfolio, and destination rollback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,7 @@ This section describes intent and operational ordering, not argument-by-argument
 - **Deposit** (tag 3)
   - transfers collateral into vault; credits engine balance for that account
 - **Withdraw** (tag 4)
+  - binds the signed payload to the current program-assigned portfolio incarnation ID
   - performs oracle-read + engine checks; withdraws from vault via PDA signer; debits engine
 - **ClosePortfolio** (tag 8)
   - closes a flat, empty portfolio account and sweeps its rent to the market slab
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -801,7 +801,7 @@ pub mod state {
     }
 
     #[inline]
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #300, opened 2026-07-23
[BLOCKER DoS] Prevent foreign expiry from stranding resolved source liens

aeyakovenko:
## What changed

- Pin the wrapper to engine commit `b9efec69740bbc5002035d64e8f98b7906f362ce` from engine PR #152.
- Add a public LiteSVM terminal lifecycle that creates two backed winners, gives one winner a legitimate source-credit initial-margin lien, lets the other winner expire the shared bucket first, and then retries the liened winner's sole terminal route.
- Assert SVM rollback on rejected retries, exact account-local live-to-impaired relabeling on the fixed engine, unchanged token custody, and a 1,000,000-CU ceiling.

## Public exploitability

The regression uses only initialized accounts and public instructions:

1. Four portfolios deposit and open two ordinary matched positions.
2. An authenticated mark move creates source-attributed positive PnL for both longs.
3. One long performs an ordinary risk increase that creates a 51-atom counterparty-backed IM lien.
4. The market resolves at the configured backing expiry.
5. The lien-free long closes first and moves the shared bucket to `Impaired`.
6. The liened long calls `CloseResolved`, the only user/keeper continuation available in Resolved mode.

On the exact parent pinned to engine `68364eac72c47d73e07820608f08acb58e38b08f`, all three retries fail with custom error `0x19` after about 13.7k CU and LiteSVM proves market/portfolio rollback each time. The account cannot trade, rebalance, withdraw, or use Live auto-crank after resolution, so an honest correctly-parameterized cranker has no alternate bounded path.

With engine #152, the first identical call succeeds, moves exactly the account-local counterparty lien and reserved effective amount to impaired state, leaves the already-impaired aggregate bucket and token vault conserved, and stays below the CU guardrail.

## Scope and dependencies

- Stacked on wrapper #298 and engine #152.
- Engine #152 is stacked on #151/#150 and shares helper semantics with pending Live-expiry PR #112; the helper will deduplicate when those stacks are rebased.
- Known terminal payout PR #145/#283 remains separate and is not duplicated here.

## Validation

- Exact-parent public LiteSVM RED: three rollback-clean `0x19` failures.
- Fixed targeted public LiteSVM GREEN.
- `cargo test --all-targets -- --quiet`: 5/5 unit and 578/578 LiteSVM/CU tests passed after installing the standard auth/hostile matcher fixtures.
- Wrapper Kani truncation harness: 823 checks passed. The unchanged aggregate unknown-tag harness remained unwind-bound and was stopped without a result.
- Engine full runtime suite: 171 passed.
- Engine targeted Kani proof: 245 checks passed, 3/3 covers satisfied.

Engine fix: https://github.com/aeyakovenko/percolator/pull/152

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=68364eac72c47d73e07820608f08acb58e38b08f#68364eac72c47d73e07820608f08acb58e38b08f"
+source = "git+https://github.com/aeyakovenko/percolator?rev=b9efec69740bbc5002035d64e8f98b7906f362ce#b9efec69740bbc5002035d64e8f98b7906f362ce"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "68364eac72c47d73e07820608f08acb58e38b08f" }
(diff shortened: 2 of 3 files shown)


aeyakovenko:

PR #301, opened 2026-07-23
[HARDENING LoF] Bind PnL conversion to portfolio incarnations

aeyakovenko:
## Summary

- add the market-assigned `portfolio_id` to `ConvertReleasedPnl` tag 28
- reject an ID mismatch before any engine conversion and reuse the explicit `u64::MAX` legacy-account migration path
- add a public-instruction LiteSVM regression and update every caller plus the wire proofs

## Root cause and impact

`ConvertReleasedPnl` authenticated only the owner, market, portfolio address, and amount. A relayer could retain an owner-signed transaction for portfolio incarnation A, wait for A to be closed and the same address and owner to initialize incarnation B, then replay the old conversion against B.

The exact-parent LiteSVM reproducer creates released PnL through normal deposits, backed trades, authenticated mark pushes, and permissionless cranks. After B withdraws all ordinary capital, the stale A conversion moves B junior PnL into fee-bearing capital. A permissionless maintenance crank then receives a positive reward funded by B. This is an independent victim loss, not point inflation or an initialization footgun.

## Security properties

- No new admin or DAO authority is introduced.
- The signed payload names one monotonic portfolio incarnation.
- Mismatches return `EngineProvenanceMismatch` before economic mutation.
- Rejection preserves market, portfolio, and vault bytes; the cranker receives zero.
- Legacy portfolios remain usable only with the explicit `u64::MAX` ID and are migrated through the existing binder.

## TDD evidence

- RED/exploit on exact parent `f72c8bdc076c107fd8085ae60a7bcd9dd03a0e6c`: retained A conversion succeeds against B and a public cranker receives a positive reward.
- GREEN on `15f6d588305f3ec66fd63ed841cde04df818dc04`: `v16_attack_presigned_convert_rejects_reinitialized_portfolio` passes against rebuilt real SBF.
- Full fixed suite: 6 unit tests and 567 LiteSVM tests passed.
- Fixed SBF SHA256: `05bacd937698db9a8acf3d8b77990b3229bc166712ec716a46c9ea9d0a1e729f`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,9 @@ This section describes intent and operational ordering, not argument-by-argument
 - **Withdraw** (tag 4)
   - binds the signed payload to the current program-assigned portfolio incarnation ID
   - performs oracle-read + engine checks; withdraws from vault via PDA signer; debits engine
+- **ConvertReleasedPnl** (tag 28)
+  - binds the signed payload to the current program-assigned portfolio incarnation ID
+  - converts currently released, source-backed junior PnL into senior portfolio capital
 - **ClosePortfolio** (tag 8)
   - closes a flat, empty portfolio account and sweeps its rent to the market slab
 - **CloseResolved** (tag 30)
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #302, opened 2026-07-23
[BLOCKER DoS] Route prospective loss after source impairment

aeyakovenko:
## What changed

- pin the wrapper to engine commit `0401193a8f277720529b7359e97bf1f6b659c475`
- add a public LiteSVM lifecycle regression for closing a prospective loser after a shared source-backing bucket becomes impaired

## Root cause

A lien-free resolved winner could expire a shared source-backing bucket while another account still held an unsettled negative K/F delta. When that losing account later crystallized its capital-backed loss, the engine tried to create fresh provider principal in the already-impaired bucket and returned `LockActive` forever. SVM rollback preserved the stuck state on every retry.

The engine fix routes newly crystallized loss directly to terminal junior residual when the source bucket is already impaired. It does not recreate senior provider backing, and it preserves the existing post-snapshot residual accounting.

## Public reproduction

The regression uses only public program instructions:

1. create two ordinary long/short pairs
2. move the authorized mark and leave one short's negative K/F delta unsettled
3. create a genuine source lien through a source-backed risk increase
4. resolve the market and close a lien-free winner, impairing the shared bucket
5. close the untouched loser and require bounded progress

Against the exact parent engine pin `b9efec69740bbc5002035d64e8f98b7906f362ce`, three close attempts fail with custom error `0x15`, with market and portfolio bytes unchanged after rollback. With the fix, the close succeeds below the 1,000,000 CU ceiling.

## Validation

- engine `cargo test --features fuzz`: 171 passed
- engine Kani `proof_v16_resolved_impaired_capital_backed_loss_routes_to_residual`: 5,990 checks, 2/2 covers passed
- wrapper SBF build with tools `v1.52`: passed
- wrapper unit tests: 5 passed
- wrapper public LiteSVM/CU tests: 579 passed

This PR intentionally does not duplicate the engine's arithmetic/state-transition proof in wrapper Kani; the wrapper-specific obligation is the public instruction route and rollback/progress regression.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=b9efec69740bbc5002035d64e8f98b7906f362ce#b9efec69740bbc5002035d64e8f98b7906f362ce"
+source = "git+https://github.com/aeyakovenko/percolator?rev=0401193a8f277720529b7359e97bf1f6b659c475#0401193a8f277720529b7359e97bf1f6b659c475"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-030 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #303, opened 2026-07-23
[HARDENING LoF] Bind trades to portfolio incarnations

aeyakovenko:
## Classification

REAL LoF. Trade payloads authenticated current owner keys and reusable portfolio addresses, but not the market-assigned portfolio incarnations. A counterparty can retain a victim-signed trade for empty incarnation A, wait for A to close and the same address to be funded as B, then execute A's trade against B.

## Public RED

The clean-room LiteSVM exploit on parent `15f6d588305f3ec66fd63ed841cde04df818dc04` uses only public System, SPL Token, and Percolator instructions:

- A signs a normal bilateral trade that is not submitted
- A closes; the same address is publicly funded and initialized as B with a larger monotonic ID
- B deposits 100,000 collateral atoms
- the retained A transaction opens B as a thin short
- authenticated marks and permissionless cranks follow the normal production-risk path
- an unprivileged relayer liquidates B and withdraws the reward as SPL collateral

```text
incarnation-A trade opened replacement B and paid the relayer a 524-atom withdrawable liquidation reward
```

Parent SBF SHA256: `05bacd937698db9a8acf3d8b77990b3229bc166712ec716a46c9ea9d0a1e729f`.

## Fix

`TradeNoCpi`, `TradeCpi`, `BatchTradeNoCpi`, and `BatchTradeCpi` now carry both participants' market-assigned portfolio IDs. Both IDs are checked with the existing incarnation binder before engine mutation or matcher CPI.

The regression matrix independently supplies stale A and stale B IDs to all four entry points and proves `EngineProvenanceMismatch`, byte-identical market/portfolio rollback, and unchanged matcher context.

No signer, authority, PDA, custody path, token movement, recipient, or administrator surface is added. Old unbound trade payloads are rejected.

## Verification

- real-SBF LiteSVM: 569 passed
- host unit tests: 6 passed
- Kani wire proofs: 4 verified, 0 failures
- exploit regression: GREEN with zero reward and exact rollback
- fixed SBF SHA256: `2d58e521bc39d7a9ed7708223927b73277408762a02e44359a23eb2112884ec0`

This draft is intentionally stacked on percolator-prog PR #301 so its diff contains only the trade-incarnation fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,12 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
   a different ID, so a stale snapshot cannot authenticate the replacement account. IDs are local to
   one market, so `portfolio_id` alone is not an identity. **✅**
   (`v16_portfolio_incarnation_id_separates_close_and_reuse`.)
+- **Signed trades bind both portfolio incarnations.** `TradeNoCpi`, `TradeCpi`,
+  `BatchTradeNoCpi`, and `BatchTradeCpi` carry the market-assigned `portfolio_id` for both account A
+  and account B. Both IDs are checked before engine mutation or matcher CPI, so an unlanded trade
+  signed for a closed account cannot execute against a later portfolio at the same address. **✅**
+  (`v16_attack_presigned_trade_rejects_reinitialized_portfolio`,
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #304, opened 2026-07-23
[HARDENING LoF] Bind matcher grants to portfolio incarnations

aeyakovenko:
## Classification

REAL LoF reachable through public instructions on a normally initialized live market.

## Root cause

`SetMatcherConfig` signed only the enabled byte. Its account tuple was bound to the LP account address and owner, but not to the market-owned monotonic portfolio ID. After a portfolio was closed and the same account address was initialized again for the same owner, a retained grant from incarnation A could enable the matcher on funded incarnation B.

The trade-entrypoint ID checks from #303 do not stop this: after replaying the stale grant, an attacker can submit a fresh `TradeCpi` naming B's public current ID.

## Reproduction

The LiteSVM regression uses only public protocol and matcher instructions:

1. an LP owner signs a matcher grant for empty portfolio incarnation A;
2. A disables the grant, closes, and the same account address is publicly reinitialized as funded incarnation B;
3. an unprivileged relayer replays A's retained grant;
4. the relayer submits a fresh matcher trade using B's current portfolio ID;
5. authenticated mark updates and permissionless cranks make B liquidatable;
6. the relayer liquidates B and withdraws a 524-atom victim-funded reward as SPL collateral.

On the exact parent, the test fails with:

`incarnation-A matcher grant authorized a fresh trade against B and paid the attacker a 524-atom withdrawable liquidation reward`

## Fix

- Add `portfolio_id: u64` to the tag 68 payload.
- Validate it against the portfolio's existing monotonic ID before matcher-config reallocation or mutation.
- Reject stale grants with `EngineProvenanceMismatch`.
- Reject the old unbound ABI.
- Update clients/tests and document the incarnation-bound grant.
- Add Kani coverage for field preservation, trailing bytes, and truncation.

This adds no signer, PDA, authority, admin, token-transfer, or custody surface.

## Validation

- Fixed SBF SHA-256: `1d0eb05d18c47eb1c35724ff9305408bc632b4d178decbcf28ab96f5f7364b99`
- 570/570 serial LiteSVM tests passed.
- 6/6 unit tests passed.
- 3/3 relevant Kani harnesses verified with zero failures:
  - `kani_v16_set_matcher_config_decode_preserves_portfolio_id_and_enabled`
  - `kani_v16_trade_and_crank_payloads_reject_trailing_byte`
  - `kani_v16_every_active_payload_rejects_one_byte_truncation`
- Stale replay asserts exact `Custom(16)` and byte-identical market, portfolio, and matcher-context rollback.

Stacked on #303 because the exploit regression intentionally exercises the current trade-incarnation ABI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,8 +274,8 @@ Positions and PnL use native `i128`/`u128` (`POS_SCALE = 1_000_000`, `ADL_ONE = 
 ### Two trade paths
 - **TradeNoCpi**: no external matcher; used for baseline integration, local testing, and deterministic program-test scenarios.
 - **TradeCpi**: production matcher path; the LP owner configures a matcher program/context once on
-  the LP portfolio with `SetMatcherConfig` (tag 68). Fills then run without the LP owner
-  signing each transaction. Direct LP-signed bilateral trading is `TradeNoCpi`.
+  the LP portfolio with an incarnation-bound `SetMatcherConfig` (tag 68). Fills then run without
+  the LP owner signing each transaction. Direct LP-signed bilateral trading is `TradeNoCpi`.
 
 ### MatchingEngine trait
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #305, opened 2026-07-23
[HARDENING LoF] Bind deposits to portfolio incarnations

aeyakovenko:
## Root cause

Deposit tag 3 committed only to an amount. A victim-signed deposit for an empty portfolio incarnation could therefore be retained, the account closed and reinitialized at the same address, and the old transaction replayed against the leveraged replacement portfolio before its recent blockhash expired.

Because deposits become replacement-portfolio collateral, the opposing trader could consume and withdraw those wallet funds. The exact-parent LiteSVM exploit extracted 100,000 SPL atoms beyond the replacement portfolio own 150,000-atom capital.

## Fix

- Add the market-assigned monotonic portfolio ID to the Deposit payload.
- Validate that ID before portfolio realloc, market mutation, or token CPI.
- Reject the old unbound payload shape and malformed trailing/truncated forms.
- Update the public ABI documentation and all deposit fixtures.

This adds no signer, PDA, authority, admin, ledger, or custody destination.

## TDD evidence

Exact parent 5bb82aa:

- v16_attack_presigned_deposit_rejects_reinitialized_portfolio reaches a normal live-market terminal payout and reports 100,000 victim atoms extracted as SPL collateral.

Fixed 89664c0:

- The identical retained transaction rejects with Custom(16) before mutation.
- Market, replacement portfolio, source token account, and canonical vault remain byte-identical.
- Source retains all 100,000 atoms.

## Validation

- 571/571 serial LiteSVM tests
- 6/6 unit tests
- 3/3 focused Kani harnesses: field preservation, trailing-byte rejection, and truncation rejection
- cargo check --tests
- cargo fmt --all -- --check
- git diff --check
- Fixed SBF SHA256: d709e3781da82185799342a26b7268a76776ec75bea132316af39bb49258d37b

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,7 @@ This section describes intent and operational ordering, not argument-by-argument
 - **InitPortfolio** (tag 1)
   - initializes a program-owned portfolio account and binds `owner = signer`
 - **Deposit** (tag 3)
+  - binds the signed payload to the current program-assigned portfolio incarnation ID before moving tokens
   - transfers collateral into vault; credits engine balance for that account
 - **Withdraw** (tag 4)
   - binds the signed payload to the current program-assigned portfolio incarnation ID
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2476,6 +2476,7 @@ pub mod ix {
         },
         InitPortfolio,
         Deposit {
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #306, opened 2026-07-23
[BLOCKER DoS] Prove cross-domain source rounding progress

aeyakovenko:
## Impact
A live two-asset portfolio can become permanently unable to process a fair authenticated mark reversal. Two source domains each expose fractional support; the old engine aggregates those fractions into one extra atom, but no individual domain can consume it. Permissionless crank, owner rebalance, and an honest later-slot retry all return engine `LockActive` with SVM rollback.

## Public RED
The first commit is an exact-parent LiteSVM reproduction on wrapper `da64be63` / engine `143e68c4`. It uses only public configure-mark, deposit, `TradeNoCpi`, push-mark, permissionless-crank, and owner-signed rebalance instructions. It does not mutate protocol state or assume a dishonest oracle.

## Fix pin
The second commit pins engine PR [#154](https://github.com/aeyakovenko/percolator/pull/154), which floors and caps realizable support per domain using the same helper as consumption. The test is unchanged between RED and GREEN.

## Verification
- freshly rebuilt normal, authenticated, and hostile matcher SBF fixtures
- focused exact public regression GREEN
- `cargo test --all-targets --quiet`: 6 unit + 566 LiteSVM/CU tests, 0 failures
- engine: 159 fuzz-enabled runtime tests; focused Kani 458 checks, 2/2 covers, 0 failures

Exact wrapper parent: `da64be639168b496833dd4c701607a94fcb06b6c`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=ad1902d24f67dc1f34e73729e785059c54c74348#ad1902d24f67dc1f34e73729e785059c54c74348"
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #307, opened 2026-07-23
[HARDENING LoF] Bind deposits to market generations

aeyakovenko:
## Classification

**REAL LoF.** An unprivileged holder of a generation-A signed `Deposit` can broadcast it after the same market address is closed and recreated, forcing the signer’s still-funded SPL source into a leveraged generation-B portfolio. The public LiteSVM trace realizes the resulting loss and pays the opposing trader exactly 100,000 collateral atoms beyond its principal and replacement B’s own 150,000-atom capital.

This PR is intentionally stacked on #231 because that branch supplies the persistent market-generation counter needed to distinguish full `CloseSlab` + `InitMarket` cycles.

## Root cause

Tag 3 authenticated only `amount`. It did not authenticate the market incarnation whose portfolio risk would receive the collateral. A retained signature therefore remained valid when the same market and portfolio addresses were recreated.

## Fix

- Add asset 0’s `market_id` to the canonical `Deposit` wire payload.
- Reject a stale generation with `AssetGenerationMismatch` from a read-only preflight before portfolio resize, state mutation, or token CPI.
- Bind normal test/client deposits to the current live generation.
- Document the generation requirement.
- Add Kani coverage for field preservation, trailing bytes, and one-byte truncation.

No signer, authority, PDA, custody destination, subledger, or admin surface changes.

## TDD evidence

The first commit, `36a8e3f`, is the portable RED regression. On the exact #231 parent (`4c5753f`) it completes the public lifecycle and fails with:

```
generation-A deposit drained the signer's token source and paid 100000 atoms beyond replacement B's own capital to the opposing trader
```

On the fixed SBF it passes, requiring the exact generation-mismatch error and byte-identical market, replacement portfolio, source-token, and vault accounts.

## Validation

- Fixed SBF SHA-256: `8db100d9eac728778a1c9bb40e30bc6a159ae444c6430fe326d8fa5454403b18`
- Focused real-SBF LiteSVM regression: 1 passed
- Serial real-SBF LiteSVM suite: 569 passed
- Library tests: 5 passed
- `cargo check --all-targets`
- Kani field-preservation proof: successful
- Kani trailing-byte proof: successful
- Kani truncation proof: successful
- `cargo fmt --all`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,8 @@ This section describes intent and operational ordering, not argument-by-argument
 - **InitPortfolio** (tag 1)
   - initializes a program-owned portfolio account and binds `owner = signer`
 - **Deposit** (tag 3)
+  - carries asset 0's current `market_id`; a stale market generation rejects before portfolio resize,
+    state mutation, or token movement
   - transfers collateral into vault; credits engine balance for that account
 - **Withdraw** (tag 4)
   - performs oracle-read + engine checks; withdraws from vault via PDA signer; debits engine
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2523,6 +2523,7 @@ pub mod ix {
         },
         InitPortfolio,
         Deposit {
+            market_id: u64,
             amount: u128,
         },
         Withdraw {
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #155, opened 2026-06-20
Reject stale TradeCpi matcher context replay

aeyakovenko:
## Summary
- Bind single `TradeCpi` matcher request ids to the pre-call matcher context bytes, market/accounts, matcher program/context, asset, oracle price, and requested size.
- Prevent a hostile matcher from returning `Ok(())` without writing fresh context bytes and replaying a stale valid return in a later same-transaction `TradeCpi`.
- Add fixture source support for a no-write replay mode and a liveness positive test showing two identical honest same-transaction fills still execute when the matcher writes fresh returns.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `(cd tests/fixtures/hostile_matcher && cargo build-sbf)`
- `cargo build-sbf --no-default-features`
- `cargo test --test v16_cu v16_attack_hostile_matcher_no_write_cannot_replay_stale_single_context -- --nocapture`
- `cargo test --test v16_cu v16_attack_honest_matcher_can_repeat_identical_single_fills_in_one_tx -- --nocapture`
- `cargo test --test v16_cu -- --nocapture` (541/541)

PR #308, opened 2026-07-23
[BLOCKER DoS] Settle B by collateral-atom budget

aeyakovenko:
## Finding

The engine treated the collateral-atom B-settlement endpoint budget as a raw B-index delta cap after already converting the configured atom cap through `loss_weight`.

The LiteSVM regression reaches the state exclusively through public instructions: market initialization, three portfolio deposits, bounded auth-mark updates, trades, rebalance, permissionless cranks, resolution, and resolved closes. It creates:

- B target: `107486458947473684210526315`
- survivor loss weight: `19000000`
- total survivor loss: about `2.042e12` atoms
- configured atom budget: `1e16` atoms

The old engine nevertheless advanced only `1e16` B-index units per successful close, requiring more than 10.7 billion transactions to release the survivor. Every public continuation uses the same chunk planner, so an honest cranker has no bounded alternative.

## TDD

- Commit 1 pins the exact prerequisite engine parent and fails at the bounded-settlement assertion.
- Commit 2 pins engine #155 and passes the same public lifecycle.
- The older injected-state B test now checks the actual contract: charge no more than the atom cap while advancing the full index gap when that loss fits.

## Verification

- `cargo test --all-targets --quiet`
  - 6 unit tests passed
  - 567 LiteSVM/CU tests passed

## Dependencies

- Wrapper #216 / engine #115 remove an earlier independent ADL terminal-close blocker needed to expose this second-order path.
- Engine fix: aeyakovenko/percolator#155

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=f2019ef76d06b28f40136f4d099e644e3b131f10#f2019ef76d06b28f40136f4d099e644e3b131f10"
+source = "git+https://github.com/aeyakovenko/percolator?rev=214e3f823dfb6a9068821385b005214af871c055#214e3f823dfb6a9068821385b005214af871c055"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_b_budget_lock_prerequisite_rejects_post_adl_basis_reissue`. The public PR308 prefix requires ordinary trades after quantity ADL. The fixed engine rejects the first such basis reissue with exact rollback, while owner RebalanceReduce still makes bounded progress, so the downstream B-budget state cannot commit. The executable disposition remains in PR #135; no separate production fix is required.

PR #309, opened 2026-07-23
[HARDENING LoF] Bind portfolio close to its incarnation

aeyakovenko:
## Classification

**REAL LoF** reachable through the public `ClosePortfolio` interface on a normally initialized live market.

A portfolio address can be closed and later reinitialized with a new monotonic portfolio ID. A close transaction signed for the old incarnation previously carried no incarnation witness, so it remained valid against the replacement account. The exact-parent real-SBF LiteSVM RED replay swept all **1,000,000,000 lamports** owned by replacement incarnation 2 into the market slab.

This is the upstream fix for the existing Meta DAO RED coverage in aeyakovenko/percolator-meta#351.

## TDD history

- Exact parent: `da64be6`
- RED commit: `18501cf`
- Fix commit: `184d27d`

## Fix

- Carry the expected portfolio ID in `ClosePortfolio`.
- Compare it with the account's current ID before resize, deregistration, mutation, or lamport movement.
- Preserve tag-only compatibility only for recognized pre-ID layouts whose stored ID is zero.
- Reject partial and trailing payloads canonically.
- Document the wire/lifecycle rule and add decoder proofs.

This adds no signer, authority, custody, admin, or fund-moving surface.

## Verification

- Real-SBF serial LiteSVM: **567 passed, 0 failed**
- Library tests: **6 passed, 0 failed**
- Focused stale-close, legacy-layout, and legacy-realloc tests: passed
- Kani:
  - close decode preserves the incarnation and legacy zero
  - resolved recovery payloads reject trailing bytes
  - active payloads reject one-byte truncation
- `cargo check --all-targets`: passed
- `rustfmt --check`: passed
- `git diff --check`: passed
- Rebuilt SBF SHA-256: `9ae56ff9522d0ceba5b3643b2c879e5b011beab877adfff4de64f7f87bed6faf`

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,7 +404,9 @@ This section describes intent and operational ordering, not argument-by-argument
 - **Withdraw** (tag 4)
   - performs oracle-read + engine checks; withdraws from vault via PDA signer; debits engine
 - **ClosePortfolio** (tag 8)
+  - carries the current portfolio ID; stale incarnations reject before deregistration or lamport movement
   - closes a flat, empty portfolio account and sweeps its rent to the market slab
+  - legacy tag-only closes are accepted only for pre-ID portfolios whose stored ID is zero
 - **CloseResolved** (tag 30)
   - resolved-market payout/finalization path for a supplied portfolio; pays only the stored owner token account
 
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #311, opened 2026-07-23
[BLOCKER LoF] Bind market resolution to asset generations

aeyakovenko:
## Classification

REAL loss of funds reachable through the public interface.

An unprivileged submitter can retain an administrator-signed `ResolveMarket` transaction from empty asset-0 generation A, wait for an ordinary public shutdown/restart into generation B, and submit the stale transaction after replacement users fund and trade. Resolution crystallizes temporary generation-B PnL and makes the opposing portfolio's gain withdrawable.

## Reproduction

- Exact parent: `4c5753f0bfe91cce2116ce2434fdce34537423de`
- RED commit: `6d60b89`
- Fix commit: `9f408ac`
- No direct account-state injection is used.
- Public lifecycle: empty shutdown at slot 2, restart at slot 3, replacement deposits/trade, authenticated mark move, retained resolve submission.
- Exact RED loss: victim deposit `1,000,000` pays out `900,000`; opposing deposit `1,000,000` pays out `1,100,000`. The stale signature transfers `100,000` SPL atoms of replacement victim capital.

## Fix

`ResolveMarket { market_id }` now carries the intended asset-0 generation. The processor compares it with the active generation before reading time or mutating resolution state. Legacy tag-only, truncated, mismatched-generation, and trailing-byte payloads fail closed. A current-generation resolve remains available.

This adds no authority, custody, withdrawal, or administrator fund-moving surface.

## Verification

- Focused exact-parent RED fails with the 100,000-atom transfer.
- Focused fixed real-SBF LiteSVM regression passes.
- Full serial real-SBF LiteSVM: `569 passed; 0 failed` in 227.89s.
- Resolution-filtered LiteSVM: `66` paths exercised; `65` initially passed and one fixture-missing case passed after building the hostile matcher.
- Library tests: `5 passed; 0 failed`.
- `cargo check --all-targets` passes.
- Focused Kani wire proof: `0 of 845 failed`.
- `rustfmt --check` and `git diff --check` pass.
- Fixed SBF SHA-256: `2900cac88a42411594e60e93934de584f80aeb585967953370c3a058e788a598`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -855,8 +855,10 @@ These are governance powers, not bugs:
 3. `UpdateAssetAuthority { asset_index = 0, market_id, kind = ASSET_AUTH_ORACLE }` (while marketauth holds asset-0's `asset_admin`)
    - choose who can push asset-0 AuthMark/EwmaMark updates.
    - impact: authority mark input control/censorship surface.
-4. `ResolveMarket`
+4. `ResolveMarket { market_id }`
    - transition market to resolved mode using stored authority price.
+   - the signed terminal action is bound to asset 0's current generation; retained resolves from a
+     shutdown/restart generation reject before changing market state.
    - impact: trading/deposits/new accounts are halted; market enters wind-down.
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #313, opened 2026-07-23
[PRIVILEGED LoF] Require LP consent for CPI base fees

aeyakovenko:
## Root cause

PR #224 correctly removes the taker's unsigned caller fee from CPI fills, but the replacement still charges the live market `trade_fee_base_bps`. The asset-0 insurance authority can raise that mutable base after an LP authorizes a matcher. A taker controlled by that authority can then execute a retained single or batch CPI open/close, charge the unsigned LP, withdraw the resulting insurance, and realize the LP's loss.

The exact-parent RED commit `834a2a15` proves both public routes with LiteSVM: two 1,000,000-atom deposits, a post-authorization increase to 500 bps, paired open/close fills, a 10,000-atom independent LP loss, and authority withdrawal of the 20,000-atom insurance charge.

## Fix

- Extend the owner-signed `SetMatcherConfig` payload with a cap for the authority-mutable base fee.
- Store the cap in previously reserved bytes of the existing 104-byte zero-copy portfolio matcher tail; account size is unchanged and compile-time asserted.
- Decode legacy two-byte matcher authorizations as cap zero, so they fail closed at nonzero base fees until the LP reauthorizes.
- Reject single and batch CPI before matcher invocation when the live base exceeds the retained LP cap.
- Keep the matcher ABI unchanged. Deterministic hybrid/EWMA externality fees remain outside this mutable-base cap.
- Bound caps to 10,000 bps and preserve rollback on invalid updates.

The fix commit is `2c33c88c`. The regression exercises the exact 499/500 bps reject/accept boundary for both CPI routes and verifies a freshly signed consent still trades normally.

## Validation

- `cargo build-sbf --tools-version v1.52` (program, auth matcher, hostile matcher)
- `cargo test --all-targets -- --test-threads=1`: 5 unit + 565 LiteSVM/CU tests passed
- `cargo kani --tests --harness kani_v16_set_matcher_config_decode_preserves_fee_consent`: 1/1 harness, 1,494 checks passed
- `cargo check --all-targets`
- `git diff --check`

This PR is intentionally stacked on #224 because the exploit exists after #224's caller-fee fix.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -654,8 +654,15 @@ pub mod state {
         pub matcher_program: [u8; 32],
         pub matcher_context: [u8; 32],
         pub matcher_delegate: [u8; 32],
-        pub enabled: u64,
+        pub enabled: u8,
+        pub reserved0: u8,
+        /// LP-signed cap for the authority-mutable base fee component.
+        pub trade_fee_cap_bps: u16,
+        pub reserved1: [u8; 4],
     }
+
+    const _: [(); PORTFOLIO_MATCHER_CONFIG_LEN] =
+        [(); core::mem::size_of::<PortfolioMatcherConfigV16>()];
 
     pub type AssetOracleStorageV16 = [u8; ASSET_ORACLE_WRAPPER_LEN];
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #314, opened 2026-07-23
[BLOCKER LoF] Bind permissionless activation fees to creator consent

aeyakovenko:
## What changed

- add a caller-signed `max_init_fee` to `UpdateAssetLifecycle`
- reject permissionless activation when the live fee exceeds that cap, both during preflight and after account revalidation
- keep retained transactions usable when the live fee falls below their cap, charging only the lower live fee
- extend lifecycle codec proofs and all wrapper call sites for the new tag-40 field

## Public attack and impact

Permissionless activation authorizes an SPL-token debit but the vulnerable instruction did not serialize the fee or a caller limit. It read mutable `permissionless_market_init_fee` only when the transaction landed.

The exact-parent public LiteSVM reproduction uses only honest signatures and normal instructions:

1. the market admin signs a 1,000-unit fee policy but does not land it;
2. the admin lands a newer correction setting the fee to 1;
3. an independent creator observes that state and signs activation;
4. an unprivileged relayer lands the retained older 1,000-unit policy, then the unchanged activation.

On `origin/main` `36fa587e`, activation succeeds and drains the creator source from 1,000 to 0, crediting 1,000 to canonical insurance/vault. The relayer does not need an authority key, oracle control, state injection, or victim cooperation with the high fee. This is an independent-victim LoF caused by replay/reordering of valid public transactions.

The RED test is isolated in `abb75c6c`; the fix is `bd76804f`. Tag 40 gains a required 16-byte field, so legacy payloads reject rather than silently authorize an unbounded debit.

## Verification

- `cargo build-sbf --tools-version v1.52 --sbf-out-dir target/deploy`
- exact-parent RED: stale schedule lands, activation succeeds, victim loses all 1,000
- fixed-head GREEN: stale schedule is rejected and victim retains all 1,000
- lower-live-fee positive boundary remains accepted
- `cargo test --all-targets -- --test-threads=1`: 6 unit + 567 public LiteSVM/CU tests passed
- `cargo kani --tests --harness kani_v16_asset_lifecycle_decode_preserves_wire_fields --unwind 34`: 0/1512 failed
- `cargo check --all-targets`, `cargo fmt --all -- --check`, and `git diff --check` passed

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2635,6 +2635,7 @@ pub mod ix {
             asset_index: u16,
             now_slot: u64,
             initial_price: u64,
+            max_init_fee: u128,
             insurance_authority: [u8; 32],
             insurance_operator: [u8; 32],
             backing_bucket_authority: [u8; 32],
@@ -2908,6 +2909,7 @@ pub mod ix {
                     asset_index: read_u16(&mut rest)?,
                     now_slot: read_u64(&mut rest)?,
                     initial_price: read_u64(&mut rest)?,
+                    max_init_fee: read_u128(&mut rest)?,
                     insurance_authority: read_bytes32(&mut rest)?,
                     insurance_operator: read_bytes32(&mut rest)?,
                     backing_bucket_authority: read_bytes32(&mut rest)?,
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #315, opened 2026-07-23
[BLOCKER LoF] Bind asset shutdown to market generations

aeyakovenko:
## Impact

A retained `UpdateAssetLifecycle(SHUTDOWN)` authorization from empty asset generation A remains valid after the same slot is normally shut down and restarted as generation B. An opposing trader can relay it after generation-B users open matched positions, freeze settlement at a favorable committed mark, permissionlessly force-close, and withdraw capital that could not be realized unilaterally in the live market.

The exact-parent LiteSVM RED path funds two independent generation-B portfolios with 1,000,000 atoms each. After stale shutdown and public wind-down, the victim withdraws 900,000 while the opposing trader withdraws 1,100,000. This is a public, extractable 100,000-atom loss against a normally initialized live asset.

## Root cause

Legacy tag 40 authenticated `marketauth` or `asset_admin` but carried no asset-generation witness, so a signed terminal capability survived shutdown and restart of the same slot.

## Fix

- Add `ShutdownAsset { asset_index, market_id, now_slot }` as tag 70.
- Require `market_id` to match the current engine asset generation before mutation.
- Reject generation-free shutdown through legacy `UpdateAssetLifecycle`.
- Preserve the existing authority check, trader exit window, and force-close path.
- Add wire round-trip, trailing-byte, and truncation proofs.
- Add no authority or fund-moving surface.

## TDD evidence

- Exact parent: `32f9b381`
- Exact-parent SBF SHA-256: `8c267d12effff796ff560a44ecd38fee0304b4fde3a5fdb59090d9b79888369f`
- RED commit: `5a529176`
- Fix commit: `4973cc41`
- Fixed SBF SHA-256: `d613a1507def1d21f66da0c70fdbc4b97eeeb5669786ed64789453a575e96331`

The adaptive regression fails on the exact parent only after proving the 900,000 / 1,100,000 withdrawals. It passes on the fix by proving stale rejection is byte-atomic and a shutdown signed for generation B remains reachable.

## Validation

- `cargo test --test v16_cu --quiet`: 579 passed
- `git diff --check`: passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,8 +163,9 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
   `WrapperConfigV16.marketauth` key (it replaced the former separate `admin` / `asset_authority` /
   `base_unit_authority`). `marketauth` is the only key that can: **create market 0** (`InitMarket`),
   **create/retire assets 1..N and set the permissionless-create-fee policy**, **safely force-shutdown
-  any asset including asset 0** (`ASSET_ACTION_SHUTDOWN` → RECOVERY with the `force_close_delay_slots`
-  exit window so traders can exit), **resolve/close the market** (`ResolveMarket { market_id }`/`CloseSlab`),
+  any asset including asset 0** (`ShutdownAsset { asset_index, market_id, now_slot }` → RECOVERY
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #316, opened 2026-07-23
[HARDENING LoF] Bind rebalance consent to position episodes

aeyakovenko:
## What changed

- add a wrapper-owned, checked `position_epoch` to each portfolio
- bind owner-signed `RebalanceReduce` payloads to both `portfolio_id` and `position_epoch`
- advance the epoch after successful exposure-changing single/batch trades, rebalance, liquidation, recovery force-close, and recovery-leg forfeit
- leave refresh-only cranks unchanged so permissionless maintenance cannot invalidate valid consent
- reject both pre-incarnation and portfolio-only legacy tag-44 payloads

## Root cause and impact

PR #292 binds reduction consent to a portfolio incarnation, but a still-live portfolio can flatten and reopen the same asset without changing its market, account address, portfolio ID, or asset identity. A retained owner-signed reduction for the old position therefore applied to the replacement position.

The exact-parent public LiteSVM reproduction uses only normal instructions and no state injection:

1. victim opens a 2,000-unit long and signs a 1,000-unit reduction
2. victim publicly closes that position and later opens a replacement 1,000-unit long in the same portfolio
3. the authenticated mark moves from 100 to 50
4. the retained old reduction lands and closes the replacement at the adverse mark
5. settlement pays the victim 950,000 and the counterparty 1,050,000

The same RED reproduced on #292 head `791ff0c9`, so this is distinct from account-incarnation replay.

## Validation

- exact-parent public LiteSVM RED: victim 950,000 / attacker 1,050,000
- fixed wrapper and all matcher SBFs rebuilt from clean worktrees
- `cargo test --all-targets --no-default-features`: 7 unit + 567 LiteSVM/CU tests passed
- all four public trade APIs explicitly advance both portfolio epochs
- refresh-only crank preserves epoch; successful partial liquidation advances it exactly once
- Kani tag-44 round trip: 1,507 checks, 0 failures
- Kani legacy payload rejection: 829 checks, 0 failures
- `git diff --check`

This PR intentionally targets #292's branch so the account-incarnation and position-episode bindings remain reviewable as two separate roots. Engine pin remains `143e68c4917ed0400a27b952f036a5677047cd84`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,7 +482,12 @@ This section describes intent and operational ordering, not argument-by-argument
   - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget
 - **RebalanceReduce** (tag 44)
   - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector;
-    the payload binds consent to the program-assigned `portfolio_id`
+    the payload binds consent to both the program-assigned `portfolio_id` and the portfolio's
+    `position_epoch`
+  - `position_epoch` advances after every successful exposure-changing trade, rebalance,
+    liquidation, recovery force-close, or recovery-leg forfeit; refresh-only cranks do not advance
+    it, so stale consent cannot target a replacement position and permissionless maintenance cannot
+    invalidate current consent
 - **ClaimResolvedPayoutTopup** (tag 46)
   - permissionless resolved-payout top-up claim; pays only the stored owner receipt token account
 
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-004 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #317, opened 2026-07-23
[BLOCKER LoF] Bind fee redirects to market generations

aeyakovenko:
## Summary
- bind `UpdateFeeRedirectPolicy` (tag 58) to the persistent market generation
- reject a stale signed policy before any replacement-market write
- update ABI proofs/call sites and add a public close/reinitialize LiteSVM regression

## Public LoF
A generation-A zero-redirect policy can be retained through public `ResolveMarket` / `CloseSlab`, then replayed after same-address generation B explicitly routes 100% of fees to protected market-0 insurance. A permissionless asset-1 creator can withdraw both redirected trade fees and finish 1,000 atoms ahead, exactly matching an independent victim's 1,000-atom loss.

The fix adds no authority or fund-moving surface. It only includes `market_id` in the signed payload and compares it with the persistent generation before writing policy state.

## TDD evidence
- exact pinned parent SBF `e4948402cbd85b58d0de6ad57550da9ab20aed8ec5d494540cda26fb1d46f2ab`: RED, attacker withdraws 2,000 atoms and protected insurance retains only the one-atom init fee
- fixed SBF `16dcb7c1504a3d0d65074fcee6675270f08b15ab730706b509c72a4e661a8bc8`: GREEN, stale replay rejects with `AssetGenerationMismatch`, market stays byte-identical, and 2,001 atoms remain in market-0 insurance
- `cargo test --all-targets`: 570 LiteSVM tests and 5 unit tests pass after building the two matcher fixtures
- `cargo fmt --check`
- `git diff --check`

Stacked on the trade-fee generation binding because both depend on the persistent market-generation prerequisite.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2616,6 +2616,7 @@ pub mod ix {
             trade_fee_base_bps: u64,
         },
         UpdateFeeRedirectPolicy {
+            market_id: u64,
             redirect_bps: u16,
         },
         UpdateMarketInitFeePolicy {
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #318, opened 2026-07-23
[BLOCKER LoF] Bind backing fees to asset generations

aeyakovenko:
## What changed

- Add the current asset `market_id` to `UpdateBackingFeePolicy` (tag 51).
- Reject a retained policy when its signed generation does not match the active asset generation.
- Reject the legacy generation-free wire payload and retain exact decode/trailing-byte proofs.
- Add an adaptive public LiteSVM regression that is RED on the exact parent binary and GREEN on this fix.

## Impact and root cause

Tag 51 was bound only to a reusable backing domain. After asset retirement and permissionless slot reuse, an unprivileged relayer could submit a policy signed for generation A into generation B. The retained policy added a 75-atom fee to an independent victim's already-signed generation-B trade and credited all 75 atoms to the attacker's replacement backing-provider ledger. The attacker withdrew the full amount, ending 75 atoms ahead while the victim ended 75 atoms behind.

This patch adds no admin authority and no fund-moving path. It only validates the monotonic asset generation before mutating the policy.

## Validation

- Exact-parent RED SBF SHA-256: `16dcb7c1504a3d0d65074fcee6675270f08b15ab730706b509c72a4e661a8bc8`
- Fixed GREEN SBF SHA-256: `a49a00ab3fa7589872435b50d93f64d3827e3e219db674a306ea7eb926fb1388`
- Full suite: 571 LiteSVM tests passed
- Unit suite: 5 tests passed
- `cargo fmt --check`
- `git diff --check`

Stacked on #317 so this PR contains only the backing-fee generation binding.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2608,6 +2608,7 @@ pub mod ix {
         },
         UpdateBackingFeePolicy {
             domain: u16,
+            market_id: u64,
             fee_bps: u16,
             insurance_share_bps: u16,
         },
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #320, opened 2026-07-23
[BLOCKER LoF] Bind insurance top-ups to asset generations

aeyakovenko:
## What changed

- Add the active asset `market_id` to `TopUpInsuranceDomain` (tag 56).
- Reject a retained top-up when its signed generation does not match the active asset generation.
- Reject the legacy generation-free wire payload and preserve decode proofs.
- Add an adaptive public LiteSVM regression that is RED on the exact parent binary and GREEN on this fix.

## Impact and root cause

Tag 56 was bound only to a reusable domain. An insurance authority could sign a generation-A top-up without submitting it. After public retirement, an attacker could permissionlessly reuse the slot, name the old signer as `insurance_authority`, name itself as the distinct `insurance_operator`, relay the retained transfer, and immediately withdraw all credited insurance.

The regression demonstrates a direct 10,000-atom transfer from an independent depositor to the replacement operator. It uses only public instructions and a retained signed transaction: no positions, oracle move, admin signature, ledger mutation, or injected program state.

This patch adds no authority and no new fund-moving path. It validates the monotonic asset generation before accounting or SPL transfer.

## Validation

- Exact-parent RED SBF SHA-256: `a49a00ab3fa7589872435b50d93f64d3827e3e219db674a306ea7eb926fb1388`
- Fixed GREEN SBF SHA-256: `a04f9ef2644acef23a7714f3fdd14dfbdc5ac1da1aada1086c09cb7cb9e4d92e`
- Full suite: 572 LiteSVM tests passed
- Unit suite: 5 tests passed
- `cargo fmt --check`
- `git diff --check`

Stacked on #318 so this PR contains only the insurance top-up generation binding.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2565,6 +2565,7 @@ pub mod ix {
         },
         TopUpInsuranceDomain {
             domain: u16,
+            market_id: u64,
             amount: u128,
         },
         CloseSlab,
@@ -2842,6 +2843,7 @@ pub mod ix {
                 },
                 56 => Self::TopUpInsuranceDomain {
                     domain: read_u16(&mut rest)?,
+                    market_id: read_u64(&mut rest)?,
                     amount: read_u128(&mut rest)?,
                 },
                 13 => Self::CloseSlab,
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #321, opened 2026-07-23
[BLOCKER LoF] Bind backing top-ups to asset generations

aeyakovenko:
## Impact

A signed `TopUpBackingBucket` capability was bound only to a domain index. After that asset generation retired, a permissionless replacement creator could assign the old backing authority to the reused slot and relay a retained generation-A top-up into generation B.

The LiteSVM exploit uses only public post-reuse actions: it opens a real self-matched position, advances the replacement oracle, lets the signed backing expire, invokes permissionless resolution, and orders terminal closes so the replacement-market winner receives all 150 atoms of the independent provider's backing.

## Root cause

Tag 24 did not carry the asset's monotonic `market_id`, so authorization survived slot retirement and reuse.

## Fix

- Add `market_id` to the tag-24 wire format.
- Reject a generation mismatch before accounting changes or the SPL transfer.
- Update all callers and the Kani wire-format proof.
- Keep a current-generation positive control in the adaptive LiteSVM regression.

## TDD and validation

- RED commit: `27e0b0fe` against exact parent `96b332d6`.
- Exact-parent SBF SHA-256: `a04f9ef2644acef23a7714f3fdd14dfbdc5ac1da1aada1086c09cb7cb9e4d92e`.
- Fixed commit: `8d7b32e8`.
- Fixed SBF SHA-256: `ea93ff175732f82f72abedea5446f9067ce137fae660dbb4648ec4f38e2493c5`.
- Targeted exploit: pass.
- Full LiteSVM suite: 573 passed.
- Library tests: 5 passed.
- Kani test target: compiles.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2572,6 +2572,7 @@ pub mod ix {
         ResolveMarket,
         TopUpBackingBucket {
             domain: u16,
+            market_id: u64,
             amount: u128,
             expiry_slot: u64,
         },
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #322, opened 2026-07-23
[BLOCKER LoF] Bind oracle updates to asset generations

aeyakovenko:
## Root cause

The signed oracle capability family authenticated only `asset_index` plus the configured oracle key. It did not bind `ConfigureHybridOracle` (tag 34), `ConfigureEwmaMark` (35), `PushEwmaMark` (36), `ConfigureAuthMark` (62), or `PushAuthMark` (63) to the current asset generation. After retirement and permissionless same-index reuse, a creator could name an independent generation-A oracle key and replay retained generation-A signatures into generation B without possessing that key.

## Public loss proof

The LiteSVM regression starts from exact parent `8d7b32e8` and uses only public lifecycle, trade, crank, conversion, and withdrawal instructions. A retained AuthMark configuration and price push land on the replacement asset, move a real 20-unit position from 100 to 105, and settle 100 atoms from an independent victim to the attacker. The attacker withdraws 1,100 from a 1,000 deposit while the victim recovers 900.

RED commit: `938faf4c`
Exact-parent SBF SHA-256: `ea93ff175732f82f72abedea5446f9067ce137fae660dbb4648ec4f38e2493c5`

## Fix

- Add `market_id` to every oracle configure/push wire format.
- Reject `AssetGenerationMismatch` before profile, engine, or collateral state can mutate.
- Keep current-generation configure/push behavior intact.
- Update all callers, symbolic payload proofs, and the README ABI notes.

Production fix commit: `c40e3583`
Formal payload migration: `8d3b4636`

This adds no admin authority and no fund-moving path.

## Validation

- Adaptive exploit: RED on exact parent, GREEN on fixed SBF.
- Fixed SBF SHA-256: `f7c2823f8b2bf65a5c09a263af9a7c2160b9232b578affd199f2cd4927252f4a`.
- `cargo test --test v16_cu --quiet`: 574 passed.
- `cargo test --lib --quiet`: 5 passed.
- `cargo kani --tests --only-codegen`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,7 @@ This section describes intent and operational ordering, not argument-by-argument
   accounting.
 - AuthMark markets use **ConfigureAuthMark** (tag 62) and **PushAuthMark** (tag 63), signed by the configured mark authority, to store a direct authority mark without EWMA smoothing.
 - EwmaMark markets use **ConfigureEwmaMark** (tag 35) and **PushEwmaMark** (tag 36), signed by the configured mark authority, to update a smoothed EWMA mark input.
+- Every oracle configure/push instruction includes the asset's current `market_id`; a signature for a retired generation cannot configure or move a replacement at the same `asset_index`.
 - The per-slot effective-price movement cap is a risk parameter set at init; there is no standalone `SetOraclePriceCap` instruction in the current ABI.
 
 ### Insurance management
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #323, opened 2026-07-23
[BLOCKER DoS] Bind market resolution to its generation

aeyakovenko:
## Root cause

`ResolveMarket` was an unscoped signed capability: tag 19 carried no market generation. After the market account closed and was reinitialized at the same address under a new base-asset `market_id`, any retained generation-A resolve transaction still authenticated against the reused `marketauth` key and could irreversibly resolve generation B.

## Public DoS proof

The adaptive LiteSVM regression pre-signs both reinitialization and the stale resolve before generation A closes. The transactions are relayed without the admin private key. Generation B is then live, independently funded, and has a real open position before the stale A resolve lands. The market enters terminal resolved mode, a normal position-closing trade fails, and no instruction can transition that generation back to live.

RED commit: `026c9d99`
Exact-parent commit: `8d3b4636`
Exact-parent SBF SHA-256: `f7c2823f8b2bf65a5c09a263af9a7c2160b9232b578affd199f2cd4927252f4a`

## Fix

- Change the wire to `ResolveMarket { market_id }`.
- Reject `AssetGenerationMismatch` before the terminal state transition.
- Update every caller and the README authority-surface documentation.
- Preserve current-generation resolution behavior.

Production fix commit: `5ce20455`
Formal payload migration: `b8e439e1`

This adds no admin authority, recovery override, or fund-moving path.

## Validation

- Adaptive exploit: RED on exact parent, GREEN on fixed SBF.
- Fixed SBF SHA-256: `686fa3afe3248a3b48c652eff616e076e5c8b87877d714cb0a4799fe419a397e`.
- `cargo test --test v16_cu --quiet`: 575 passed.
- `cargo test --lib --quiet`: 5 passed.
- `cargo kani --tests --only-codegen`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
   `base_unit_authority`). `marketauth` is the only key that can: **create market 0** (`InitMarket`),
   **create/retire assets 1..N and set the permissionless-create-fee policy**, **safely force-shutdown
   any asset including asset 0** (`ASSET_ACTION_SHUTDOWN` → RECOVERY with the `force_close_delay_slots`
-  exit window so traders can exit), **resolve/close the market** (`ResolveMarket`/`CloseSlab`),
+  exit window so traders can exit), **resolve/close the market** (`ResolveMarket { market_id }`/`CloseSlab`),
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #324, opened 2026-07-23
[BLOCKER DoS] Bind stale resolution policy to market generation

aeyakovenko:
## Classification

REAL persistent public DoS reachable through signed public transactions on a normally initialized replacement market.

## Exploit

1. While generation A is live, retain a valid marketauth-signed `ConfigurePermissionlessResolve` transaction with one-slot stale and force-close timers, plus a valid transaction that can reinitialize the slab after closure.
2. Resolve and close generation A through its legitimate lifecycle.
3. Any relay reinitializes generation B at the same market address from the retained init transaction.
4. Independent users deposit and open a real matched position in generation B.
5. A public relay submits the retained generation-A policy. Before this fix it mutates generation B without the current admin key.
6. After one slot, any signerless caller invokes `ResolveStalePermissionless`. Generation B enters terminal `Resolved` mode, ordinary users cannot close through trading, and there is no bounded transition back to `Live`.

This is not transaction ordering or rollback-only griefing: the attacker leaves a funded, live replacement market in a persistent terminal state.

## TDD

- Exact parent: `b8e439e1`
- Exact-parent SBF SHA-256: `686fa3afe3248a3b48c652eff616e076e5c8b87877d714cb0a4799fe419a397e`
- RED commit: `1e055c9c`
- RED result: retained generation-A policy lands, public stale resolution succeeds, and a real generation-B close cannot progress.
- Fix commit: `e23031fa`
- Fixed SBF SHA-256: `e8a75bcaf13ea889cff991163786accf2ad050526dbb4423230fdf1381629b05`
- GREEN result: stale policy rejects with `AssetGenerationMismatch`, the market remains byte-identical and live, users close normally, and a current-generation policy remains usable.

## Fix

Tag 38 now carries the current base-asset `market_id`. The handler checks it before authority validation or policy mutation. All callers, README documentation, and symbolic wire-layout checks are migrated. No authority or fund-moving surface is added.

## Verification

- `cargo test --test v16_cu --quiet`: 576 passed
- `git diff --check`: passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,6 +394,10 @@ This section describes intent and operational ordering, not argument-by-argument
   - per-asset authorities (insurance/operator/backing/oracle, incl. asset 0) are rotated via
     `UpdateAssetAuthority { asset_index, market_id, kind, new_pubkey }`, which requires the target
     asset's current generation, not this instruction
+- **ConfigurePermissionlessResolve** (tag 38)
+  - `ConfigurePermissionlessResolve { market_id, stale_slots, force_close_delay_slots }` is signed by
+    `marketauth`; `market_id` must match the current base-asset generation so a policy signed for a
+    retired slab cannot arm permissionless shutdown on a replacement at the same address
 - **UpdateAssetLifecycle** (tag 40)
   - appends/reactivates/retires assets 1..N, including permissionless create/reuse when the configured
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #325, opened 2026-07-23
[BLOCKER LoF] Bind maintenance fee policy to market generation

aeyakovenko:
## Classification

REAL public loss of funds: a retained generation-A fee split lets an unprivileged cranker extract generation-B user-paid collateral from the canonical vault.

## Exploit

1. Retain a valid generation-A `UpdateMaintenanceFeePolicy` transaction setting the cranker share to 100%, plus a valid transaction that can reinitialize the slab after closure.
2. Resolve and close generation A through its legitimate lifecycle.
3. A public relay reinitializes generation B at the same address. Its intended cranker share is zero.
4. Independent users fund the market and open real matched OI; a separate independent depositor remains eligible for the account maintenance fee.
5. A public relay submits the retained generation-A policy without the current admin key.
6. At the next fee sync, the depositor pays 580 collateral atoms. Instead of remaining in canonical insurance, all 580 become attacker cranker capital.
7. The attacker withdraws all 580 atoms from the canonical token vault in the same slot.

This is an actual custody withdrawal, not points inflation, an opportunity-only change, or rollback griefing.

## TDD

- Exact parent: `e23031fa`
- Exact-parent SBF SHA-256: `e8a75bcaf13ea889cff991163786accf2ad050526dbb4423230fdf1381629b05`
- RED commit: `752a02f2`
- RED result: stale policy lands, a user pays 580 atoms, the attacker receives 580 internal capital and withdraws 580 vault tokens.
- Fix commit: `731463ba`
- Fixed SBF SHA-256: `cbfad08ac369f35f64b8f4e0b2395d230edb844e0d3bf66b3b3e734062ed11b0`
- GREEN result: stale policy rejects with `AssetGenerationMismatch`; the first 580 atoms remain in insurance, while a correctly bound current-generation policy still routes the next fee as configured.

## Fix

Tag 49 now carries the current base-asset `market_id`. The handler checks it before authority validation or fee-policy mutation. All callers, README documentation, and symbolic wire-layout checks are migrated. No authority, recovery override, or new fund-moving path is added.

## Verification

- `cargo test --test v16_cu --quiet`: 577 passed
- `cargo fmt --all`: passed
- `git diff --check`: passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,10 @@ This section describes intent and operational ordering, not argument-by-argument
   - resolved-market payout/finalization path for a supplied portfolio; pays only the stored owner token account
 
 ### Risk / maintenance
+- **UpdateMaintenanceFeePolicy** (tag 49)
+  - `UpdateMaintenanceFeePolicy { market_id, cranker_share_bps }` is signed by `marketauth`;
+    `market_id` must match the current base-asset generation so a retired slab's fee split cannot
+    redirect replacement-market user fees into a public cranker portfolio
 - **PermissionlessCrank** (tag 5)
   - permissionless maintenance entrypoint for a supplied asset/account path
   - authenticates clock/oracle state in the wrapper, then delegates bounded public progress to the engine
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #326, opened 2026-07-23
[BLOCKER LoF] Bind liquidation fee policy to market generation

aeyakovenko:
## Impact

A retained market-authority transaction from generation A can set the liquidation cranker share to 100% after the same market address is reinitialized as generation B. Once independent users open matched OI and a legitimate oracle move liquidates a position, any public cranker can receive the entire victim-paid liquidation fee, withdraw it from canonical custody, and leave insurance with zero fee share.

This is a public, extractable loss against a normally initialized live replacement market. It does not depend on injected exploit state or an administrator moving funds.

## Root cause

Tag 37 authenticated the market authority but did not bind the signed policy update to the market generation. Reusing the account address therefore extended an old signed capability into the replacement market.

## Fix

- Add `market_id` to the tag-37 wire format.
- Reject a policy update when the supplied generation does not match asset 0.
- Update callers, decoding proofs, and the generation-binding documentation.
- Keep authority and fund-moving surfaces unchanged.

## TDD evidence

- Exact parent: `731463ba`
- Exact-parent SBF SHA-256: `cbfad08ac369f35f64b8f4e0b2395d230edb844e0d3bf66b3b3e734062ed11b0`
- RED commit: `ccbb452e`
- Fix commit: `32f9b381`
- Fixed SBF SHA-256: `8c267d12effff796ff560a44ecd38fee0304b4fde3a5fdb59090d9b79888369f`

The adaptive LiteSVM regression fails on the exact parent only after proving a nonzero liquidation reward, zero insurance delta, successful attacker withdrawal, and an equal canonical-vault decrease. It passes on the fix by proving the stale transaction rejects byte-atomically and the current zero-share policy sends the real fee to insurance.

## Validation

- `cargo test --test v16_cu --quiet`: 578 passed
- `git diff --check`: passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,10 @@ This section describes intent and operational ordering, not argument-by-argument
   - resolved-market payout/finalization path for a supplied portfolio; pays only the stored owner token account
 
 ### Risk / maintenance
+- **UpdateLiquidationFeePolicy** (tag 37)
+  - `UpdateLiquidationFeePolicy { market_id, cranker_share_bps }` is signed by `marketauth`;
+    `market_id` must match the current base-asset generation so a retired slab's fee split cannot
+    redirect replacement-market liquidation penalties into a public cranker portfolio
 - **UpdateMaintenanceFeePolicy** (tag 49)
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #328, opened 2026-07-24
[BLOCKER LoF] Bind insurance withdrawals to asset generations

aeyakovenko:
## Root cause

`WithdrawInsuranceAsset` authenticated the operator and asset slot, but its signed payload did not bind the withdrawal to that slot's `market_id`. A valid transaction retained from one asset incarnation could therefore debit a replacement incarnation after public retire/reactivate reuse.

## Public exploit

The exact-parent LiteSVM RED test retains a fully signed generation-2 withdrawal, clears and retires the old asset through public instructions, reactivates the same slot, and has an independent provider fund generation 3 while legitimately reusing the operator service. Relaying the retained transaction within the recent-blockhash window succeeds on the parent, transfers 50,000 atoms to the old destination, and zeros the replacement provider's domain insurance.

No protocol-state injection, dishonest oracle, compromised current key, or failed-transaction mutation is involved. The transaction succeeds, so SVM rollback does not mitigate it.

## Fix

- Add the expected asset `market_id` to the tag-57 wire payload.
- Reject a generation mismatch before custody processing and recheck at the mutation boundary.
- Reject the legacy 19-byte unbound payload.
- Update positive callers and document the ABI.

This PR is intentionally stacked on #231 (`codex/probe-asset-generation-replay`), which introduces the persistent generation source used by the guard.

## Verification

- Exact-parent focused RED: fails with `generation-2 withdrawal drained generation-3 insurance`
- Fixed focused GREEN
- `cargo test --all-targets`: 569/569 integration tests plus unit/example targets pass
- `cargo check --all-targets`
- Kani wire-field preservation: 829 checks, zero failures
- Kani legacy-payload rejection: 821 checks, zero failures
- Kani custody trailing-byte rejection: 1,518 checks, zero failures
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,6 +480,8 @@ This section describes intent and operational ordering, not argument-by-argument
   - requires market resolved and all accounts closed
 - **WithdrawInsuranceAsset** (tag 57)
   - live/recovery asset-scoped withdrawal from that asset's funded long+short insurance budget
+  - carries the current asset `market_id`; retained withdrawals signed for a retired generation
+    reject before any budget or token mutation
   - gated by that asset's `insurance_operator`; `marketauth` can only use the shutdown-drain path
     for nonzero assets after the shutdown timeout and after the asset is empty
   - the asset `insurance_authority` funds the asset domains and owns terminal recovery accounting,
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #329, opened 2026-07-24
[BLOCKER LoF] Round composite oracle prices once

### LINKED ISSUE #327: Intermediate E6 rounding in three-leg composite oracles can cause false liquidations
Validated commit: `da64be639168b496833dd4c701607a94fcb06b6c`

## Summary

Three-leg composite oracle prices are evaluated one leg at a time in E6 integer precision. The implementation floors the accumulator after every multiplication or division, then uses that rounded integer for the next leg.

For a three-leg divide composite, the program calculates:

```text
floor(floor(A * 1e6 / B) * 1e6 / C)
```

The E6 value of the configured mathematical cross-rate is:

```text
floor(A * 1e6 * 1e6 / (B * C))
```

These are not equivalent. A later division can amplify a fraction discarded by an earlier division into a large price error. The resulting mark is accepted and propagated into PnL, margin, liquidation, and liquidation-fee accounting.

This was reproduced against the deployable BPF. With an exact cross-rate that remained constant at `1,500,000`, the affected program stored `1,000,000`, falsely liquidated a healthy long, charged a liquidation fee, paid part of it to a permissionless cranker, and allowed the cranker to withdraw the reward as SPL tokens.

## Description

[`compose`](https://github.com/aeyakovenko/percolator-prog/blob/da64be639168b496833dd4c701607a94fcb06b6c/src/v16_program.rs#L3940-L3958) calculates each leg and immediately converts the result back to a `u64`:

```rust
let next = if divide {
    (acc_e6 as u128)
        .checked_mul(1_000_000)?
        / leg_e6 as u128
} else {
    (acc_e6 as u128)
        .checked_mul(leg_e6 as u128)?
        / 1_000_000
};
```

Both external-oracle readers pass that rounded value into the following leg:

- [`read_external_price_e6`](https://github.com/aeyakovenko/percolator-prog/blob/da64be639168b496833dd4c701607a94fcb06b6c/src/v16_program.rs#L3961-L4019)
- [`read_external_price_e6_profile`](https://github.com/aeyakovenko/percolator-prog/blob/da64be639168b496833dd4c701607a94fcb06b6c/src/v16_program.rs#L4022-L4080)

Consider two sets of authenticated feed values:

```text
Initial: A=3, B=1,000,000, C=2
Later:   A=3, B=2,000,000, C=1
```

Both have the same exact cross-rate:

```text
3 * 1e6 * 1e6 / (1,000,000 * 2) = 1,500,000
3 * 1e6 * 1e6 / (2,000,000 * 1) = 1,500,000
```

The initial values also produce `1,500,000` sequentially. For the later values, however:

```text
first divide:
floor(3 * 1e6 / 2,000,000) = floor(1.5) = 1

second divide:
floor(1 * 1e6 / 1) = 1,000,000
```

The first operation discards `0.5` of one intermediate E6 unit. The final one-micro-unit divisor magnifies that discarded fraction into a `500,000`-unit error.

## Impact and executed reproduction

The reproduction used the default-feature BPF built from the affected commit:

```text
BPF SHA-256:
74e7c6d845ba0eef4b5d8a63d3540439063b56ca203a6f6d4cd3b6b045cae7ae
```

It exercised the production instruction paths for hybrid-oracle configuration, deposits, `TradeNoCpi`, authenticated oracle refresh, `PermissionlessCrank`, liquidation, and withdrawal.

The market used the repository's `production_risk_params()` profile:

```text
maintenance margin:          5% (500 bps)
initial margin:              5% (500 bps)
liquidation fee:           0.05% (5 bps)
price movement cap:        0.24% per slot (24 bps)
maximum accrual interval:     20 slots
cranker fee share:            50%
```

Test sequence:

1. Configure the three-leg divide composite with the initial feeds above.
2. Deposit `540,000` collateral into each side.
3. Open a one-unit long and short at `1,500,000`.
4. Publish the later authenticated feed values. The exact cross-rate remains `1,500,000`.
5. Repeatedly crank with fresh values. The 24-bps movement cap delays the affected mark's convergence but does not prevent it.
6. Refresh and permissionlessly liquidate the long, supplying the cranker’s portfolio as the reward account.
7. Withdraw the cranker reward.

Affected BPF result:

```text
exact cross-rate:                 1,500,000
stored mark:                      1,000,000
long certified equity:               40,000
long maintenance requirement:        50,000
long liquidation deficit:             10,000

long OI before liquidation:        1,000,000
long OI after liquidation:           797,979
victim capital before liquidation:    40,000
victim capital after liquidation:     39,898
total liquidation fee charged:           102
cranker reward withdrawn:                   51
```

The remaining `51` fee units were retained by insurance.

The same sequence against a temporary single-round implementation produced:

```text
stored mark:                      1,500,000
long certified equity:              540,000
long maintenance requirement:        75,000
long liquidation deficit:                  0
long OI after crank:               1,000,000
victim capital after crank:          540,000
cranker reward:                            0
```

The affected arithmetic therefore caused both:

- forced reduction of a position that was healthy at the correct price; and
- a collateral fee that a third party could withdraw from the market vault.

The incorrect mark can also affect settled PnL, initial-margin checks, liquidation deficits, and whether an actually unhealthy account is liquidated.

## Preconditions

- The asset uses a three-leg composite whose operation order produces a non-integral intermediate result followed by an operation that amplifies the discarded remainder.
- The trusted per-asset oracle authority has configured the feed identities and divide flags. Reconfiguration is restricted to an active asset with no relevant position or loss state.
- Authenticated feed values later reach a rounding-sensitive combination while users have exposure. A permissionless attacker cannot choose arbitrary feed accounts or manufacture publisher values.
- A victim's account is healthy at the correctly evaluated cross-rate but unhealthy at the stored rounded mark.
- Direct cranker extraction requires a nonzero liquidation fee and nonzero `liquidation_cranker_fee_share_bps`. If the cranker share is zero, false liquidation and the victim's fee loss can still occur, but the fee remains in insurance.
- No currently deployed market has been identified that uses a vulnerable composite. The minimal example uses unusually small feed values, although the defect applies generally to scale-cancelling composites.

Once those conditions exist, the attacker only needs normal permissionless crank access. They can select the victim portfolio, supply their own reward portfolio, receive the configured fee share, and withdraw it.

## Recommended fix

Evaluate all configured legs as one checked rational expression and round only when producing the final E6 value. Do not replace the retained numerator with the floored quotient between legs.

For example:

```text
num = leg0
den = 1

multiply leg:
    num *= leg
    den *= 1e6

divide leg:
    num *= 1e6
    den *= leg

result = floor(num / den)
```

Use checked `u128` arithmetic and reject a zero denominator. With the current three-leg cap and `MAX_ORACLE_PRICE`, the largest required numerator and denominator fit within `u128`.

If the current per-leg zero and maximum-price checks are intentional, calculate `floor(num / den)` for those checks without discarding the retained remainder. Apply the same shared arithmetic to both external-oracle readers. If final inversion is enabled, apply it to the retained rational value before the final rounding so that an already-rounded composite is not rounded a second time.

aeyakovenko:
## What changed

- Replace sequential per-leg E6 flooring with one checked rational numerator/denominator carried across all configured oracle legs.
- Apply inversion to the retained fraction and round only the final transformed E6 output.
- Preserve the existing per-leg zero and maximum-price gates without feeding their rounded quotient into the next leg.
- Add a public LiteSVM TDD regression with production risk parameters, authenticated three-leg Pyth observations, the real dt clamp, permissionless liquidation, cranker reward credit, and SPL withdrawal.

## Root cause and impact

For divide composites the wrapper evaluated `floor(floor(A*1e6/B)*1e6/C)`. A later small divisor can amplify the fraction discarded by the first divide. Two authenticated feed tuples with the same exact cross-rate of 1,500,000 therefore moved the stored target and mark to 1,000,000.

The exact-parent RED at `ccba79ed` opened an independent victim long and then used only permissionless cranks. The bad mark reduced long OI from 1,000,000 to 797,979, charged victim capital from 40,000 to 39,898, credited 51 atoms to the cranker, and allowed all 51 atoms to be withdrawn. The committed liquidation used 290,371 CU. Neither trader signed an oracle update or liquidation, and the oracle values themselves preserved the configured mathematical rate.

The fixed head keeps target and effective mark at 1,500,000, leaves the health certificate non-liquidatable, and preserves OI, victim capital, and cranker balance. Valid three-leg products are bounded by 1e36, below `u128::MAX`; checked arithmetic remains in place.

This is wrapper-specific oracle arithmetic. The engine pin remains `143e68c4917ed0400a27b952f036a5677047cd84`.

Fixes #327.

## Verification

- Exact-parent public SBF RED at `ccba79ed` with committed false liquidation and extracted SPL reward
- Fixed public SBF GREEN at `9cbfcda1`
- Built `tests/fixtures/auth_matcher` and `tests/fixtures/hostile_matcher` with platform tools v1.52
- `cargo test --all-targets`: 6/6 unit and 566/566 public LiteSVM/CU tests passed
- New configuration and crank paths remain below `CUSTODY_CU_LIMIT` and `CRANK_CU_LIMIT`
- `cargo check --all-targets`
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -3658,22 +3658,6 @@ pub mod oracle_v16 {
         feeds[2] != [0u8; 32] && feeds[2] != feeds[0] && feeds[2] != feeds[1]
     }
 
-    fn leg_divides(config: &WrapperConfigV16, idx: usize) -> bool {
-        match idx {
-            1 => (config.oracle_leg_flags & ORACLE_LEG_FLAG_DIVIDE_LEG2) != 0,
-            2 => (config.oracle_leg_flags & ORACLE_LEG_FLAG_DIVIDE_LEG3) != 0,
-            _ => false,
-        }
-    }
-
-    fn profile_leg_divides(profile: &AssetOracleProfileV16, idx: usize) -> bool {
-        match idx {
-            1 => (profile.oracle_leg_flags & ORACLE_LEG_FLAG_DIVIDE_LEG2) != 0,
-            2 => (profile.oracle_leg_flags & ORACLE_LEG_FLAG_DIVIDE_LEG3) != 0,
-            _ => false,
-        }
-    }
-
     pub fn read_pyth_price_e6(
         price_ai: &AccountInfo,
         expected_feed_id: &[u8; 32],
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #331, opened 2026-07-24
[BLOCKER LoF] Reject temporally skewed composite oracle reports

aeyakovenko:
## Root cause

Composite oracle legs were authenticated and freshness-checked independently, but the wrapper never required the submitted legs to describe one coherent observation time. A permissionless caller could combine a newer numerator report with a retained-but-still-fresh denominator report. The resulting false cross-rate became the oracle target and could drive independent users into liquidation.

## Public RED

The test-first commit `e7c5587f` uses only public LiteSVM instructions and fully verified Pyth-owned fixtures:

- configure a two-leg A/B cross-rate at `3@100 / 2@100 = 1.5`
- open an independent victim short at 1.5
- submit `6@101 / 2@100 = 3.0`, although the honest synchronized tuple is `6@101 / 4@101 = 1.5`
- let bounded permissionless cranks move the mark and liquidate

On the exact parent, victim capital fell from 540,000 to 143,420 and the permissionless attacker immediately withdrew a 41-atom liquidation reward. The liquidation committed at 284,354 CU.

## Fix

Treat a multi-leg composite as one oracle observation: all authenticated legs must have the same publish time. A mismatched tuple returns `OracleStale`; single-leg feeds are unchanged. Rejection leaves cached per-leg observations, target, mark, OI, and user balances unchanged through transaction rollback.

## Verification

- exact-parent public RED and fixed-head GREEN
- real `auth_matcher` and `hostile_matcher` SBF fixtures
- `cargo test --all-targets`: 6/6 unit tests and 566/566 LiteSVM/CU tests
- existing synchronized two- and three-leg composite positive tests
- `cargo check --all-targets`
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -3979,7 +3979,7 @@ pub mod oracle_v16 {
         }
         let mut acc = 0u64;
         let mut advanced = false;
-        let mut max_publish_time = i64::MIN;
+        let mut composite_publish_time = 0i64;
         let mut i = 0usize;
         while i < count {
             let (price, publish_time) = read_oracle_price_e6(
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-020 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #332, opened 2026-07-25
[BLOCKER LoF] Stage authenticated mark targets before CPI fills

aeyakovenko:
## Summary

Stage every changed `PushAuthMark` and `PushEwmaMark` target into the engine in the same successful instruction. This increments `oracle_epoch` and lets the engine reject risk increases until a permissionless crank advances the bounded effective mark, while preserving risk-reducing exits at the current effective price.

## Public exploit

Exact-parent RED commit: `57865f81`

An honest AuthMark authority moved the mark from 100 to 200. Before this fix, only the wrapper profile changed; the engine raw target, effective price, and certificate epoch remained at 100. An unsigned taker then used the official pre-authorized CPI matcher to add 10,000 units at the superseded price, cranked the mark to 200, closed, converted PnL, and withdrew 1,000,100 collateral atoms taken from an independent LP.

The LiteSVM test uses only public instructions, honest mark authority behavior, the production SBF, and the official matcher fixture. It does not inject protocol state or assume a dishonest oracle or LP signature.

Fixed commit: `91c011f9cbd12ef97257f522ca52357a02c088a3`

The fixed test also proves the rejected CPI instruction rolls back market, portfolio, matcher sequence/context state, and that an existing position can still close during target/effective lag.

## Verification

- Exact-parent public LiteSVM exploit: RED with 1,000,100 atoms withdrawn from the LP
- Fixed exploit and rollback/exit controls: GREEN
- `cargo test --all-targets`: 6 unit + 566 LiteSVM/CU tests GREEN
- `cargo build-sbf --no-default-features`: GREEN
- Real `auth_matcher` and `hostile_matcher` SBF fixtures: GREEN
- Push CU: AuthMark 7,169; EwmaMark 7,730
- Kani active-payload truncation theorem: 823/823 checks GREEN

The aggregate wrapper Kani command then reached the unchanged unconstrained instruction-decoder frontier (10 GB RSS, unwind 132) and was stopped; it had not reached either changed push handler.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10211,6 +10211,9 @@ pub mod processor {
             profile.oracle_target_price_e6 = next_mark;
             profile.oracle_target_publish_time = 0;
             profile.last_good_oracle_slot = authenticated_slot;
+            group
+                .set_asset_raw_oracle_target_not_atomic(asset_index_usize, next_mark)
+                .map_err(map_v16_error)?;
             write_oracle_profile_to_view(&mut group, asset_index_usize, &profile)?;
             if asset_index_usize == 0 {
                 cfg.last_good_oracle_slot =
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #333, opened 2026-07-25
[BLOCKER LoF] Stage trade-driven mark targets before later fills

aeyakovenko:
## Summary

When an EWMA or stale-hybrid trade commits a fee-supported mark, stage that mark into both the wrapper oracle target and the engine raw target in the same instruction. The shared helper is called from the production single-trade and batch-trade paths, so it also covers their CPI entrypoints.

A changed target increments the engine oracle epoch. Later risk increases wait for a permissionless crank to advance the bounded effective mark, while reductions continue to execute at the current effective price.

## Public exploit

Exact-parent RED commit: `b29e3ece`

1. An attacker-owned one-unit self-trade at slot 1 reported 200 against an EWMA mark of 100 and fully paid the dynamic fee supporting the resulting mark of 150.
2. The attacker closed both discovery portfolios in the same slot, leaving no discovery exposure.
3. Before this fix, the wrapper mark was 150 but the engine raw target, effective price, and oracle epoch remained at 100.
4. A separate unsigned taker used the official matcher already authorized by an unrelated LP to open 10,000 units at 100. Because no slot elapsed, that second trade caused no additional EWMA movement and paid no mark-movement externality.
5. Public cranks advanced the effective mark to 150. The taker closed, converted PnL, and withdrew 500,000 collateral atoms taken from the LP.

The LiteSVM reproducer uses only public instructions, production SBF, honest configuration, and the official pre-authorized matcher. It does not mutate protocol state or assume a dishonest oracle or victim signature.

Fixed commit: `13df82247f8f2473d31a838671b409d01caed04f`

The fixed test proves the later CPI risk increase rejects atomically. A separate batch regression proves the independent batch profile-write path stages the same target/epoch and that both sides can still batch-reduce during target/effective lag.

## Verification

- Exact-parent public exploit: RED with 500,000 atoms withdrawn from an independent LP
- Fixed single exploit and batch reduction regression: GREEN
- `cargo test --all-targets`: 6 unit + 567 LiteSVM/CU tests GREEN
- `cargo build-sbf --no-default-features`: GREEN
- Real `auth_matcher` and `hostile_matcher` SBF fixtures: GREEN
- Changed single and one-leg batch routes are asserted below the existing 345,000 CU ceiling
- `cargo check --all-targets`, formatting, and diff checks: GREEN

Wrapper Kani currently covers wire/decoder properties rather than these stateful public trade handlers; the exact public SBF RED/GREEN and complete CU matrix are the applicable gates.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -5922,10 +5922,16 @@ pub mod processor {
                 asset_index as usize,
                 collected_post_trade_mark,
             )?;
+            stage_trade_driven_mark_target_view(
+                &mut oracle_profile,
+                &mut group,
+                asset_index as usize,
+            )?;
             write_oracle_profile_to_view(&mut group, asset_index as usize, &oracle_profile)?;
             if asset_index == 0 && oracle_v16::profile_is_price_managed(&oracle_profile) {
                 cfg.mark_ewma_e6 = oracle_profile.mark_ewma_e6;
                 cfg.mark_ewma_last_slot = oracle_profile.mark_ewma_last_slot;
+                cfg.oracle_target_price_e6 = oracle_profile.oracle_target_price_e6;
                 cfg_after = Some(cfg);
             }
             group.validate_shape().map_err(map_v16_error)?;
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #334, opened 2026-07-25
[BLOCKER LoF] Sequence-bind matcher config intents

aeyakovenko:
## Summary

- bind `SetMatcherConfig` to the current portfolio incarnation and matcher-config sequence
- advance the sequence on every successful enable or revoke while preserving the existing 104-byte zero-copy tail
- document the ABI change and prove its decoder fields with targeted Kani harnesses

## Root cause

`SetMatcherConfig` authenticated the LP owner but carried no freshness value. Two honest transactions signed under the same live recent blockhash could land out of order: a newer revoke could land first, then an unprivileged relayer could submit the older enable and restore permissionless matcher access.

The exact-parent public LiteSVM reproduction uses normal initialized accounts and signed transactions, with no protocol-state injection. After replaying the stale enable, an unsigned attacker trades against the independent LP, moves the honest AuthMark from 100 to 200, closes, converts PnL, and withdraws 1,000,000 collateral atoms taken from the LP.

## Fix

Tag 68 now carries `portfolio_id`, `expected_sequence`, and `enabled`. The program verifies both freshness fields and increments the sequence before committing the config. The sequence occupies seven bytes adjacent to the one-byte enabled flag, so `PortfolioMatcherConfigV16` remains exactly 104 bytes. Legacy zero tails begin at sequence zero; the incarnation binding rejects intents from closed/recreated portfolios.

## TDD evidence

- exact-parent RED commit: `12f1fa08` (`v16_attack_delayed_matcher_enable_cannot_override_newer_revoke` loses 1,000,000 LP atoms)
- fixed GREEN commit: `93f7225c`
- `cargo test --test v16_cu -- --test-threads=1`: 566/566 passed, including max-shape CU tests
- `cargo test --lib -- --test-threads=1`: 7/7 passed
- `cargo kani --tests --harness kani_v16_set_matcher_config_decode_preserves_intent_binding`: passed
- `cargo kani --tests --harness kani_v16_every_active_payload_rejects_one_byte_truncation`: passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,14 +352,17 @@ This makes it a "pure identity signer" and prevents it from becoming an attack s
 ### LP matcher config (TradeCpi / BatchTradeCpi)
 Unsigned LP matcher fills require an enabled matcher config stored directly on the LP portfolio.
 
-`SetMatcherConfig` (tag 68) is signed by the LP owner and writes:
+`SetMatcherConfig` (tag 68) is signed by the LP owner, supplies the current portfolio incarnation
+ID and matcher-config sequence, and writes:
 
-`matcher_program, matcher_context, matcher_delegate, enabled`
+`matcher_program, matcher_context, matcher_delegate, enabled, next_sequence`
 
 During `TradeCpi` / `BatchTradeCpi`, Percolator reads this LP-account config and requires the
 instruction's matcher program, matcher context, and matcher delegate PDA to match it byte-for-byte.
 Extra matcher CPI tail accounts begin immediately after the matcher delegate. Disabled configs,
 wrong matcher program/context/delegate, wrong LP owner, or wrong LP portfolio all reject.
+Every successful config update advances the sequence, so delayed signed grants cannot override a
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Independent current-pin confirmation from percolator-meta: https://github.com/aeyakovenko/percolator-meta/pull/383 (`4722904f9180aede6ea2a6d135426ebab16be3e3`).

The clean-room LiteSVM proof system-allocates/publicly initializes every Percolator state account and runs the pinned real Percolator/authenticated-matcher SBFs. The independent LP starts with 4,000,000 and the attacker with 2,000,000. The LP signs an enable and newer revoke, but neither exploit trade. After revoke-first then delayed-enable ordering, attacker-only open/close transactions execute across an honest `100 -> 200` AuthMark move. Public conversion/withdrawal ends at LP 3,000,000 and attacker 3,000,000, independently confirming an exact 1,000,000 attacker-extractable LP loss.

aeyakovenko:

PR #335, opened 2026-07-25
[BLOCKER LoF] Reject delayed authenticated mark intents

aeyakovenko:
## Summary

An unprivileged relayer can reorder honest, fully signed authenticated-mark transactions that share a still-valid recent blockhash. Before this change, landing-slot freshness did not order either mark pushes or oracle-mode configuration. A withheld older intent could therefore replace a newer correction and create a false mark that an attacker trades against through the public CPI path.

Two exact-parent public LiteSVM RED tests demonstrate the same missing per-asset ordering invariant:

- Delayed `PushAuthMark`: land honest sequence 2 (mark 100) before withheld sequence 1 (mark 50), then settle and withdraw 500,000 collateral atoms from an independent LP.
- Delayed `ConfigureAuthMark`: land honest configuration sequence 2 (mark 100) before withheld configuration sequence 1 (mark 50), open through the official matcher, apply honest future sequence 3, then settle and withdraw 500,000 atoms from the LP.

Neither exploit uses state injection, a dishonest oracle value, a compromised key, or a victim signature.

## Fix

- Add signer-chosen `observation_sequence` to `ConfigureHybridOracle`, `ConfigureEwmaMark`, `ConfigureAuthMark`, `PushEwmaMark`, and `PushAuthMark`.
- Make all five instructions advance one shared per-asset watermark, including across mode changes.
- Persist the watermark in six bytes of existing oracle-profile padding, preserving the 400-byte zero-copy layout.
- Require strict monotonicity while allowing gaps and out-of-order transaction landing.
- Preserve the sequence across oracle-profile and asset-lifecycle rewrites.
- Reject values outside the persisted 48-bit range.
- Update direct API documentation, decoder tests, and Kani wire contracts.

Asset-generation binding remains a separate concern from PR #275.

## TDD evidence

- Configuration RED: `23a6e800`
- Push RED: `0ae218da` (same test as original `393ea841`)
- Push fix: `abf2198f`
- Comprehensive shared-sequence fix: `d5f21527`

Fixed-head controls cover stale AuthMark and EWMA pushes, cross-mode stale configurations, accepted sequence gaps, byte-identical rollback on rejection, and ordinary positive configuration/push paths.

## Verification

- Pinned v1.52 program, authenticated matcher, and hostile matcher SBF builds: PASS
- `cargo test --test v16_cu -- --test-threads=1`: 569/569 PASS, including all max-shape CU gates
- `cargo test --lib -- --test-threads=1`: 7/7 PASS
- `cargo check --all-targets`: PASS
- `cargo fmt --all -- --check`: PASS
- `git diff --check`: PASS

The updated Kani sources compile. The 164-byte hybrid field-preservation query was stopped after 10 minutes because the existing whole-enum decoder also symbolically explored unrelated variable-length crank/batch branches (unwind 94); no Kani pass is claimed for that final query.

### DIFF
--- a/README.md
+++ b/README.md
@@ -560,7 +560,7 @@ AuthMark and EwmaMark are authority-pushed pricing modes for markets that do not
 
 AuthMark is the direct authority-mark path:
 
-- **Direct mark API**: `ConfigureAuthMark { asset_index, now_slot, initial_mark_e6 }` and `PushAuthMark { asset_index, now_slot, mark_e6 }`.
+- **Direct mark API**: `ConfigureAuthMark { asset_index, now_slot, initial_mark_e6 }` and `PushAuthMark { asset_index, now_slot, mark_e6, observation_sequence }`. `PushEwmaMark` carries the same signer-chosen, strictly increasing `observation_sequence`. The program persists the last accepted sequence per asset across oracle-profile rewrites, so delayed same-incarnation reports cannot replace newer authenticated observations while gaps and out-of-order transaction landing remain supported.
 - **No EWMA configuration**: there is no halflife, mark-min-fee, feed id, confidence filter, invert flag, or unit-scale configuration in the AuthMark API.
 - **Authority boundary**: only the configured mark authority can push a new mark; public cranks can only consume the stored mark.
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #336, opened 2026-07-25
[BLOCKER LoF] Reject delayed liquidation-fee policy replay

aeyakovenko:
## Root cause

`UpdateLiquidationFeePolicy` authenticated the market authority but carried no intent ordering. Two honestly signed policy transactions could remain valid under the same recent blockhash and land in reverse order. An unprivileged relayer/cranker could therefore land a newer 0% correction first, wait for independent users to trade, then land the older 100% cranker-share intent immediately before liquidating a victim.

The exact-parent public LiteSVM RED test redirected 524 liquidation-fee atoms away from canonical insurance and then withdrew all 524 atoms through the public cranker portfolio path. It uses no protocol-state injection, dishonest oracle, initialization mistake, or compromised authority.

## Fix

- Add a signer-chosen `u64 policy_sequence` to tag 37.
- Require strict advancement over the last accepted liquidation-policy sequence; gaps are allowed, while zero, duplicate, and lower values reject without mutation.
- Reject legacy unsequenced and trailing-byte payloads through exact instruction framing.
- Preserve the 448-byte zero-copy market ABI by narrowing `permissionless_market_init_fee` to its already-enforced `u64` domain and using the former high 64-bit word for the sequence. Every publicly reachable legacy account has that word zero.
- Keep liquidation live after a rejected replay; the fee remains in canonical insurance when the accepted share is 0%.

## TDD history

- `7af894cf`: exact-parent RED reproduction against `36fa587e` / engine `143e68c`; 524 atoms publicly extracted.
- `32848fa2`: production fix and GREEN public-interface tests.
- `9f55856b`: dedicated exact-framing Kani proof.

## Verification

- Solana SBF build, tools v1.52
- 567/567 `v16_cu` LiteSVM/CU tests green
- 6/6 library tests green, including fixed `WrapperConfigV16` size
- Focused Kani field-preservation proof: 817 checks, 0 failures
- Focused Kani exact/short/trailing framing proof: 1,533 checks, 0 failures
- `cargo check --all-targets`
- `cargo fmt --check`
- `git diff --check`

Engine pin remains `143e68c4917ed0400a27b952f036a5677047cd84`; no engine change is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,10 @@ This section describes intent and operational ordering, not argument-by-argument
   - candidate accounts are untrusted hints, not a liveness precondition; honest keepers should include the worst known stale/bankrupt/liquidatable accounts, but the engine also makes cursored progress
   - may perform bounded catchup/recovery, liquidation, touch-only settlement, round-robin lifecycle progress, and empty-account reclaim
   - liquidation rewards are optional: when `liquidation_cranker_fee_share_bps > 0`, a keeper may append its writable Percolator portfolio account as the final account in the instruction. Oracle accounts remain immediately after the target portfolio account; the reward portfolio, if present, is last. If no reward portfolio is supplied, the full retained liquidation penalty stays in insurance.
+- **UpdateLiquidationFeePolicy** (tag 37)
+  - sets the optional cranker share and carries a market-authority-chosen `policy_sequence`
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #337, opened 2026-07-25
[BLOCKER LoF] Reject delayed maintenance-fee policy replay

aeyakovenko:
## What changed

- add an exact-parent public LiteSVM RED proving that a delayed, still-valid market-authority maintenance-fee policy can overwrite a newer correction
- sequence `UpdateMaintenanceFeePolicy` intents and reject zero, duplicate, lower, and delayed sequences while accepting forward gaps
- preserve the 448-byte zero-copy market config ABI by splitting the former deposits-only `u128` principal counter into its reachable `u64` value and a new `u64` sequence word
- add public boundary/rollback coverage and focused Kani wire-field and exact-framing proofs

## Impact and root cause

The instruction authenticated the market authority but had no intent ordering. An untrusted relayer could hold an older signed 100% cranker-share update, land the authority's newer 0% correction first, wait for an independent user to accrue maintenance fees, then land the old update. The exact-parent RED charges 580 atoms, diverts all 580 from canonical insurance to the relayer's cranker portfolio, and publicly withdraws them.

The fix requires `policy_sequence` to strictly exceed the last accepted value. Gaps remain valid so dropped transactions do not stall administration. Legacy three-byte payloads fail exact decoding.

## TDD history

- RED: `05dcf8eec135255235f2d154836d7fae553cf07b`
- GREEN: `eab4443a992190bea780184e18d400c66fc4bada`
- pinned engine: `143e68c4917ed0400a27b952f036a5677047cd84`

## Verification

- public LiteSVM/CU suite: 567/567 passed
- library tests: 6/6 passed
- `cargo check --all-targets`
- Kani exact/short/trailing framing: 1,533 checks, 0 failures


aeyakovenko:

PR #338, opened 2026-07-25
[BLOCKER LoF] Reject delayed trade-fee policy replay

aeyakovenko:
## What changed

- add an exact-parent public LiteSVM RED proving that a delayed, still-valid asset-0 insurance-authority trade-fee policy can overwrite a newer correction
- sequence `UpdateTradeFeePolicy` intents and reject zero, duplicate, lower, and delayed sequences while accepting forward gaps
- preserve the 448-byte zero-copy market config ABI by splitting the former publicly `u64`-bounded init-fee `u128` into its reachable low word and a new `u64` sequence word
- add public boundary/rollback coverage and focused Kani wire-field and exact-framing proofs

## Impact and root cause

Tag 55 authenticated the current authority but did not order that authority's signed intents. An untrusted relayer can retain an older 10,000 bps update, land the honest authority's newer zero-fee correction, let an independent victim sign a zero-fee trade on the visible policy, restore the old floor, and submit the already-signed trade.

The exact-parent RED uses an attacker-operated permissionless asset. The trade charges 1,000 atoms to each side; the attacker withdraws both 1,000-atom domain fees and nets exactly the independent victim's 1,000-atom loss. This is distinct from #256 (compromised-authority live fee hike) and #296 (cross-generation replay): it needs neither key compromise nor market reinitialization.

The fix requires `policy_sequence` to strictly exceed the last accepted value. Gaps remain valid so dropped transactions do not stall administration. Legacy nine-byte payloads fail exact decoding, and the independently signed trade remains executable at zero fee after the replay rejects.

## TDD history

- RED: `a6fa601cc509126505124894ed9ec85d8c67b7ed`
- GREEN: `fef78ea87a4dbf1e1032439351b60c4846608ae6`
- pinned engine: `143e68c4917ed0400a27b952f036a5677047cd84`

## Verification

- public LiteSVM/CU suite: 567/567 passed
- library tests: 6/6 passed
- `cargo check --all-targets`
- Kani exact/short/trailing framing: 1,533 checks, 0 failures

### DIFF
--- a/README.md
+++ b/README.md
@@ -429,6 +429,13 @@ This section describes intent and operational ordering, not argument-by-argument
   - transfers collateral into vault; credits insurance fund in engine
 
 ### Trading
+- **UpdateTradeFeePolicy** (tag 55)
+  - sets the market-wide minimum trade fee and carries an asset-0 insurance-authority-chosen
+    `policy_sequence`
+  - the sequence must strictly exceed the last accepted trade-fee sequence; gaps are allowed, while
+    delayed, duplicate, zero, and lower intents reject
+  - tag 55's payload is exactly `u64 trade_fee_base_bps || u64 policy_sequence`; legacy
+    unsequenced payloads reject
 - **TradeNoCpi**
   - trade without external matcher (used for testing / deterministic scenarios)
 - **TradeCpi**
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #339, opened 2026-07-25
[BLOCKER LoF] Bind backing top-ups to provider-approved fee terms

aeyakovenko:
## Root cause

`UpdateBackingFeePolicy` authenticated the domain insurance authority but did not bind an independent provider's backing deposit to the fee rate and insurance split that the provider accepted. An unprivileged relayer could reorder older and newer honestly signed transactions in either direction:

1. fund first, then land an older policy; or
2. land an older policy while the bucket is empty, then land the provider's already-signed top-up.

A subsequent public trade accrued a real utilization fee under the displaced split, and a distinct domain insurance operator could withdraw value owed to the backing provider.

## Change

- Freeze each backing domain's fee rate and insurance split while its bucket contains provider principal, liens, consumed/impaired backing, or unwithdrawn provider earnings.
- Extend `TopUpBackingBucket` with `expected_fee_bps` and `expected_insurance_share_bps`; reject atomically unless both match live policy before any accounting or token movement.
- Keep exact idempotent policy writes valid and allow reconfiguration after the bucket is fully empty.
- Update all wrapper call sites and add Kani wire-field, truncation, and trailing-byte coverage.
- Add public LiteSVM coverage for both transaction orders, byte-identical rollback, a matching-policy retry, real fee accrual, and provider/operator withdrawals.

This changes tag 24's payload from 27 to 31 bytes. It does not reject trades or exits; only policy changes over live provider value and deposits whose signed terms no longer match are rejected.

## Red/green evidence

- Exact main parent `36fa587e`: RED commit `9b8ac341` publicly diverts and extracts 70 atoms from an independent provider when the stale policy lands after funding.
- Incomplete funded-only fix `395149cd`: RED commit `af03de2b` publicly diverts and extracts the same 70 atoms when the stale policy lands immediately before the provider's retained top-up.
- Fixed head `400f24f6`: both orderings are closed. The stale top-up rolls back market, source, vault, and ledger byte-identically; after restoring the accepted terms, the provider can retry, earn the fee, and withdraw it.

No state injection, dishonest oracle, or compromised signer is required; the adversary only relays honest signed transactions within the live blockhash window.

## Validation

- backing-focused public LiteSVM/CU tests: 42 passed
- `cargo check --tests`
- `git diff --check`
- Kani `kani_v16_top_up_backing_bucket_decode_preserves_wire_fields`: 821 checks passed
- Kani `kani_v16_custody_payloads_reject_trailing_byte`: 1,500 checks passed
- Kani `kani_v16_every_active_payload_rejects_one_byte_truncation`: 849 checks passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,10 @@ The economic and governance model for a permissionless multi-asset market. Statu
     `v16_attack_backing_fee_split_conserves` drives a real BPF risk-increase that grows a
     counterparty-backing lien and asserts the fee splits with no leakage: `charged ==
     insurance_pool_delta + provider_delta`, `insurance_delta == floor(charged * share_bps)`, and the
-    per-domain insurance budget mirrors the insurance share.
+    per-domain insurance budget mirrors the insurance share. The
+    per-domain fee rate and insurance split are immutable while that backing bucket holds provider
+    principal, liens, consumed/impaired backing, or unwithdrawn earnings; idempotent writes remain
+    valid and an empty domain can be reconfigured.
 
 ### Isolation — traders are safe from other assets, even faulty ones
 Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guarantee:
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #340, opened 2026-07-25
[BLOCKER LoF] Reject delayed fee-redirect policy replay

aeyakovenko:
## Root cause

`UpdateFeeRedirectPolicy` authenticated `marketauth` but carried no intent ordering. Two honestly signed updates could remain valid under one recent blockhash and land in reverse order. An unprivileged relayer could therefore restore an old 0% redirect after a visible 100% correction, route a subsequent permissionless-asset trade's fees away from canonical asset-0 insurance, and let that asset's distinct operator withdraw them.

The exact-parent public LiteSVM RED keeps the fee amount fixed: an independent victim signs the same fee-bearing trade after the canonical redirect is visible. The delayed policy changes only the fee destination. The operator extracts 2,000 atoms and ends 1,000 ahead, equal to the victim/canonical fund's independent loss.

## Fix

- Add a signer-chosen `u64 policy_sequence` to tag 58.
- Require strict advancement; forward gaps succeed, while zero/equal/lower intents reject atomically.
- Reject legacy short and trailing-byte payloads through exact framing.
- Preserve the 448-byte zero-copy market ABI by splitting the publicly `u64`-bounded permissionless-init-fee `u128` into its existing low word plus a sequence high word. Publicly reachable legacy accounts start at sequence zero.
- Keep the already-signed trade executable under the accepted redirect.

This is same-generation intent ordering and is distinct from #317's market-generation binding. The two protections compose.

## Red/green evidence

- Exact parent `36fa587e`, RED commit `5af86122`: stale 0% intent lands after 100%; canonical insurance receives zero; attacker-operated asset withdraws 2,000 atoms.
- Fixed head `ae107419`: stale intent rejects; the trade executes; canonical insurance receives all 2,000 atoms; attacker extracts zero.

## Validation

- `cargo test --test v16_cu -- --nocapture`: 567 passed, including all max-shape CU gates
- Kani exact framing: 1,533 checks, 0 failures
- `cargo check --all-targets`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,9 @@ The economic and governance model for a permissionless multi-asset market. Statu
 - **Fee routing (configured percentages):**
   - a % of **all trading fees → asset-0 insurance** (`fee_redirect_to_market_0_bps`). **✅**
     (`v16_attack_fee_redirect_split_lands_correctly` asserts the 20% split + conservation;
-    `v16_attack_market0_fees_stay_local`, `..._fee_redirect_full_boundary`.)
+    `v16_attack_market0_fees_stay_local`, `..._fee_redirect_full_boundary`.) Redirect updates carry
+    a strictly increasing signer sequence, so delayed transactions cannot restore a superseded fee
+    destination.
   - a % of **asset-N backing yield → asset-N insurance** and **→ asset-0 insurance**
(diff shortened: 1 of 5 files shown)


aeyakovenko:

PR #342, opened 2026-07-25
[NOT REAL LoF][CUSTODY HARDENING] Make asset insurance withdrawals at-most-once

aeyakovenko:
## Summary

- add a monotonic per-asset insurance-withdrawal sequence in the existing wrapper reserve
- require tag 57 callers to provide the current `expected_sequence`
- consume the sequence atomically with the engine debit and SPL transfer
- reject legacy, truncated, trailing, or stale withdrawal payloads

## Impact

An honest insurance operator can sign two fee-bump variants of the same withdrawal under one recent blockhash. Before this change, an unprivileged relayer could land one, wait for the independent insurance authority to replenish the same asset domain, then land the retained variant and transfer the new contribution to the operator. The later top-up restored the amount/budget preconditions, so no committed-state incarnation distinguished the already-consumed intent.

The public LiteSVM reproduction uses the same asset, amount, destination, operator signature, and recent blockhash. The transaction variants differ only by a priority-fee instruction. It uses only normal public activation, top-up, and withdrawal instructions; there is no protocol-state injection.

## TDD evidence

- RED test commit: `1650b5ac`
- fixed commit: `f80cc5d7`
- pre-fix result: the retained retry succeeds after the independent top-up, transfers a second 50,000 atoms, and empties the replacement domain budget
- fixed result: the retained retry returns `EngineStale`, leaves market/vault custody byte-identical, preserves the replacement budget, and a freshly sequenced withdrawal remains live

## Verification

- `cargo test --all-targets -- --test-threads=1` (566 LiteSVM/CU tests + 6 unit tests)
- `cargo check --all-targets`
- Kani wire-field round-trip harness
- Kani legacy/truncated payload rejection harness
- Kani custody trailing-byte rejection harness

No engine change and no market account growth are required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,6 +466,8 @@ This section describes intent and operational ordering, not argument-by-argument
   - requires market resolved and all accounts closed
 - **WithdrawInsuranceAsset** (tag 57)
   - live/recovery asset-scoped withdrawal from that asset's funded long+short insurance budget
+  - carries the per-asset `expected_sequence`; every successful withdrawal advances the sequence, so
+    only one concurrently signed retry can spend a reserve even if a later top-up restores its balance
   - gated by that asset's `insurance_operator`; `marketauth` can only use the shutdown-drain path
     for nonzero assets after the shutdown timeout and after the asset is empty
   - the asset `insurance_authority` funds the asset domains and owns terminal recovery accounting,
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Strict live-instance LoF/DoS re-audit: the published retry is real intent ABA, but not an unprivileged attacker-extractable blocker.

The retained tag-57 transaction is signed by the current insurance operator and can transfer only to a token account owned by that operator. The relayer receives no value. If the operator is malicious it can authorize a fresh current withdrawal; if it is honest, the replacement contribution lands in its own custody and is not stolen by the relayer. The market remains terminable, so this is not persistent DoS.

In the Meta path the operator is the Subledger pool PDA, which signs only an in-transaction CPI after owner-claim checks and cannot produce an externally retained signed transaction variant.

Closing as signed-intent/custody hardening rather than REAL LoF.

PR #344, opened 2026-07-25
[BLOCKER LoF] Make insurance top-ups one-shot across retry variants

aeyakovenko:
## Summary

- add a checked per-asset insurance top-up sequence in the existing wrapper reserve
- require both `TopUpInsurance` and `TopUpInsuranceDomain` to carry the current `expected_sequence`
- consume the shared sequence atomically with the insurance credit and SPL debit
- reject legacy, truncated, trailing, and stale top-up payloads

## Impact

An honest insurance authority can sign two priority-fee variants of one top-up under the same recent blockhash. Before this change, an unprivileged relayer could land one, wait for an ordinary SPL transfer to replenish the exact source token account, then land the retained variant and debit the authority again.

The public LiteSVM reproduction configures the intended two-key split: a cold insurance authority funds the domain and a distinct live operator can withdraw it. After the retained top-up lands, the operator withdraws exactly the duplicated 50,000 atoms while the authority's intended first 50,000-atom contribution remains insured. The exploit uses only public asset activation, SPL transfer, top-up, and withdrawal instructions; it does not mutate protocol state or require a dishonest oracle.

## Fix

Each asset stores a monotonic top-up sequence in its existing 112-byte wrapper reserve. Asset 0's tag 9 route and both tag 56 side domains share that asset's sequence. A successful top-up advances it with checked arithmetic; stale retry variants return `EngineStale`. SVM rollback keeps the sequence, market accounting, ledger, and SPL custody atomic when any later check or transfer fails.

This changes only the wrapper ABI and wrapper-owned state. The engine pin and engine code are unchanged.

## TDD evidence

- RED commit: `e06c3eb18735d3b3b22c70f031503f10292c2e7f`
- fixed commit: `64dd4dd969738d3b7252cd621daf6f06d9e4b619`
- pre-fix: retained retry succeeds, doubles the domain budget, debits the replenished source, and the distinct operator extracts the duplicate
- fixed: retained retry returns `EngineStale`, touched accounts remain byte-identical, the replenished source remains funded, and a fresh sequence-1 top-up succeeds

## Verification

- SBF builds for `percolator-match`, `auth_matcher`, and `hostile_matcher`
- `cargo test --all-targets -- --test-threads=1` (566/566 LiteSVM/CU tests, 6/6 unit tests)
- `cargo check --all-targets`
- Kani top-up sequence/amount wire-field round trip
- Kani legacy/truncated top-up payload rejection
- Kani custody trailing-byte rejection

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,6 +427,10 @@ This section describes intent and operational ordering, not argument-by-argument
   - validates side encoding and engine readiness; it is not an admin override
 - **TopUpInsurance** (tag 9)
   - transfers collateral into vault; credits insurance fund in engine
+  - carries asset 0's current `expected_sequence`; every successful top-up advances the sequence,
+    so only one concurrently signed retry can debit a replenished source account
+- **TopUpInsuranceDomain** (tag 56)
+  - funds one asset-side insurance domain and shares that asset's top-up sequence across both sides
 
 ### Trading
 - **TradeNoCpi**
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -49,6 +49,9 @@ pub mod constants {
     pub const WRAPPER_CONFIG_LEN: usize = 448;
     pub const ASSET_ORACLE_PROFILE_LEN: usize = 400;
     pub const ASSET_ORACLE_WRAPPER_LEN: usize = 512;
+    // The first wrapper-reserve word is assigned to live insurance-withdrawal intents.
+    pub const ASSET_INSURANCE_TOPUP_SEQUENCE_OFF: usize = ASSET_ORACLE_PROFILE_LEN + 8;
+    pub const ASSET_INSURANCE_TOPUP_SEQUENCE_LEN: usize = 8;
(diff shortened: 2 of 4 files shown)


aeyakovenko:
A stronger strict extraction path is confirmed on the current Meta pin and real pinned SBF binary: https://github.com/aeyakovenko/percolator-meta/pull/377 (`86f4fbc3c67cd111deb9317a247ac0d54282fdcb`).

The relayer owns the winning portfolio and uses no insurance-operator/admin key. After the provider source receives replacement funds, the retained asset-0 top-up lands and ordinary honest-mark terminal settlement changes `(provider source=500, attacker=2250, loser=0)` to `(provider source=0, attacker=2500, loser=0)`. Asset-0's side split leaves the other 250 duplicated atoms in insurance. Every Percolator state account is system-allocated/publicly initialized; no initialized protocol state is injected.

aeyakovenko:

PR #345, opened 2026-07-25
[BLOCKER LoF] Prevent market-authority handoff ABA replay

aeyakovenko:
## Impact

A displaced incoming market authority can retain a fully signed `A -> B` handoff. If governance legitimately rotates `A -> C -> A` while that transaction remains valid, the old bytes become authorized again because the wrapper previously bound the handoff only to the current key, not to a specific authority-state incarnation.

The exact-parent LiteSVM reproduction uses only public instructions and real signed transactions. After `A -> C -> A`, A contributes 50,000 tokens to asset-0 insurance. Replaying the retained `A -> B` transaction rekeys the live asset-0 insurance operator to B, and B withdraws the full 50,000-token contribution.

This is a BLOCKER LoF under `scripts/loop.md`: the victim's keys remain honest, while an unprivileged relayer resubmits previously honest signed bytes during their validity window.

## Root cause

`UpdateAuthority` required both current and incoming signatures but had no program-owned monotonic sequence. Returning to a prior public key recreated the complete authorization predicate for every still-valid transaction that key had signed.

## Fix

- Add `expected_sequence` to the tag-32 ABI.
- Persist a checked monotonic market-authority sequence in asset 0's existing wrapper reserve; account sizes and the engine pin are unchanged.
- Require exact sequence equality and increment on every successful handoff.
- Reject stale retained handoffs with `EngineStale`; SVM rollback leaves the market and vault byte-identical.
- Keep fresh handoffs live after a stale rejection.
- Document the one-shot handoff contract and prove the sequence transition exact/monotonic with Kani.

## TDD evidence

- RED commit: `3d6d7468` (`v16_attack_market_authority_handoff_cannot_replay_after_same_generation_aba`)
- Fix commit: `14d3b88d`
- Before the fix: retained transaction succeeds and B publicly withdraws 50,000 tokens.
- After the fix: retained transaction rejects specifically as `EngineStale`, market/vault bytes are unchanged, B cannot withdraw, and a fresh sequence-2 handoff succeeds.

## Verification

- `cargo kani --tests --harness kani_v16_marketauth_sequence_is_exact_and_monotonic`: 102 checks, 0 failures
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,8 +172,9 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
   any asset including asset 0** (`ASSET_ACTION_SHUTDOWN` → RECOVERY with the `force_close_delay_slots`
   exit window so traders can exit), **resolve/close the market** (`ResolveMarket`/`CloseSlab`),
   **market policies**, and **rotate/swap the base-unit mint**. It is rotated via
-  `UpdateAuthority { new_pubkey }` (current `marketauth` signs and the non-zero replacement co-signs;
-  burn-to-zero is rejected). Everything else — restart, insurance/operator/backing/oracle on
+  `UpdateAuthority { new_pubkey, expected_sequence }` (current `marketauth` signs, the non-zero
+  replacement co-signs, and the checked market sequence makes the handoff one-shot; burn-to-zero is
+  rejected). Everything else — restart, insurance/operator/backing/oracle on
   **every** asset including 0 — is per-asset (`asset_admin` + `UpdateAssetAuthority`), never a separate
   marketauth-only path. `marketauth` can restart an asset only while it is also that asset's
   `asset_admin` (the asset-0 bootstrap state).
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Independent current-pin confirmation from percolator-meta: https://github.com/aeyakovenko/percolator-meta/pull/381 (`b0809a1beebc42553a834b3dc18589d2d406500c`).

The clean-room LiteSVM probe system-allocates/publicly initializes the market and runs the pinned real SBF (`557f15b6`, SHA-256 `74ce6b85576fcbcb7596fdc7a774ff92661853d50ca9eb9475fcba7b1b7f40a4`). It separates insurance authority to an independent provider and burns the cold asset admin before `A -> C -> A`. The provider deposits 50,000 only after that honest cycle. Control remains `(attacker 0, insurance 50,000)`; replaying the retained market handoff produces `(attacker 50,000, insurance 0)` through public `WithdrawInsuranceAsset`. This independently confirms strict public-interface LoF against later provider funds.

aeyakovenko:

PR #346, opened 2026-07-25
[BLOCKER LoF] Make asset-authority handoffs one-shot

aeyakovenko:
## Impact

A displaced incoming per-asset authority can retain a fully signed `A -> B` handoff. If the live holder later rotates `A -> C -> A` in the same asset generation, the old bytes again satisfy the wrapper's complete authorization predicate.

The exact-parent LiteSVM reproduction removes the privileged explanation entirely:

1. Activate asset 1 with a distinct insurance provider P and insurance operator A.
2. Burn `asset_admin`, leaving no live admin for the asset.
3. Retain a holder-owned, A+B-signed operator handoff `A -> B`.
4. Land legitimate holder-owned handoffs `A -> C -> A`.
5. P publicly contributes 50,000 tokens to the asset's insurance domain.
6. Replay the retained transaction; B becomes operator and publicly withdraws all 50,000 tokens.

This is a BLOCKER LoF under `scripts/loop.md`: the independent provider and every current authority key remain honest; the attacker only resubmits previously valid signed bytes.

The same public probe also fails on PR #251 at `255e56ee`. Holder consent prevents an asset admin from seizing another holder's role, but it cannot distinguish an old handoff signed by the same holder after an `A -> C -> A` cycle. The fixes are complementary, not duplicates.

## Root cause

`UpdateAssetAuthority` authenticated current/incoming keys but did not bind their signatures to a specific handoff state. Public-key equality therefore recreated authorization after a same-key ABA cycle.

## Fix

- Add `expected_sequence` to the tag-65 ABI.
- Store five independent checked `u64` sequences per asset, one for each authority kind, in existing wrapper reserve bytes.
- Require exact equality and increment only on successful handoff.
- Keep role counters isolated so an oracle rotation cannot invalidate an insurance or backing handoff.
- Reject stale handoffs as `EngineStale`; SVM rollback leaves market and vault byte-identical.
- Preserve fresh handoff liveness after rejecting stale bytes.
- Keep account sizes and engine pin unchanged.

The reserve range is `profile +24..+64`, disjoint from the pending insurance-topup (`+8..+16`) and market-authority (`+16..+24`) counters.

## TDD evidence

- RED commit: `7c5f2748`
- Fix commit: `20da1dd3`
- Before: retained handoff succeeds and B withdraws P's full 50,000-token contribution.
- After: replay rejects specifically as `EngineStale`; operator sequence remains 2; unrelated insurance-authority sequence remains 0; B cannot withdraw; a fresh sequence-2 handoff succeeds and advances to 3.

## Verification

- Reproduced RED on exact parent and independently on PR #251 fixed head `255e56ee`
- `cargo check --all-targets`
- `cargo kani --tests --harness kani_v16_asset_authority_sequence_is_exact_and_monotonic`: 102 checks, 0 failures
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,11 +154,14 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
 - **Per-asset admin keys, isolated — uniform across all assets including asset 0** — one asset's admin
   can never be used against another asset. **✅** Every asset (0..N) carries its own `asset_admin`
   (`AssetOracleProfileV16`): assets 1..N bootstrap it to the activator, **asset 0 bootstraps it to the
-  market admin at `InitMarket`**. `UpdateAssetAuthority { asset_index, kind, new_pubkey }` is scoped to
-  that asset's profile only and now operates on **asset 0 too** (the old `asset_index == 0` rejection is
-  gone). Asset 0 is **not** special for authorities — its only special properties are **fee capture**
-  (it's the insurance-redirect target) and that it **cannot be permissionlessly created** (it's created
-  at `InitMarket`, not via `UpdateAssetLifecycle`).
+  market admin at `InitMarket`**.
+  `UpdateAssetAuthority { asset_index, kind, new_pubkey, expected_sequence }` is scoped to that
+  asset's profile only and now operates on **asset 0 too** (the old `asset_index == 0` rejection is
+  gone). Each authority kind has an independent program-owned sequence; an exact match is consumed
+  on success so a co-signed handoff is one-shot. Asset 0 is **not** special for authorities — its
+  only special properties are **fee capture** (it's the insurance-redirect target) and that it
+  **cannot be permissionlessly created** (it's created at `InitMarket`, not via
+  `UpdateAssetLifecycle`).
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Independent reduced current-pin verification is now in aeyakovenko/percolator-meta#379 at commit `4291fb5be6e177549fadaabf0ac4f2109c051cf6`.

The ignored LiteSVM RED runs against Meta's real pinned Percolator SBF (`557f15b63c7a5f9e058a2c384ca8728829dc2b63`, sha256 `74ce6b85576fcbcb7596fdc7a774ff92661853d50ca9eb9475fcba7b1b7f40a4`). It reduces the scenario to asset 0: public market init, independent provider/operator separation, admin burn, retained `A -> attacker`, legitimate `A -> C -> A`, then a later 50,000-atom independent insurance contribution.

Observed result:

- control: attacker `0`, insurance `50,000`;
- replay: attacker `50,000`, insurance `0`.

This confirms strict real LoF on the current Meta pin. No oracle, position, state injection, admin compromise, or private engine call is involved; the displaced attacker only relays old signed bytes and drains the entire later provider contribution.

aeyakovenko:

PR #347, opened 2026-07-25
[BLOCKER LoF] Reject delayed stale-resolution policy replay

aeyakovenko:
## Impact

An unprivileged relayer can retain an older, honestly signed short stale-resolution policy and land it after the market authority's later long-window correction. Once that short timer is already mature, every public crank rejects as oracle-stale while `ResolveStalePermissionless` succeeds, so an authenticated but unapplied mark is frozen out of terminal settlement.

The exact-parent public LiteSVM reproduction uses a normally initialized AuthMark market, two funded independent users, and no state injection:

1. The long victim and short attacker open 10,000 lots at 100 with 1,000,000 atoms each.
2. The honest market authority signs ordinary retry variants for a 2-slot policy; one lands.
3. The authority corrects the policy to 100 slots while the retained variant remains valid.
4. The honest oracle commits AuthMark 110.
5. At slot 3, the relayer lands the retained 2-slot policy.
6. An honest crank now rejects, but permissionless resolution succeeds at effective price 100.
7. After the normal exit delay, both users receive 1,000,000. Applying the already-authenticated mark first pays the victim 1,100,000 and the short 900,000.

The stale policy therefore leaves the attacker 100,000 atoms better off and the independent victim 100,000 atoms worse off relative to the accepted oracle target. This is `[BLOCKER LoF]` under `scripts/loop.md`: no admin key is compromised; the attacker only resubmits previously honest signed bytes.

This supplies the missing value-impact proof for the earlier closed #341. That probe showed early terminal shutdown but did not establish loss because both users could still withdraw. The new RED demonstrates stale-price value transfer, not fund stranding.

## Root cause

Tag 38 authenticated `marketauth` but did not bind a signed policy to an ordering watermark. Solana transaction deduplication distinguishes fee variants, so an old policy could become live after its correction.

## Fix

- Add a signer-chosen monotonic `sequence` to `ConfigurePermissionlessResolve`.
- Require nonzero strict advancement; delayed equal/lower intents return `EngineStale`.
- Store the watermark in asset 0's existing wrapper reserve at `+64..+72`.
- Keep this range disjoint from pending insurance top-up (`+8..+16`), market-authority (`+16..+24`), and per-asset-authority (`+24..+64`) counters.
- Reject legacy/truncated/trailing payloads under exact framing.
- Preserve account sizes, wrapper config layout, engine pin, and every user crank/exit path.

## TDD evidence

- Exact parent: `36fa587e1424a8c7fdde2c701c10938188a51876`
- RED commit: `cbaa2bf6`
- Fixed commit: `8747c2c9`
- Before: stale replay succeeds; honest crank rejects; stale resolution freezes 110 at 100; payouts are 1,000,000/1,000,000 instead of 1,100,000/900,000.
- After: replay returns `EngineStale`; market bytes remain identical; the sequence-2 watermark survives the AuthMark profile write; honest cranks apply 110; payouts are 1,100,000/900,000.

## Verification

- `cargo check --all-targets`
- `cargo test --all-targets -- --test-threads=1`: 6 unit + 566 public LiteSVM/CU tests passed
- production matcher `cargo test --all-targets`: 35/35 passed
- `cargo kani --tests --harness kani_v16_permissionless_resolve_decode_preserves_wire_fields`: 841 checks, 0 failures
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,6 +383,9 @@ This section describes intent and operational ordering, not argument-by-argument
   - setting `new_pubkey` to all zeros is rejected; `marketauth` must remain live for final slab reclaim
   - per-asset authorities (insurance/operator/backing/oracle, incl. asset 0) are rotated via
     `UpdateAssetAuthority`, not this instruction
+- **ConfigurePermissionlessResolve** (tag 38)
+  - `sequence` must strictly advance the stored market policy watermark; delayed older policies
+    cannot shorten a corrected stale timer and freeze an unapplied authenticated mark
 - **UpdateAssetLifecycle** (tag 40)
   - appends/reactivates/retires assets 1..N, including permissionless create/reuse when the configured
     create fee is nonzero
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -49,6 +49,7 @@ pub mod constants {
     pub const WRAPPER_CONFIG_LEN: usize = 448;
     pub const ASSET_ORACLE_PROFILE_LEN: usize = 400;
     pub const ASSET_ORACLE_WRAPPER_LEN: usize = 512;
+    pub const PERMISSIONLESS_RESOLVE_POLICY_SEQUENCE_OFF: usize = ASSET_ORACLE_PROFILE_LEN + 64;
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #348, opened 2026-07-25
[NOT REAL LoF][CUSTODY HARDENING] Bind backing withdrawals to asset generations

aeyakovenko:
## Root cause

`WithdrawBackingBucket` (tag 50) and `WithdrawBackingBucketEarnings` (tag 52) authenticated only the current `backing_bucket_authority` and the domain index; their signed payloads did not bind the withdrawal to the asset slot's `market_id`. A transaction retained from one asset incarnation could therefore debit a replacement incarnation after public retire/reactivate slot reuse — the exact slot-reuse drain #328 fixed for `WithdrawInsuranceAsset` (tag 57), on the backing domain.

## Public exploit

An independent provider funds a backing domain in one generation and the custodial `backing_bucket_authority` signs a withdrawal that is retained. The slot is publicly retired and reactivated; an **independent** provider funds the reused backing domain in the next generation under the re-appointed authority. Relaying the retained withdrawal within the recent-blockhash window drains the replacement provider's freshly funded backing to the old destination. No protocol-state injection, dishonest oracle, or compromised current key is involved; the transaction succeeds, so SVM rollback does not mitigate it.

## Fix

- Add the expected asset `market_id` to the tag-50 and tag-52 wire payloads.
- Reject a generation mismatch (`AssetGenerationMismatch`) at the read-only preflight and re-check at the mutation boundary (asset index = `domain / 2`), before the engine debit and token CPI.
- Reject the legacy unbound payloads (fail closed).
- Update all callers and the ABI.

## Verification

- **Measured RED** (guards removed, real BPF): the retained generation-2 `WithdrawBackingBucket` drains generation-3 backing — `a retained generation-2 backing withdrawal drained generation-3 backing` (stale destination receives the amount; replacement `fresh_unliened_backing_num == 0`).
- **GREEN**: the relayed stale withdrawal is rejected with `Custom(21)` (`AssetGenerationMismatch`); market + vault byte-identical; replacement backing intact. Positive control: a current-generation withdrawal still succeeds.
- `cargo test --test v16_cu`: **570 passed, 0 failed**.
- Kani wire-field-preservation, legacy-payload-rejection, and trailing-byte proofs added for both tags.

Intentionally stacked on #328 (`codex/probe-insurance-withdraw-generation-replay-20260724`), which introduces the persistent generation source and binds the sibling insurance withdrawal.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2576,6 +2576,7 @@ pub mod ix {
         },
         WithdrawBackingBucket {
             domain: u16,
+            market_id: u64,
             amount: u128,
         },
         ConvertReleasedPnl {
@@ -2622,6 +2623,7 @@ pub mod ix {
         },
         WithdrawBackingBucketEarnings {
             domain: u16,
+            market_id: u64,
             amount: u128,
         },
         SyncBackingDomainLedger {
@@ -2851,6 +2853,7 @@ pub mod ix {
                 },
                 50 => Self::WithdrawBackingBucket {
                     domain: read_u16(&mut rest)?,
+                    market_id: read_u64(&mut rest)?,
                     amount: read_u128(&mut rest)?,
                 },
                 28 => Self::ConvertReleasedPnl {
@@ -2918,6 +2921,7 @@ pub mod ix {
                 },
                 52 => Self::WithdrawBackingBucketEarnings {
                     domain: read_u16(&mut rest)?,
+                    market_id: read_u64(&mut rest)?,
                     amount: read_u128(&mut rest)?,
                 },
                 53 => Self::SyncBackingDomainLedger {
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Strict live-instance LoF/DoS re-audit: this is generation-binding hardening, but the published path does not meet the blocker gate.

Tag 50/52 preflight requires the signer to be the *current* backing authority, and the SPL destination must be owned by that same authority. A third-party relayer therefore cannot receive the replacement provider's backing. The exploit also requires that exact custodial key to be reappointed in the new generation after having signed the retained withdrawal; if that custodian is malicious it can authorize a current withdrawal directly, while an honest custodian receives the funds into its own account rather than losing them to the relayer. The path terminates and does not create persistent DoS.

For the Meta integration specifically, the canonical backing authority is a program PDA that signs CPI only during a live Subledger/controller call, so there is no externally retainable PDA-signed transaction variant across restart.

Closing as privileged custody/generation hardening, not unprivileged attacker-extractable LoF.

PR #349, opened 2026-07-25
[BLOCKER LoF] Reject delayed backing-fee policy replay

aeyakovenko:
## Impact

`UpdateBackingFeePolicy` had no ordering watermark. An unprivileged relayer can land a newer honest-authority correction, then land an older still-valid signed policy transaction and reinstate a superseded fee. The public LiteSVM repro uses a normally initialized live market, two independent traders, authenticated marks, permissionless cranks, and real provider backing; it does not inject account bytes or engine state.

On the exact parent (`36fa587e`), the stale transaction succeeds at 92,099 CU. The same source-backed risk increase then debits the independent trader by 750 atoms and credits provider earnings by 750 atoms, while the control schedule charges neither and both create the same lien.

## Fix

- Add `policy_sequence` to tag 51.
- Persist a 48-bit monotonic per-asset control-plane watermark in existing profile padding.
- Reject non-increasing intents as `EngineStale`.
- Preserve the watermark across all profile/lifecycle rewrites.
- Prove exact decoding plus trailing-byte rejection in the focused Kani harness.

The watermark is deliberately generic (`asset_intent_sequence`) so #335 can rebase and reuse it for authenticated-mark intent ordering instead of consuming the same padding independently.

## TDD

- RED: `72f4e7cf` on exact parent `36fa587e`.
- GREEN: `a6792054`.

## Verification

- `cargo test --all-targets -- --test-threads=1` (6/6 unit, 566/566 LiteSVM/CU)
- `cargo check --tests`
- `cargo kani --tests --harness kani_v16_update_backing_fee_policy_decode_preserves_wire_fields` (834 checks, 0 failures)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -568,7 +568,8 @@ pub mod state {
         pub backing_trade_fee_bps_short: u16,
         pub backing_trade_fee_insurance_share_bps_long: u16,
         pub backing_trade_fee_insurance_share_bps_short: u16,
-        pub _padding0: [u8; 6],
+        /// Monotonic watermark shared by signed control-plane intents for this asset.
+        pub(crate) asset_intent_sequence_le: [u8; 6],
         pub insurance_authority: [u8; 32],
         pub insurance_operator: [u8; 32],
         pub backing_bucket_authority: [u8; 32],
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #343, opened 2026-07-25
[BLOCKER LoF] Make trade intents one-shot across retry variants

aeyakovenko:
## Classification

`[BLOCKER LoF]`. An honest taker can sign ordinary priority-fee/blockhash retry variants for one economic trade. A malicious configured LP that pays fees and relays those variants can retain both valid transactions and land both, doubling the victim's exposure without another authorization. This matches the retained honest-signature replay rule in `scripts/loop.md`.

## Exact-parent public RED

Baseline: `36fa587e1424a8c7fdde2c701c10938188a51876`.

The isolated exact-parent RED is commit [`44afcf98`](https://github.com/aeyakovenko/percolator-prog/commit/44afcf9879a9a09eb7cde2f8ea44b3f558b3acfc). It uses only normal LiteSVM public instructions:

- honest taker and malicious configured LP deposit 4,000,000 atoms each;
- the taker signs two priority variants of one `TradeCpi` short intent;
- the LP is fee payer/relayer and lands both retained transactions;
- an honest AuthMark moves 100 to 110, followed by public cranks and a normal close;
- the victim withdraws 3,800,000 and the LP withdraws 4,200,000;
- the retained retry transfers an extra 100,000 atoms from the independent victim to the LP.

The branch also begins with a direct bilateral retry RED (`b27c224e`). No oracle dishonesty, admin compromise, protocol-state injection, or private engine call is used.

## Fix

All four public trade routes now bind signed one-shot intents:

- `TradeNoCpi` and `BatchTradeNoCpi` consume a nonzero portfolio-local intent for both signers.
- `TradeCpi` and `BatchTradeCpi` consume only the signed taker's intent and bind the unsigned LP's current portfolio incarnation without burning LP replay state.
- A 64-ID sliding bitmap permits distinct intents to land out of order while rejecting duplicates and expired IDs.
- Portfolio IDs bind retained bytes to an incarnation. Pre-ID portfolios migrate to the reserved legacy incarnation sentinel on their first trade; close/recreate receives a fresh monotonic ID.
- Intent mutation occurs after auth/provenance/currentness preflight. Any later engine, matcher, or margin error propagates and SVM rollback restores the window.

This supersedes the narrower trade-incarnation work in #276/#303. Asset/full-market generation binding remains the separate lifecycle issue in #231.

## Verification

- `cargo build-sbf --tools-version v1.52`
- `cargo test --test v16_cu -- --test-threads=1`: **575/575 passed**, including all max-shape CU gates
[...]

PR #350, opened 2026-07-26
[BLOCKER LoF] Make deposit intents one-shot across retry variants

aeyakovenko:
## Impact

An unprivileged relayer can retain multiple fee/blockhash variants of the same owner-signed `Deposit` intent and land more than one variant. If the duplicate lands after an adverse mark, it can force a second top-up into the now-insolvent portfolio and transfer that value to an independent winning trader during terminal settlement.

The public LiteSVM reproduction uses only normal instructions and real SPL accounts. A victim and independent winner open opposite positions with 1,000 collateral each. The victim signs two priority variants for one 500 deposit. After the intended variant lands, honest AuthMark updates move 100 -> 200 -> 300. Before this fix, the retained variant also succeeds at 49,193 CU. The victim source falls from 500 to 0 and the independent winner's terminal payout rises from 2,500 to 3,000.

## Root cause

`Deposit` was owner-signed but carried no portfolio incarnation or economic-intent identity. Solana transaction deduplication distinguishes fee/blockhash variants, so the program treated both as independent deposits.

## Fix

- Bind `Deposit` to the program-assigned `portfolio_id`.
- Consume a nonzero portfolio-local `intent_id` in a 64-ID replay bitmap.
- Permit distinct owner intents to land out of order while rejecting duplicate, zero, malformed, and too-old IDs.
- Grow legacy portfolio accounts and migrate pre-incarnation portfolios to the existing deterministic sentinel.
- Reject the legacy 17-byte deposit payload instead of silently accepting an unbound request.

The replay storage is intentionally generic owner-intent storage. PR #343 should rebase onto this change and reuse this window rather than allocating a second portfolio replay tail.

## TDD evidence

- `d80ae797` is the test-only RED commit on exact parent `36fa587e1424a8c7fdde2c701c10938188a51876`; it fails because the retained transaction succeeds and transfers the extra 500.
- `11ca66a7` makes the same test GREEN and asserts source, market, portfolio, and vault rollback on duplicate rejection.
- A positive public LiteSVM test lands two distinct deposit IDs in reverse order and withdraws the exact 300 deposited.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 8/8 library and 567/567 LiteSVM/CU tests pass.
- `cargo kani --tests --harness kani_v16_deposit_intent_decode_preserves_wire_fields`: pass.
- `cargo kani --tests --harness kani_v16_owner_intent_consumption_is_at_most_once`: 12/12 checks pass.
- `cargo kani --tests --harness kani_v16_custody_payloads_reject_trailing_byte`: 1,500/1,500 checks pass (8 unreachable).
- Current wrapper, auth matcher, hostile matcher, and external matcher SBF fixtures were rebuilt before the full run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,9 @@ This section describes intent and operational ordering, not argument-by-argument
   - initializes a program-owned portfolio account and binds `owner = signer`
 - **Deposit** (tag 3)
   - transfers collateral into vault; credits engine balance for that account
+  - binds the signed request to `portfolio_id` and consumes a nonzero portfolio-local `intent_id`;
+    a 64-ID replay window allows distinct deposits to land out of order but rejects fee-bump or
+    blockhash retry variants for an already consumed economic intent
 - **Withdraw** (tag 4)
   - performs oracle-read + engine checks; withdraws from vault via PDA signer; debits engine
 - **ClosePortfolio** (tag 8)
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -65,7 +65,11 @@ pub mod constants {
     pub const PORTFOLIO_MATCHER_CONFIG_LEN: usize = 104;
     pub const PORTFOLIO_ID_OFF: usize = PORTFOLIO_MATCHER_CONFIG_OFF + PORTFOLIO_MATCHER_CONFIG_LEN;
     pub const PORTFOLIO_ID_LEN: usize = 8;
-    pub const PORTFOLIO_ACCOUNT_LEN: usize = PORTFOLIO_ID_OFF + PORTFOLIO_ID_LEN;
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Current Meta pin independently confirms direct extraction on the real pinned SBF binary: https://github.com/aeyakovenko/percolator-meta/pull/376 (`c16fc18a422fe50430828e0e06fc3c6d80e14547`).

The clean-room LiteSVM path system-allocates and publicly initializes every Percolator market/portfolio account. The relayer is also the winning counterparty; after honest marks commit the victim's insolvency, current pin `557f15b63c7a5f9e058a2c384ca8728829dc2b63` produces control `(victim source=500, attacker=2500, victim=0)` versus replay `(victim source=0, attacker=3000, victim=0)`. No initialized protocol state is injected.

aeyakovenko:

PR #351, opened 2026-07-26
[BLOCKER LoF] Make backing top-ups one-shot across retry variants

aeyakovenko:
## What changed

- Add a nonzero `intent_id` to `TopUpBackingBucket`.
- Track a 64-entry replay window independently for each backing domain, allowing bounded out-of-order landing while rejecting duplicate retained variants.
- Reject initial or forward jumps greater than 64 so an old authority cannot poison the counter with `u64::MAX`.
- Add wire-format, truncation, replay-window, cross-domain isolation, and public LiteSVM regression coverage.

## Public exploit

An honest backing provider signs two fee-bump variants of the same 500-unit top-up using the same blockhash and instruction payload. The intended variant lands. After a normal public trade and an honest adverse mark move, an unprivileged relayer lands the retained variant before resolution.

On exact parent `36fa587e1424a8c7fdde2c701c10938188a51876`:

- control: provider source `500`, independent winner payout `2500`, loser payout `0`
- replay: provider source `0`, independent winner payout `3000`, loser payout `0`
- retained transaction succeeds at `34,844 CU`

The exploit uses only public instructions, a normally initialized AuthMark market, real SPL token accounts, and no protocol-state injection. The second debit becomes backing for an independent winner, so this is net extractable loss from the honest provider.

## Root cause

`TopUpBackingBucket` had no at-most-once identity. Solana fee-bump variants may share the same semantic instruction and recent blockhash but have distinct transaction signatures, so both variants could execute.

## TDD commits

- RED: `e2bd5e3e` reproduces the retained-variant loss on the exact parent.
- GREEN: `1037c355` rejects the retained duplicate atomically while preserving bounded out-of-order intents.

## Verification

- `cargo test --all-targets` (`8` library + `567` LiteSVM/CU tests green)
- focused RED/GREEN exploit and out-of-order/cross-domain LiteSVM tests
- Kani: replay-window at-most-once, backing-top-up wire preservation, custody trailing-byte rejection
- `cargo fmt --check`
- `git diff --check`

## Integration note

The new `intent_id` is appended after `expiry_slot`. The replay state occupies the final 32 bytes of each 512-byte asset wrapper as two independent 16-byte domain windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,6 +427,10 @@ This section describes intent and operational ordering, not argument-by-argument
   - validates side encoding and engine readiness; it is not an admin override
 - **TopUpInsurance** (tag 9)
   - transfers collateral into vault; credits insurance fund in engine
+- **TopUpBackingBucket** (tag 24)
+  - transfers collateral into an asset's backing bucket under that domain's backing authority
+  - carries a nonzero, per-domain `intent_id`; a 64-entry replay window permits bounded
+    out-of-order landing while rejecting retained transaction variants for an already-consumed intent
 
 ### Trading
 - **TradeNoCpi**
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -49,6 +49,9 @@ pub mod constants {
     pub const WRAPPER_CONFIG_LEN: usize = 448;
     pub const ASSET_ORACLE_PROFILE_LEN: usize = 400;
     pub const ASSET_ORACLE_WRAPPER_LEN: usize = 512;
+    pub const BACKING_TOPUP_REPLAY_WINDOW_LEN: usize = 16;
+    pub const ASSET_BACKING_TOPUP_REPLAY_OFF: usize =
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Current Meta pin independently confirms the strict extraction on the real pinned SBF binary: https://github.com/aeyakovenko/percolator-meta/pull/375 (`314692de3a39e139bcbdc4867e1257124a4fa9de`).

The clean-room LiteSVM path system-allocates and publicly initializes every Percolator market/portfolio account, uses a real SPL `MintTo` for the provider's replacement funds, and makes the relayer the winning portfolio owner. Current pin `557f15b63c7a5f9e058a2c384ca8728829dc2b63` produces control `(provider=500, attacker=2500, loser=0)` versus replay `(provider=0, attacker=3000, loser=0)`. No initialized protocol state is injected.

aeyakovenko:

PR #352, opened 2026-07-26
[NOT REAL LoF][SIGNED-INTENT HARDENING] Make backing withdrawals one-shot across retry variants

aeyakovenko:
## Impact

An unprivileged relayer can retain a fee-bump variant of an honest backing provider's already-landed `WithdrawBackingBucket` transaction. If the provider replenishes that bucket while the retained recent blockhash is still valid, the old variant can land again, remove the replacement backing, and reduce an independent trader's terminal payout after a legitimate adverse mark.

This needs no key compromise, oracle compromise, protocol-state injection, or malformed account.

## Exact-parent TDD

- Parent: `36fa587e1424a8c7fdde2c701c10938188a51876`
- RED: `0eeccc06`
- GREEN: `d2e370c7`
- Pre-fix retained withdrawal succeeds at 26,396 CU.
- Public LiteSVM differential: control `(provider_dest=500, winner=2500, loser=0)`; replay `(provider_dest=1000, winner=2000, loser=0)`.

## Fix

Tag 50 now carries an asset-level `intent_id`. A persistent 64-intent replay window accepts distinct withdrawals in bounded out-of-order order across both backing domains, but rejects duplicate, zero, malformed, too-old, and forward-poisoning IDs. The window is consumed only after authority/domain validation and rolls back with the instruction on every later error.

`read_next_backing_withdraw_intent_id` exposes the next client ID. Legacy tag-50 payloads reject instead of silently bypassing replay protection.

## Verification

- built `tests/fixtures/auth_matcher` and `tests/fixtures/hostile_matcher` with the same SBF tools
- `cargo test --all-targets -- --test-threads=1`: 8 unit + 567 LiteSVM/CU tests passed
- `cargo kani --tests --harness kani_v16_intent_replay_window_is_at_most_once`
- `cargo kani --tests --harness kani_v16_withdraw_backing_bucket_decode_preserves_wire_fields`
- `cargo kani --tests --harness kani_v16_custody_payloads_reject_trailing_byte`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,6 +459,9 @@ This section describes intent and operational ordering, not argument-by-argument
 - The per-slot effective-price movement cap is a risk parameter set at init; there is no standalone `SetOraclePriceCap` instruction in the current ABI.
 
 ### Insurance management
+- **WithdrawBackingBucket** (tag 50)
+  - provider withdrawal intents carry an asset-level `intent_id`; a 64-intent window accepts
+    bounded out-of-order landing across both domains while rejecting every intent at most once
 - **WithdrawInsurance** (tag 41)
   - unbounded resolved-market insurance withdrawal
   - gated by the asset-0 insurance authority, with shutdown-drain cases for nonzero assets gated by
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -49,6 +49,9 @@ pub mod constants {
     pub const WRAPPER_CONFIG_LEN: usize = 448;
     pub const ASSET_ORACLE_PROFILE_LEN: usize = 400;
     pub const ASSET_ORACLE_WRAPPER_LEN: usize = 512;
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Strict live-instance LoF/DoS re-audit: this reproduces a real signed-intent ABA, but it does not meet the project's blocker gate.

The retained transaction is signed by the canonical backing authority and sends only to that authority's chosen token destination. The public reproducer gives the broadcaster no value, gives the losing trader no value (`0` in both branches), and terminates normally; the only delta is that the provider receives its own replacement 500 atoms and the winner has 500 fewer atoms available. The same authority can authorize the same withdrawal directly after the trade, so the replay does not confer a new fund-moving capability. In the Meta controller path, external subledger depositors also do not hold this canonical Percolator signer.

This is useful transaction-intent hardening, but not an unprivileged attacker-extractable LoF and not a persistent DoS. Closing rather than carrying production replay-window surface as a blocker.

PR #353, opened 2026-07-26
[BLOCKER LoF] Bind terminal resolves to authority incarnations

aeyakovenko:
## Classification

`[BLOCKER LoF]`

Stacked on #345, which introduces the program-owned monotonic market-authority handoff sequence.

## Public exploit

An unprivileged relayer retains a valid `ResolveMarket` transaction signed by authority A. The
transaction initially fails while the market is not resolvable. After legitimate A -> C -> A
authority handoffs and an honest authenticated mark move from 100 to 110, the old bytes become
valid again because the wrapper checked only the current signer pubkey.

On exact parent `36fa587e1424a8c7fdde2c701c10938188a51876`, RED commit `4f8b1290`
demonstrates the retained transaction resolving at the stale 100 mark:

- victim payout: 900,000
- winner payout: 1,100,000
- independent victim loss transferred to the winner: 100,000 atoms

The reproduction uses only normal public instructions, initialized SPL custody, valid authority
handoffs, two independent traders, and an unprivileged relayer. It does not inject protocol state,
assume a compromised key, or rely on an initialization footgun.

## Fix

- Encode `ResolveMarket { expected_sequence }`.
- Require that sequence to equal the program-owned market-authority handoff sequence before
  resolution.
- Reuse the existing handoff counter from #345; do not add another mirror counter.
- Prove exact sequence-gate behavior and wire-field/trailing-byte decoding with focused Kani
  harnesses.

The sequence is an authority-incarnation identifier. It advances only on a successful authority
handoff, so retained bytes from an earlier A incarnation remain stale after A returns.

## Verification

- `cargo check --all-targets`
- `cargo test --all-targets -- --test-threads=1`
  - 6/6 library tests
  - 567/567 LiteSVM/CU tests
- `cargo kani --tests --harness kani_v16_resolve_market_decode_preserves_authority_sequence`
- `cargo kani --tests --harness kani_v16_marketauth_sequence_gate_is_exact`

Both Kani harnesses verified with zero failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,8 @@ Assets 1..N are **truly permissionless ⇒ untrusted**. The protocol must guaran
   `base_unit_authority`). `marketauth` is the only key that can: **create market 0** (`InitMarket`),
   **create/retire assets 1..N and set the permissionless-create-fee policy**, **safely force-shutdown
   any asset including asset 0** (`ASSET_ACTION_SHUTDOWN` → RECOVERY with the `force_close_delay_slots`
-  exit window so traders can exit), **resolve/close the market** (`ResolveMarket`/`CloseSlab`),
+  exit window so traders can exit), **resolve/close the market**
+  (`ResolveMarket { expected_sequence }`/`CloseSlab`),
   **market policies**, and **rotate/swap the base-unit mint**. It is rotated via
   `UpdateAuthority { new_pubkey, expected_sequence }` (current `marketauth` signs, the non-zero
   replacement co-signs, and the checked market sequence makes the handoff one-shot; burn-to-zero is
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Independent current-pin confirmation from percolator-meta: https://github.com/aeyakovenko/percolator-meta/pull/382 (`9e0b5be117c65fbbd66fcdafadc5329ee63e46aa`).

The clean-room LiteSVM proof system-allocates/publicly initializes every Percolator state account and runs the pinned real SBF (`557f15b6`, SHA-256 `74ce6b85576fcbcb7596fdc7a774ff92661853d50ca9eb9475fcba7b1b7f40a4`). The unprivileged relayer is also the long-side owner; an independent victim owns the short. After honest `A -> C -> A`, public deposits/trade, and an honest `100 -> 110` authenticated mark, the retained A-signed resolve succeeds. Public resolved close-outs pay the victim 900,000 and attacker/relayer 1,100,000, confirming a 100,000 victim-funded attacker gain rather than terminal-state griefing alone.

aeyakovenko:

PR #354, opened 2026-07-26
[BLOCKER DoS] Bind recovery forfeits to position episodes

aeyakovenko:
## What changed

- Bind `ForfeitRecoveryLeg` consent to the program-assigned `portfolio_id` and the wrapper-owned `position_epoch`.
- Reject stale provenance before dispatching the recovery mutation into the engine.
- Keep successful forfeits on the existing checked epoch-advance path introduced by #316.
- Reject legacy unbound and trailing-byte payloads with Kani coverage.

## Impact and root cause

An owner can pre-sign fee variants of one legitimate recovery-leg forfeit. After one variant lands, the retained variant remains valid for the blockhash lifetime. Before this change, the wrapper authenticated only the owner and asset, so the retained transaction could land during a later position/recovery episode on the same live portfolio.

The public LiteSVM reproduction creates two ordinary bankruptcy episodes without account reinitialization or injected protocol state. Replaying first-episode consent in episode two deletes a fresh 20,000-atom backed gain. Resolution then pays the victim only 30,000 of the expected 50,000 atoms and leaves 20,000 atoms permanently unowned in the canonical vault; all portfolios close, but `CloseSlab` has no successful path. This is classified as blocker DoS rather than LoF because the value is permanently stranded rather than extracted by another user.

This PR is intentionally stacked on #316 so it reuses the same position-episode primitive instead of adding a duplicate counter.

## TDD

- Exact-main RED: `6a6d0dce` (`v16_attack_presigned_forfeit_does_not_replay_across_recovery_episodes` fails after proving the 20,000-atom terminal residual).
- Stacked RED: `5296d91a`.
- GREEN fix: `fd4cf323`.

## Verification

- built `tests/fixtures/auth_matcher` and `tests/fixtures/hostile_matcher` with platform-tools v1.52
- `cargo test --all-targets -- --test-threads=1`: 7 unit + 568 LiteSVM/CU tests passed
- focused recovery-forfeit suite: 4/4 passed
- Kani wire-field preservation: passed
- Kani legacy unbound payload rejection: passed
- Kani trailing-byte rejection: passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,8 @@ This section describes intent and operational ordering, not argument-by-argument
 - **CureAndCancelClose** (tag 42)
   - owner-signed close recovery path; optional deposit is transferred first, then the engine cancels the pending close if the cure succeeds
 - **ForfeitRecoveryLeg** (tag 43)
-  - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget
+  - owner-signed recovery-leg forfeit for a selected asset and bounded B-delta budget; the payload
+    binds consent to both the program-assigned `portfolio_id` and the portfolio's `position_epoch`
 - **RebalanceReduce** (tag 44)
   - owner-signed risk-reducing rebalance against the wrapper-authenticated effective price vector;
     the payload binds consent to both the program-assigned `portfolio_id` and the portfolio's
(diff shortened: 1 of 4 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-004 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #355, opened 2026-07-26
[BLOCKER LoF] Make withdrawals one-shot across retry variants

aeyakovenko:
## What changed

- Add `portfolio_id` and a nonzero economic `intent_id` to the public `Withdraw` payload.
- Consume withdrawals through the generic portfolio owner-intent sliding replay window introduced by #350.
- Preserve out-of-order execution for distinct signed withdrawals within the 64-intent window.
- Reject the legacy amount-only withdrawal wire shape and add wrapper-specific Kani coverage.

## Root cause and impact

`Withdraw` was owner-authorized but replayable across transaction retry variants. An owner could sign priority-fee variants of one intended withdrawal. After one variant landed, a relayer could retain another, wait for the owner to sign a fresh trade, and land the stale withdrawal immediately before that trade.

The exact-main public LiteSVM repro performs one intended 50,000,000-atom withdrawal, lands the retained duplicate while the victim is flat, then lands the already-signed fresh position. Honest AuthMark updates within the configured 20 bps/slot envelope make only the replayed state liquidatable. An independent permissionless cranker extracts and withdraws a 7,917-atom victim-funded liquidation reward; the single-withdraw counterfactual remains maintenance-healthy.

This is not state injection, an init footgun, or a dishonest-oracle scenario. The stale request is a retained honest signature, all state transitions use public wrapper instructions, and the value loss is independently extractable.

## TDD evidence

- Exact `origin/main` RED: `d8ffdb6a`
- Stacked RED: `01d23ed8`
- GREEN fix: `29444960`

The fix is stacked on #350 to reuse one generic replay primitive instead of introducing a second counter or replay tail. It supersedes the narrower incarnation-only withdrawal protection in #299.

## Verification

- rebuilt `auth_matcher` and `hostile_matcher` SBF fixtures
- `cargo test --all-targets -- --test-threads=1`: 8 unit + 569 LiteSVM/CU tests green
- 59 withdrawal-focused LiteSVM tests green
- targeted Kani: withdrawal wire/legacy rejection, at-most-once owner-intent consumption, and custody trailing-byte rejection; 3/3 green

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,9 @@ This section describes intent and operational ordering, not argument-by-argument
     blockhash retry variants for an already consumed economic intent
 - **Withdraw** (tag 4)
   - performs oracle-read + engine checks; withdraws from vault via PDA signer; debits engine
+  - binds the signed request to `portfolio_id` and consumes a nonzero portfolio-local `intent_id`;
+    distinct owner intents may land out of order within the same 64-ID replay window used by
+    deposits, while retry variants for an already consumed economic intent are rejected
 - **ClosePortfolio** (tag 8)
   - closes a flat, empty portfolio account and sweeps its rent to the market slab
 - **CloseResolved** (tag 30)
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2586,6 +2586,8 @@ pub mod ix {
             amount: u128,
         },
         Withdraw {
+            portfolio_id: u64,
+            intent_id: u64,
             amount: u128,
         },
         PermissionlessCrank {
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Independent current-pin confirmation from percolator-meta: https://github.com/aeyakovenko/percolator-meta/pull/380 (`83847f402ad0610f6596000de40ed46e5cbe4caa`).

The clean-room LiteSVM probe system-allocates/publicly initializes every Percolator state account, runs the pinned real SBF (`557f15b6`, SHA-256 `74ce6b85576fcbcb7596fdc7a774ff92661853d50ca9eb9475fcba7b1b7f40a4`), and gives the attacker only fee-payer/relayer plus permissionless cranker roles. One withdrawal leaves the later fresh victim-signed position healthy: `(victim destination 50,000,000, attacker reward 0)`. Landing a retained priority variant of that same withdrawal gives `(100,000,000, 2,020)` and the attacker publicly withdraws the victim-funded liquidation reward. This confirms strict public-interface LoF rather than only duplicate delivery to the victim.

aeyakovenko:

PR #356, opened 2026-07-27
[BLOCKER LoF] Block fee rewards ahead of committed marks

aeyakovenko:
## Finding

An unprivileged caller can front-run an already-authenticated adverse mark with the permissionless `SyncMaintenanceFee` route. The wrapper has committed the honest target, but before a public crank applies it, the engine still sees the old effective price and lets the caller collect a larger cranker reward.

Exact public LiteSVM result on parent `36fa587e`:

- mark first: cranker reward 50,000; victim payout 0; independent winner payout 50,000
- fee sync first: cranker reward 100,000; victim payout 0; independent winner payout 0

The attacker signs no privileged instruction and the test uses no state injection.

## Fix

- Pin paired engine fix [percolator#159](https://github.com/aeyakovenko/percolator/pull/159), which rejects live non-flat fee sync while an engine target is ahead of effective state.
- At the wrapper boundary, scan only the target portfolio's bounded active-leg set and reject fee sync while a price-managed wrapper profile or engine raw target is ahead of effective state.
- A permissionless crank remains the successful bounded continuation. Once it applies the committed mark, fee sync retries normally; trading and exits are not blocked.

## TDD

The public LiteSVM regression proves:

- exact pre-fix RED with the divergent economics above;
- early sync rejects as `EngineLockActive`;
- SVM rollback leaves market, victim, and reward portfolio byte-identical;
- public crank applies the mark;
- retry succeeds;
- both orderings terminate at the same `(50,000, 0, 50,000)` outcome.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 566/566 LiteSVM/CU tests passed, plus 6/6 unit tests
- Existing max-shape crank/trade/liquidation CU tests passed
- `git diff --check`

Wrapper Kani covers ABI/local validation only; the paired engine PR carries the exact target-lag kernel contract and Kani proof rather than duplicating engine accounting in this repository.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=e09ecb2f46ead56e9304e856fc58980ddabe43bf#e09ecb2f46ead56e9304e856fc58980ddabe43bf"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #357, opened 2026-07-27
[BLOCKER DoS] Bound max-source PnL conversion

aeyakovenko:
## Finding

A public, healthy-market path can leave an independent LP with all 32 source-claim domains occupied. The LP authorizes a matcher once; afterward an unrelated taker alone signs ordinary CPI fills while authenticated mark moves create source-backed PnL on both sides of 16 assets.

On the exact `main` parent (`percolator-prog` `36fa587e`, engine `143e68c`), the LP is flat with 32 occupied source domains and 32,000 atoms of backed PnL. `ConvertReleasedPnl` scans and consumes the entire table in one instruction, uses exactly 1,400,000 CU, returns `ProgramFailedToComplete`, and rolls back. Repeating the public instruction cannot reduce the rank, so the LP cannot convert or withdraw those funds while the live market remains in that state.

This is not state injection: the regression uses only public LiteSVM instructions, a real matcher program, authenticated mark pushes, and ordinary signed transactions. The affected LP does not sign the individual fills.

## Fix

- Pin engine PR aeyakovenko/percolator#160, stacked on domain-local burn PR aeyakovenko/percolator#140.
- Route `ConvertReleasedPnl` through the engine's one-source-domain step API.
- Preserve `amount` as a postcondition cap under SVM rollback.
- Permit repeated flat-account steps without an otherwise-unrefreshable health certificate, while retaining H-locks and the full active-account freshness/target-lag gate.

## TDD result

- RED on exact parent: 1,400,000 CU, `ProgramFailedToComplete`, no progress.
- GREEN: all 32 source domains decrease one per call; worst step is 519,091 CU; PnL reaches zero; exact backed capital is withdrawn.
- Existing caller-cap regression remains green.

## Verification

- Engine normal all-target tests: green (50 unit, 9 backing, 2 resolved, 70 V16 spec).
- Engine fuzz/conformance all-target tests: green (including 15 reference, 9 arithmetic/residue, 74 V16 spec).
- Engine focused Kani: 2/2 harnesses green, all covers satisfied.
- Wrapper `cargo build-sbf --tools-version v1.52`: green.
- Wrapper full suite: 6/6 unit and 566/566 LiteSVM/CU tests green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=ce5e923c48d35989785427975a82ce74b6b19c72#ce5e923c48d35989785427975a82ce74b6b19c72"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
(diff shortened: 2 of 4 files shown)


aeyakovenko:

PR #358, opened 2026-07-27
[PRIVILEGED LoF] Commit authenticated marks before asset shutdown

aeyakovenko:
## Finding

A compromised non-oracle market/asset administrator could call the public asset-shutdown instruction after an independent honest oracle had authenticated an adverse mark but before a permissionless crank committed it. Shutdown froze the previous effective price and overwrote the pending target. Subsequent public recovery force-close and resolution therefore settled at the stale mark, transferring value from the independent winner to the privileged counterparty.

The public LiteSVM repro uses separate oracle and admin keys. With a 100 -> 200 authenticated target and a 5% per-slot circuit breaker, the vulnerable path froze at 100 and paid victim/attacker 10,000/10,000, while crank-before-shutdown froze at 105 and paid 10,050/9,950.

## Fix

Before transitioning an active or drain-only asset to shutdown, derive the circuit-breaker-bounded price from the already authenticated stored target and accrue the asset to that price. Reject a mark-bearing shutdown that is too stale for one bounded accrual segment, leaving the permissionless crank path available. Manual-price shutdown behavior is unchanged.

## TDD verification

- Exact-parent RED: `v16_attack_shutdown_cannot_discard_pending_authenticated_mark` observed 10,000/10,000 on shutdown-before-crank versus 10,050/9,950 in the honest control.
- Fixed GREEN: the same public sequence freezes at 105 and both paths pay 10,050/9,950.
- `cargo test --all-targets -- --test-threads=1`: 6 unit + 566 integration/BPF/CU tests passed.
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9291,6 +9291,50 @@ pub mod processor {
                 }
                 match group.markets[asset_index].engine.asset.lifecycle {
                     ASSET_LIFECYCLE_ACTIVE | ASSET_LIFECYCLE_DRAIN_ONLY => {
+                        let asset = group.markets[asset_index].engine.asset;
+                        let current_price = asset.effective_price.get();
+                        let committed_target = if oracle_v16::profile_is_price_managed(&profile) {
+                            profile.mark_ewma_e6
+                        } else {
+                            current_price
+                        };
+                        if committed_target == 0 {
+                            return Err(PercolatorError::OracleInvalid.into());
+                        }
+                        let exposed =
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Confirmed duplicate root cause and production fix scope with #254. #254 already commits the circuit-breaker-bounded authenticated target before shutdown and additionally preserves zero-price-move funding accrual; this PR is a narrower version of that fix. I am moving the new independent mark-PnL public LiteSVM regression onto #254 instead of keeping a second fix PR.

PR #359, opened 2026-07-27
[BLOCKER DoS] Keep max-shape resolved close under CU limit

aeyakovenko:
## What changed

- pin the wrapper to engine PR aeyakovenko/percolator#161
- add a public LiteSVM TDD that reaches 14 active legs and all 32 source domains through authenticated marks, public CPI trades, and permissionless cranks
- resolve through `ResolveStalePermissionless`, wait out the exit window, and require every `CloseResolved` call to strictly reduce leg/source rank or finish
- adapt the existing two-leg deep-insolvency test to the engine's documented `ProgressOnly` behavior

## Root cause

Resolved close detached every active leg and realized every source domain in one transaction. At the public maximum shape, the first terminal close exhausted the SVM meter. Rollback restored the same state, while resolved mode disabled trade and withdrawal routes, so retries could never make progress.

## Exact-parent RED

Against parent PR #357 at engine `ce5e923c48d35989785427975a82ce74b6b19c72`, the unchanged public test fails on its first post-timeout close:

- `ProgramFailedToComplete`
- `1,400,000` CU consumed
- program log: `exceeded CUs meter at BPF instruction`

No account bytes are injected or mutated by the test.

## Fixed GREEN

With engine `e221a6066986a2103aab1228d1cfd63d90561705`:

- both portfolios terminate in 59 permissionless close calls
- maximum `CloseResolved`: `764,474` CU
- all `4,000,000` custody atoms are paid
- final engine `vault == 0` and `c_tot == 0`

## Validation

- `cargo test --all-targets -- --test-threads=1`: 6 unit + 567 LiteSVM/BPF/CU tests passed
- engine `cargo test --all-targets`: 132 tests passed

Stacked on wrapper PR #357 and engine PR #161, which itself is stacked on engine PR #160.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=ce5e923c48d35989785427975a82ce74b6b19c72#ce5e923c48d35989785427975a82ce74b6b19c72"
+source = "git+https://github.com/aeyakovenko/percolator?rev=e221a6066986a2103aab1228d1cfd63d90561705#e221a6066986a2103aab1228d1cfd63d90561705"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #360, opened 2026-07-27
[BLOCKER LoF] Reject unsigned stale-cohort novation

aeyakovenko:
## Finding

An unsigned matcher LP could accept fresh exposure after the opposite-side K/F
cohort became stale. A winning account could therefore realize source-backed PnL
first, novate its position onto a fresh LP, and leave the later bankruptcy/social
loss booking on that LP.

The LiteSVM reproduction uses only public instructions and ordinary initialized
accounts:

1. Open an attacker-controlled winning long against a thin losing short.
2. Move the authenticated mark and crank only the winner.
3. Attempt to move fresh long exposure onto an unsigned matcher LP through both
   `TradeCpi` and `BatchTradeCpi`.
4. Crank the loser and all remaining work to terminal state.

At the exact pre-fix parent, the batch matcher route succeeds and the regression
test fails. With this patch, both unsigned CPI routes reject before invoking the
matcher, state and matcher context roll back exactly, signed `TradeNoCpi`
novation remains available, pure LP reductions remain live, and honest cranks
reach the expected terminal payouts.

## Fix

- Pin engine PR aeyakovenko/percolator#162, which tracks monotonic per-side K/F
  settlement cohorts and conservatively locks stale source-backed claims.
- At the unsigned matcher boundary, reject only LP exposure increases that can
  inherit a stale opposite-side cohort or unresolved negative PnL.
- Subtract both atomically settled trade accounts from each pending aggregate.
  A locally curable negative-PnL account can therefore still exit through CPI;
  only an external pending account blocks new unsigned LP exposure.
- Scan the LP portfolio once for the whole matcher batch. The 14-leg CPI test is
  1,238,049 CU versus 1,222,481 CU at the parent.
- Prove the side mapping and the reduction-liveness exception with Kani.

## TDD evidence

- Test-only commit: `2890daf337f8d7d71706d06fc579b5f21f4faadd`
- Exact old engine pin: `143e68c4917ed0400a27b952f036a5677047cd84`
- Pre-fix result: RED at the public `BatchTradeCpi` assertion
- Fixed engine pin: `d81b86252c028d7977c14bcf75c06eb199a61b2b`
- Fixed result: GREEN for single CPI, batch CPI, signed escape, rollback,
  progress, reset, and terminal-value assertions
- Local-negative liveness test-only commit: `d26c970a`
- Exact pre-fix policy result: RED with `EngineLockActive` before matcher CPI
  after 22,200 CU
- Local-negative fixed commit: `6acc341c`
- Local-negative fixed result: GREEN; inline settlement clears the local
  negative PnL and the trade closes both positions

## Verification

- Targeted LiteSVM exploit: pass
- Full runtime suite: 6 unit tests and 567 LiteSVM/BPF/CU tests passed
- Targeted 14-leg CPI CU test: pass at 1,238,049 CU
- Targeted wrapper Kani proof: 35/35 assertions and 7/7 covers
- Full wrapper Kani roster: 41/41 harnesses verified, 0 failures
- Checked decoder proof bounds: the symbolic vector cap is 16; every added
  unwind assertion remains enabled, and the 3-by-32 Hybrid feed comparison is
  explicit rather than hidden behind a 96-byte `memcmp`

## Deployment compatibility

Engine commit `d81b8625` advances `V16_LAYOUT_DISCRIMINATOR` from 17 to 18 and
adds settlement-cohort fields. This branch is suitable for fresh/predeployment
markets; an existing live layout-17 market requires a coordinated migration
before deployment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=d81b86252c028d7977c14bcf75c06eb199a61b2b#d81b86252c028d7977c14bcf75c06eb199a61b2b"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "d81b86252c028d7977c14bcf75c06eb199a61b2b" }
 anchor-lang-v2 = { git = "https://github.com/solana-foundation/anchor", rev = "e3cf760826fa0a60f247635ea0572e59861ec9b5", package = "anchor-lang-v2", default-features = false, features = ["alloc", "guardrails"], optional = true }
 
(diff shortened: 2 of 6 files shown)


aeyakovenko:

PR #361, opened 2026-07-27
[BLOCKER DoS] Reject retained backing top-ups after signed expiry

aeyakovenko:
## Root cause

`TopUpBackingBucket` treated `expiry_slot` as fresh relative only to the engine's stored `header.current_slot`. A relayer could retain an honestly signed provider transaction and land it after the signed expiry while no market instruction had advanced that stored slot. The wrapper did not compare the expiry with the authenticated SVM `Clock`.

## Public exploit

The LiteSVM regression uses only public instructions and ordinary external account fixtures:

1. Initialize a live market and permissionless asset with an independent backing provider.
2. Open public long/short positions and move the honest authenticated mark from 100 to 350.
3. At slot 3, have the provider sign a 500-atom domain top-up expiring at slot 6.
4. Let an unprivileged relayer retain it and submit it at authenticated slot 7, within the blockhash lifetime.
5. Resolve permissionlessly and repeatedly submit owner-signed `CloseResolved` calls.

On exact parent `36fa587e1424a8c7fdde2c701c10938188a51876`, test-only commit `fcb4af76` is RED:

- control: top-up rejected/not submitted, 16 successful close steps, both users terminal, winner payout 2,000, provider source 500
- delayed submission: top-up accepted, provider source 500 -> 0, 64 successful and 64 failed close attempts, both users remain nonterminal, both payouts 0, and the provider cannot withdraw the resolved backing

This is a persistent public terminal-progress DoS, not an init footgun or injected protocol state.

## Fix

For every nonzero backing top-up, both preflight and commit now require:

```text
expiry_slot > authenticated SVM Clock slot
```

The existing zero-amount expired no-op behavior is preserved. No engine code or engine pin changes are included.

## Verification

- Exact-parent RED repro: `fcb4af76`
- Fixed GREEN: `v16_attack_retained_backing_topup_cannot_land_after_signed_expiry`
- Adjacent backing input/expiry tests: green
- `cargo build-sbf --tools-version v1.52`: green
- Full runtime suite with all program/matcher fixtures: 6 unit + 566 LiteSVM/BPF/CU tests, 0 failures
- Kani: all 40 harnesses verified; 36 completed in the full sequence, then the pathological symbolic-tag decoder and its three successors completed explicitly with `--unwind 18` and unwinding assertions enabled (the decoder observation loop is production-bounded at 16)
- `git diff --check`: clean

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -7586,6 +7586,11 @@ pub mod processor {
             {
                 return Err(PercolatorError::EngineLockActive.into());
             }
+            if amount != 0
+                && expiry_slot <= authenticated_slot_or_fallback(group.header.current_slot.get())
+            {
+                return Err(PercolatorError::InvalidInstruction.into());
+            }
             require_domain_accepts_live_topup_view(&group, domain_usize)?;
             let profile = read_oracle_profile_from_view(&group, &cfg, asset_index)?;
             let authorities = domain_authorities_from_profile(&cfg, &profile, asset_index);
@@ -7603,6 +7608,9 @@ pub mod processor {
             let (cfg, mut group) = state::market_view_mut(&mut market_data)?;
             if group.header.mode != 0 {
                 return Err(PercolatorError::EngineLockActive.into());
+            }
+            if expiry_slot <= authenticated_slot_or_fallback(group.header.current_slot.get()) {
+                return Err(PercolatorError::InvalidInstruction.into());
             }
             reject_permissionless_resolve_matured_live_view(&cfg, &group)?;
             require_domain_accepts_live_topup_view(&group, domain_usize)?;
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Stronger independent current-pin consequence from percolator-meta: https://github.com/aeyakovenko/percolator-meta/pull/384 (`3816ca313c987837bb9dd896b8028c1d581fe2e4`).

A reduced clean-room asset-0 LiteSVM path shows strict LoF, not only terminal DoS. Every Percolator state account is system-allocated/publicly initialized and the pinned real SBF runs. An independent provider signs a 500-atom top-up at engine slot 3 with expiry 6. The existing profitable account is the unprivileged fee payer/relayer and submits it at authenticated SVM slot 7 while stored engine slot remains 3. Current pin accepts it. Honest resolution and owner-signed closes terminate both portfolios: control leaves provider source 500 and pays winner 2,000; delayed submission leaves provider source/recovery 0 and pays winner/relayer 2,500. This confirms exact 500 provider loss / attacker gain through the same missing Clock check.

aeyakovenko:

PR #362, opened 2026-07-28
[BLOCKER LoF] Make retired-slot activation intents one-shot

aeyakovenko:
## What changed

Bind every retired-slot `ACTIVATE` intent to the lifecycle retirement that made the slot reusable. The signed intent slot must be nonzero, no later than the authenticated Clock slot, and strictly later than the asset's recorded retirement slot. The guard runs before fee transfer or state mutation for both permissionless reuse and privileged/marketauth reactivation.

This is wire-compatible and does not change persisted layouts.

## Public exploit

An honest creator could sign two fee-bump transaction variants carrying the same permissionless activation intent. After one variant landed and paid the activation fee, the asset could be retired normally. A relayer could then land the retained variant during the recent-blockhash window, charging the creator a second time and installing stale authority/price material into a new asset generation.

The LiteSVM regression uses only public instructions, real SPL token accounts, normal lifecycle transitions, and retained signed transaction bytes. On exact parent `36fa587e1424a8c7fdde2c701c10938188a51876`, RED commit `3ec400f7bc690a7949976ce37409b58d2f9b9c43` reproduces the second debit. Fix commit `21a7f09f` rejects it atomically while preserving delayed/out-of-order landing within the intended retired generation.

This is distinct from mutable fee consent (#314) and shutdown generation binding (#315): it prevents replay of the same valid activation consent across successive asset generations.

## Verification

- `cargo test --all-targets -- --test-threads=1`: 6 unit + 566 LiteSVM/CU tests pass
- Focused public replay regression passes
- All 41 Kani harnesses pass with unwinding checks enabled: 36 at checked bound 17, 4 aggregate decode harnesses at 34, and the 3x32-byte hybrid decode harness at 98
- Program and all matcher SBF fixtures build
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4334,6 +4334,16 @@ pub mod processor {
         Clock::get().map(|c| c.slot).unwrap_or(fallback_slot)
     }
 
+    #[doc(hidden)]
+    #[inline]
+    pub fn activation_intent_slot_is_fresh(
+        intent_slot: u64,
+        authenticated_slot: u64,
+        retired_slot: u64,
+    ) -> bool {
+        intent_slot != 0 && intent_slot <= authenticated_slot && intent_slot > retired_slot
+    }
+
     fn authenticated_market_slot_or_fallback_view(group: &state::MarketViewMutV16<'_>) -> u64 {
         core::cmp::max(
             Clock::get()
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Clean-room validation against current Meta pins confirms this is real extractable LoF, not only a forced second protocol donation.

Meta draft PR: https://github.com/aeyakovenko/percolator-meta/pull/394
Meta commit: https://github.com/aeyakovenko/percolator-meta/commit/46b4c4f43b05d36e9fb5e93b162920bc18e0c3ee

The real-SBF LiteSVM differential carries the duplicate 10-atom init fee through independent asset-0 trading, authenticated oracle movement, public cranks, permissionless stale resolution, and SPL `CloseResolved` payout:

```text
control: creator source 10, domains [15,15], attacker payout 1214, insurance 15
retry:   creator source  0, domains [20,20], attacker payout 1219, insurance 20
```

Both paths conserve `source + attacker payout + remaining insurance = 1239`. The retained variant therefore redirects five creator atoms into attacker-withdrawable collateral and five into protocol insurance. No initialized Percolator state is injected; the normal market authority performs the same empty-slot retirement in both paths.

Pinned wrapper `557f15b63c7a5f9e058a2c384ca8728829dc2b63`, engine `e3ea7251ed75f0111ba3cf5a4b87961f6c7cca63`, SBF SHA-256 `74ce6b85576fcbcb7596fdc7a774ff92661853d50ca9eb9475fcba7b1b7f40a4`.

aeyakovenko:

PR #214, opened 2026-07-11
[BLOCKER DoS] Reconcile expired source liens

aeyakovenko:
## What changed

- Pin the wrapper to engine commit `765c1d893bae44d588db196549096cbc555e00a9` from engine PR #112.
- Add a public LiteSVM lifecycle regression for a source-credit IM lien whose backing bucket expires after creation.
- Add a max-shape public regression that creates all 28 source-domain liens, proves every successful crank removes exactly one lien, then proves recertification and owner risk reduction still work.

## Root cause

A normal risk-increasing trade could create a valid counterparty-backed source lien while its bucket was fresh. At expiry, unchanged-price auto-crank treated the account as `NoAction`; owner reduction, bilateral reduction, PnL conversion, and adverse-settlement routes read the lapsed source domain and returned engine `Stale`. Depositing more capital did not reconcile the lien.

The engine fix classifies expiry as actionable and reconciles one expired source-domain lien per public auto-crank. Each call is bounded and rank-decreasing; after the finite source-domain tail, ordinary certification and owner reduction resume.

Engine fix: https://github.com/aeyakovenko/percolator/pull/112

## Public evidence

The max-shape regression uses only wrapper instructions and normal trades:

- 14 configured assets and both source domains per asset
- 28 source claims created by public mark/trade cycles
- 28 counterparty liens allocated by public risk-increasing trades
- exactly 28 successful expiry cranks, each removing one live lien
- peak expiry step: 667,401 CU
- successful post-expiry recertification and owner `RebalanceReduce`
- no direct account-data mutation or injected engine state

## Validation

- `cargo build-sbf --tools-version v1.52`
- rebuilt auth and hostile matcher SBF fixtures
- `cargo test --all-targets -- --quiet`: 5/5 library and 572/572 LiteSVM tests passed
- engine `cargo test --all-targets`: 133 passed
- engine `cargo kani --tests --features fuzz --harness proof_v16_expired_counterparty_lien_impairment_is_exact_relabel`: 245 checks passed, 3/3 covers satisfied
- `cargo fmt --all -- --check`
- `git diff --check`

PR #363, opened 2026-07-28
[BLOCKER LoF] Observe backing expiry before PnL conversion

aeyakovenko:
## Public exploit

`ConvertReleasedPnl` did not authenticate backing expiry against the SVM `Clock`. It delegated to the engine, whose source-ledger freshness check compares `expiry_slot` only with the engine's stored `header.current_slot`.

The public LiteSVM reproduction uses ordinary initialized state and no protocol-account mutation:

1. An independent backing authority deposits 150 atoms into the winning source domain with signed expiry slot 3.
2. Two users deposit, trade at 100, and an authenticated mark move to 105 creates 100 atoms of real source-backed PnL.
3. The winner flattens before expiry, making the PnL publicly convertible.
4. No crank advances the engine after slot 2. At authenticated SVM slot 4, the winner calls `ConvertReleasedPnl`.
5. On exact parent `36fa587e1424a8c7fdde2c701c10938188a51876`, conversion succeeds against the expired bucket and the winner immediately withdraws 100 real SPL atoms. The engine still reports slot 2.

This is a public ordering-dependent loss of funds: the backing provider's signed support term has ended, but a claimant can consume and withdraw that principal before an honest expiry crank lands.

## Fix

Before favorable PnL conversion, the wrapper scans the portfolio's bounded source-domain roster and compares every live claim's `Fresh` bucket expiry with the authenticated SVM slot. A lapsed bucket returns engine `Stale` before value mutation, so SVM rollback preserves market, portfolio, and custody bytes.

The check is wrapper-specific because this is where authenticated `Clock` is available. It does not advance global engine time without asset accrual and does not change the engine pin.

The existing expired-source liveness fix remains tracked by wrapper #214 / engine #112. This PR closes the value-extraction race; #214 supplies the bounded permissionless reconciliation route after the safe rejection.

## TDD

- Exact-parent RED: `11d381ec` extracts 100 SPL atoms after expiry.
- Fixed GREEN: `17223602` rejects atomically and preserves market, portfolio, and vault state.

## Verification

- rebuilt auth and hostile matcher SBF fixtures
- focused exploit regression: green
- adjacent conversion authorization, caller-cap, isolation, resolve-maturity, and backing-expiry regressions: green
- `cargo test --all-targets -- --test-threads=1`: 6 unit + 566 LiteSVM/BPF/CU tests passed
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8374,6 +8374,39 @@ pub mod processor {
         Ok(())
     }
 
+    fn reject_lapsed_source_backing_for_conversion_view(
+        group: &state::MarketViewMutV16<'_>,
+        portfolio: &percolator::PortfolioV16ViewMut<'_>,
+        authenticated_slot: u64,
+    ) -> Result<(), V16Error> {
+        let max_market_slots = group.header.config.max_market_slots.get() as usize;
+        if max_market_slots > group.markets.len() {
+            return Err(V16Error::InvalidConfig);
+        }
+        for source in portfolio.header.source_domains.iter() {
+            if !source.is_occupied() || source.source_claim_bound_num.get() == 0 {
+                continue;
+            }
+            let domain = source.domain.get() as usize;
+            let asset_index = domain / 2;
+            if asset_index >= max_market_slots {
+                return Err(V16Error::InvalidConfig);
+            }
+            let engine_slot = &group.markets[asset_index].engine;
+            let bucket = if domain % 2 == 0 {
+                engine_slot.backing_long.try_to_runtime()?
+            } else {
+                engine_slot.backing_short.try_to_runtime()?
+            };
+            if bucket.status == percolator::BackingBucketStatusV16::Fresh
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Independent current-pin confirmation from percolator-meta: https://github.com/aeyakovenko/percolator-meta/pull/385 (`cf2873093eb2597d09ea7403e6ee9787532a0832`).

The clean-room LiteSVM proof system-allocates/publicly initializes every Percolator state account and runs the pinned real SBF. With zero backing fees, an independent provider deposits 150 expiring at slot 3. Public trade/mark/refresh/flatten releases 100 PnL before expiry while engine time ends at slot 2. At authenticated SVM slot 4, current-pin `ConvertReleasedPnl` succeeds. The winner publicly withdraws 100; the provider can recover only 50; consumed backing moves 0 -> 100. This independently confirms exact external-principal loss under the canonical Meta fee policy.

aeyakovenko:

PR #364, opened 2026-07-28
[BLOCKER DoS] Release flat source liens before PnL conversion

aeyakovenko:
## Finding

A normal public trading sequence can leave a flat portfolio with positive, source-backed PnL and a source-credit lien that is no longer needed. `ConvertReleasedPnl` then returns `LockActive`, so the user cannot realize the PnL or close the live portfolio.

The LiteSVM reproduction uses only public instructions: ordinary deposits/trades, authenticated marks, permissionless cranks, backing operations, flattening trades, conversion, SPL withdrawal, and portfolio close. It does not mutate protocol state.

On exact wrapper parent `36fa587e` with engine `143e68c4`, conversion fails with `Custom(21)` after 161,386 CU. The account is flat and an honest crank lands, but the lien remains.

## Fix

Pin engine `4d4f9ed8d5bafc76c37ace7b8eb1550ff08fc739` from companion engine PR #163. The engine makes the proof-only release transition production-reachable inside conversion.

This PR is stacked on wrapper #357 / engine #160. Conversion releases and consumes at most one source domain per instruction, preserving the bounded-work guarantee instead of scanning and mutating every live lien first.

## Merge prerequisite

The final rebase must also include wrapper #306 / engine `14c2919d`, the existing cross-domain source-rounding fix. Without it, aggregate support can advertise one atom that no individual source domain can consume, which is the already-tracked per-domain progress stall rather than a new finding in this PR.

## TDD verification

- Exact parent RED: `ConvertReleasedPnl` returns `LockActive` after all positions are closed and an honest crank succeeds.
- Fixed engine GREEN: PnL converts, liens clear, all capital withdraws, and `ClosePortfolio` succeeds.
- Strengthened public cross-product: an unsigned LP creates multiple real liens, flattens, grows to all 32 source domains, then completes 32 rank-decreasing conversions. Worst deployed-SBF call is 1,321,211 CU, 78,789 below the hard limit.
- `cargo test -- --test-threads=1`: 6/6 wrapper unit and 567/567 LiteSVM/BPF/CU tests passed.
- `cargo fmt --check` and `git diff --check` pass.

The full wrapper Kani suite is running separately; this PR adds no wrapper proof model or production wrapper logic.

Companion engine PR: https://github.com/aeyakovenko/percolator/pull/163

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=ce5e923c48d35989785427975a82ce74b6b19c72#ce5e923c48d35989785427975a82ce74b6b19c72"
+source = "git+https://github.com/aeyakovenko/percolator?rev=4d4f9ed8d5bafc76c37ace7b8eb1550ff08fc739#4d4f9ed8d5bafc76c37ace7b8eb1550ff08fc739"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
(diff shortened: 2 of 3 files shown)


aeyakovenko:

PR #365, opened 2026-07-29
[BLOCKER LoF] Carry fractional oracle caps across max-dt cranks

aeyakovenko:
## What changed

- Pin the engine to the fractional oracle-cap carry implementation from engine PR #164.
- Route AuthMark, EWMA-mark, Hybrid, and pending-observation price checks through the engine's shared capped-price helper.
- Preserve carry across same-direction target updates and reset it on convergence or direction reversal.
- Replace the old rejection-only micro-price regression with successful finite-progress coverage.
- Add public LiteSVM regressions for moving targets, target reversal, and max-dt settlement value.

## Root cause and impact

The percentage cap used:

`floor(effective_price * cap_bps_per_slot * min(dt, max_dt) / 10_000)`

and discarded the fractional numerator after every successful segment. PR #246 prevents a caller from consuming a zero-delta slot before enough elapsed time exists, but elapsed time saturates at `max_accrual_dt_slots`. At sufficiently small prices, even the saturated segment budget remains below one price atom forever.

The exact-parent public repro initializes an ordinary AuthMark market, opens balanced one-unit long/short positions at price 100 between independent owners, accepts an honest target of 1, and uses a 24 bps/slot cap with max dt 20. Permissionless cranks deterministically stall at effective price 20. A later public stale resolution pays the long/short owners `999,920 / 1,000,080`, rather than target-valued `999,901 / 1,000,099`, shifting 19 real SPL atoms from the independent short to the long.

This is reachable without state injection, a compromised admin, or a dishonest oracle mark. It is classified `[BLOCKER LoF]`.

## Fix

The engine carries the sub-atom cap numerator across bounded accrual segments and commits that carry as progress. Once a whole atom is representable, a no-move submission still rejects. Reaching the target or reversing direction resets the carry; fresh reports that keep moving the target on the same side preserve it.

The wrapper computes the same target-directed step before calling engine accrual. It applies the same directional rule, preventing both a reversal-induced `RecoveryRequired` stall and a continuously advancing honest feed from discarding every fraction.

This does not round the cap up and does not permit movement beyond accumulated cap budget.

PR #255's pending-target resolution guard is complementary: it prevents resolution while a target is pending, while this PR guarantees the pending target can actually converge in finite bounded cranks.

## Validation

- Exact-parent RED: wrapper `89d95619` pinned to engine `a8b8877e` stalls at 20 and produces the stale `999,920 / 1,000,080` payouts.
- Reversal RED: the pre-correction SBF rejects the first reversed-target crank with `EngineRecoveryRequired`; fixed SBF succeeds and resets carry to the new direction's `2,400` numerator.
- Moving-feed RED: resetting carry on every fresh target leaves effective price at 100 after ten same-direction reports; fixed head reaches 102.
- Engine: 133/133 tests pass (50 unit, 9 backing fuzz, 2 resolved fuzz, 72 v16 spec).
- Wrapper: 6/6 unit and 569/569 LiteSVM/BPF/CU tests pass.

Depends on engine PR #164 and is stacked on wrapper PR #246.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=ccd834a7610a70d96127109ce9bc098517187b62#ccd834a7610a70d96127109ce9bc098517187b62"
+source = "git+https://github.com/aeyakovenko/percolator?rev=c8ea5cc6e4b56b744d8595d5c30eafe5f911b549#c8ea5cc6e4b56b744d8595d5c30eafe5f911b549"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=a8b8877e784e34fbef7bfd16ce7228130219904b#a8b8877e784e34fbef7bfd16ce7228130219904b"
+source = "git+https://github.com/aeyakovenko/percolator?rev=ccd834a7610a70d96127109ce9bc098517187b62#ccd834a7610a70d96127109ce9bc098517187b62"
 dependencies = [
  "bytemuck",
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #366, opened 2026-07-29
[BLOCKER LoF] Gate liquidation on pending later-leg accrual

aeyakovenko:
## What changed

- Check every active portfolio leg immediately before the auto-crank dispatches liquidation.
- Reject with `EngineNonProgress` when an omitted observation would change that leg's bounded effective price or funding index.
- Keep ordinary stale refresh order-insensitive: only the selected leg is mandatory until the operation becomes an irreversible liquidation.

## Root cause and impact

The wrapper's missing-observation guard covered only the engine-selected first active leg. A public caller could advance an adverse first leg, omit a later backed negative-premium funding segment whose capped price movement rounded to zero, refresh the victim into a current liquidatable certificate, and then land another no-observation crank. The second crank liquidated the victim before the later rescue funding was booked.

The exact pre-fix parent reproduction reduces the independent victim's position from `50,000,000` to `47,995,187` q-units and credits `1,001` atoms to insurance. The correctly ordered control uses ordinary source-domain backing, settles the payer, preserves the full position, and charges no liquidation fee.

This composes the later-leg coverage of #220 with the rounded-funding coverage of #253 without forcing every stale refresh to carry all active-leg observations.

## Validation

- Exact parent: new public-interface LiteSVM test fails with the liquidation and fee evidence above.
- Fixed head: the same test passes and asserts rollback of the rejected liquidation.
- `cargo test --test v16_cu -- --test-threads=1`: **566 passed**.
- Max-shape CU: 14-leg refresh `690,376`; 14-leg liquidation `1,160,120`, both below the 1.4M transaction limit.
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10660,10 +10660,18 @@ pub mod processor {
             let summary = group
                 .build_actionable_summary(&portfolio.as_view())
                 .map_err(map_v16_error)?;
-            if let Some(asset_index) =
+            if summary.liquidatable && !summary.b_stale {
+                reject_missing_pending_liquidation_observations_view(
+                    &cfg,
+                    &group,
+                    &portfolio,
+                    authenticated_now_slot,
+                    observations.as_slice(),
+                )?;
+            } else if let Some(asset_index) =
                 auto_crank_selected_asset_that_accrues_view(&portfolio, &summary)?
             {
-                reject_missing_pending_selected_observation_view(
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-053 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #367, opened 2026-07-29
[BLOCKER LoF] Reject post-expiry backing liens

aeyakovenko:
## Summary

- Reject a trade only when it creates a new counterparty-backed source lien against a bucket that is no longer fresh at the authenticated SVM landing slot.
- Apply the invariant to the shared single/CPI executor and both batch executors.
- Preserve insurance-backed liens, trades with no new lien, and risk-reducing exits.

## Root cause

The engine correctly checks backing expiry against its cached `current_slot`, but a retained transaction can land after that slot and after the provider's signed backing expiry. The wrapper previously did not reconcile a newly-created lien with authenticated `Clock::slot` before charging backing-domain fees.

A normally initialized public LiteSVM reproduction signs a valid `TradeNoCpi` at slot 1, with backing expiring at slot 2, then submits the retained bytes at slot 3 while the engine remains at slot 1. On the exact pre-fix program the trade succeeds, charges the trader 1,312 atoms, credits those atoms to provider earnings, and the provider withdraws 1,312 real SPL atoms. Trader capital falls from 52,501 to 51,189. No protocol state injection, dishonest oracle, or compromised key is used.

## Fix

Snapshot each account's sparse counterparty-backed source liens before the engine trade. After execution, any domain whose counterparty-backed lien increased must have a `Fresh` bucket with `expiry_slot > max(Clock::slot, engine current_slot)`. Otherwise the instruction returns `EngineStale`; SVM rollback restores the engine mutations. The check occurs before backing-fee accounting.

## Validation

- Exact pre-fix red: 1,312 SPL atoms extractable; trader capital `52,501 -> 51,189`.
- Fixed head green: delayed transaction rejects, market and both portfolios roll back byte-for-byte, and a post-expiry risk-reducing trade still succeeds.
- `cargo test --lib`: 6/6 passed.
- Full LiteSVM suite: 566/566 passed.
- Max-shape guard overhead: about 6,945 CU.
- Highest full-suite trade observation: 14-leg `BatchTradeCpi` at 1,236,926 / 1,400,000 CU, leaving 163,074 CU headroom.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -5864,14 +5864,8 @@ pub mod processor {
                 &account_b,
                 core::slice::from_ref(&req),
             )?;
-            let backing_before = if cfg.backing_trade_fee_policy_count == 0 {
-                None
-            } else {
-                Some((
-                    source_counterparty_backing_snapshot_view(&account_a)?,
-                    source_counterparty_backing_snapshot_view(&account_b)?,
-                ))
-            };
+            let backing_before_a = source_counterparty_backing_snapshot_view(&account_a)?;
+            let backing_before_b = source_counterparty_backing_snapshot_view(&account_b)?;
             let source_lien_before_a =
                 source_lien_effective_reserved_snapshot_for_trade_view(&account_a)?;
             let source_lien_before_b =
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Meta current-pin integration RED is now isolated in https://github.com/aeyakovenko/percolator-meta/pull/389 at commit `f32a4cdd0b71b28368395c6f9e4c217387757e0b`.

It uses the pinned real Percolator SBF with system-allocated/publicly initialized market, portfolio, and backing-ledger accounts. The external backing provider retains the user's signed trade, submits at authenticated slot 3 after expiry 2 while engine slot remains 1, and publicly withdraws 1,312 real SPL atoms; trader capital falls 52,501 -> 51,189. The ignored test requires rejection, zero extraction, full rollback, and a live fresh-trade continuation once #367 is pinned.

aeyakovenko:

PR #368, opened 2026-07-29
[BLOCKER DoS] Prevent expired partial recovery from locking market

aeyakovenko:
## Impact

A public portfolio owner can turn an account-local, partially completed recovery forfeit into a permanent global market lock after a legitimate asset shutdown. An unrelated flat user's real SPL principal becomes unreachable even with an honest cranker supplying the correct transaction.

## Public TDD reproduction

The regression uses only normal accounts and public instructions:

1. Initialize a production-risk market and three portfolios.
2. Open a normal long/short pair and fund an unrelated flat victim.
3. Move the authenticated mark through bounded public updates.
4. Shut down the asset through the public lifecycle instruction.
5. Owner-sign `ForfeitRecoveryLeg`, which commits one bounded chunk and leaves a real nonzero residual.
6. After its deadline, submit unsigned `PermissionlessCrank`.

Against exact parent `36fa587e1424a8c7fdde2c701c10938188a51876` with engine `143e68c4917ed0400a27b952f036a5677047cd84`, step 6 commits `MarketModeV16::Recovery`, creating the global sink. No protocol state is injected in this exploit test.

With engine PR aeyakovenko/percolator#165, the same crank reaches `Resolved` atomically and the unrelated owner withdraws exactly 100 real SPL tokens through the bounded `CloseResolved` path.

A separate, explicitly marked migration regression injects only a pre-upgrade persisted `Recovery` fixture and proves the wrapper's sole public `PermissionlessCrank` resolves it. It is supplemental upgrade coverage, not the exploit reachability evidence.

## Changes

- Add the public-interface LiteSVM exploit/regression.
- Add the persisted-state auto-crank migration regression.
- Update existing terminal-recovery expectations from `Recovery` to `Resolved`.
- Pin engine `d7b29edb943ebf490ac8ea3d95798cf9b44deba3` from aeyakovenko/percolator#165.

No wrapper program source is changed.

## Verification

- `cargo build-sbf --tools-version v1.52`: passed.
- `cargo test --all-targets -- --test-threads=1`: 567/567 LiteSVM/CU tests and 6/6 wrapper unit tests passed.
- Engine runtime: 130/130 passed.
- Engine Kani: 4,543-check direct resolution proof and 14,839-check whole auto-crank proof passed.

## Known dependency

The regression here proves that the unrelated flat user regains a custody exit. Complete wind-down of the loss-side cohort for this small valid B budget also requires engine #155, which fixes the independent loss-atom/B-index unit mismatch. This PR intentionally does not duplicate that existing finding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=d7b29edb943ebf490ac8ea3d95798cf9b44deba3#d7b29edb943ebf490ac8ea3d95798cf9b44deba3"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "d7b29edb943ebf490ac8ea3d95798cf9b44deba3" }
 anchor-lang-v2 = { git = "https://github.com/solana-foundation/anchor", rev = "e3cf760826fa0a60f247635ea0572e59861ec9b5", package = "anchor-lang-v2", default-features = false, features = ["alloc", "guardrails"], optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Stacking update: this DoS-only PR is now based on #255 so its terminal auto-recovery composes with the existing committed-accrual resolve guard. The newly discovered cross-asset value-loss fix is isolated in draft #370 on top of this PR; #368 itself still contains only the engine pin and DoS/migration tests.

aeyakovenko:
Independent Meta clean-room reproduction: aeyakovenko/percolator-meta#397 at `e09a0ad617c7d51a431279d29fa50f17575a4301`.

The real-SBF LiteSVM probe is RED against Meta's current pin (`percolator-prog` `557f15b`, engine `e3ea725`, SBF SHA-256 `74ce6b85...40a4`): the market remains globally stuck in `Recovery` after all bounded keeper account choices, authority resolution, owner forfeit/cure/rebalance, and unrelated `CloseResolved` probes either roll back or make byte-for-byte no progress.

The identical test is GREEN against this PR head `2928da6` / engine `d7b29ed` (rebuilt SBF SHA-256 `1a135bba...a070`) and returns the unrelated depositor's exact 100 SPL units. No initialized Percolator state is injected.

aeyakovenko:

PR #369, opened 2026-07-29
[BLOCKER LoF] Require bilateral fee support for EWMA movement

aeyakovenko:
## Root cause

`collected_fee_supported_mark` bounded trade-driven EWMA movement from the aggregate fee collected from both trade sides. A distressed risk-reducing trader can pay only its remaining capital while an independently authorized CPI LP pays the full dynamic fee. The LP's payment then supported the entire mark move, even though the underfunded side could hold a separate position that profits from that move.

This is distinct from the existing uncollectible-fee regression: the fee is genuinely collected, but from the wrong economic side for the externality it funds.

## Public exploit

The RED test uses only deployed SBF instructions and ordinary accounts:

- independently owned distressed mover, fee-paying passive LP, attacker beneficiary, victim, and extraction LP
- real SPL deposits and withdrawals
- independently authenticated EWMA mark or external Pyth hybrid feed
- public single or batch CPI trades and permissionless cranks
- no protocol-state injection, compromised key, or privileged attacker

On exact parent `36fa587e`:

| Mode / route | Queued mark | Attacker delta | Independent victim delta |
| --- | ---: | ---: | ---: |
| EWMA single/batch | 1,997,385 | +792,859 | -883,860 |
| Hybrid-after-hours single/batch | 2,090,398 | +903,989 | -903,990 |

The attacker closes through a separately authorized LP, converts released PnL, and withdraws the gain as SPL tokens.

## Fix

Subtract the equal per-side base charge independently, then support internal mark movement from `2 * min(externality_fee_a, externality_fee_b)`. Either side's payment alone therefore supports zero movement. When both sides pay, the existing aggregate economics are unchanged.

Arbitrary-price risk-reducing exits remain live. The wrapper still executes the trade and collects whatever fees are available; only its internal mark movement is clamped to bilateral paid support.

At head:

- EWMA single/batch: queued mark 1,913,198; attacker delta -49,011
- Hybrid-after-hours single/batch: mark remains 1,999,999; attacker delta -1
- the independent LP still pays a real fee and insurance receives it

## Validation

- `cargo fmt --all`
- `git diff --check`
- exact-parent RED for all four mode/route combinations
- focused GREEN for all four mode/route combinations
- `cargo test --all-targets -- --test-threads=1`: 6/6 unit and 566/566 LiteSVM/BPF/CU tests pass
- `cargo kani --tests --harness kani_v16_one_sided_externality_fee_cannot_fund_mark_movement`: 43 checks pass and the nontrivial cover property is satisfied

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4138,9 +4138,16 @@ pub mod policy_v16 {
         if quoted_mark_e6 == 0 || mark_externality_notional == 0 {
             return Some(quoted_mark_e6);
         }
-        let collected = fee_a.checked_add(fee_b)?;
-        let externality_fee = collected.saturating_sub(base_fee_paid);
-        let supported_move_bps = externality_fee
+        // Each side can hold other positions that benefit from the mark move. Do not let one
+        // underfunded side use its counterparty's payment to make that separate book profitable.
+        // The quote's base fee is exactly two equal per-side charges, so only matched externality
+        // payments can support movement.
+        let base_fee_per_side = base_fee_paid / 2;
+        let externality_fee_a = fee_a.saturating_sub(base_fee_per_side);
+        let externality_fee_b = fee_b.saturating_sub(base_fee_per_side);
+        let matched_externality_fee =
+            core::cmp::min(externality_fee_a, externality_fee_b).checked_mul(2)?;
+        let supported_move_bps = matched_externality_fee
             .checked_mul(10_000)?
             .checked_div(mark_externality_notional)?;
         Some(clamp_mark_to_supported_move_bps(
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Meta current-pin LiteSVM reproduction is now isolated in https://github.com/aeyakovenko/percolator-meta/pull/388 at commit `f8688e01ec040adde9e0670955f309ba236122e8`. It uses public initialization, real SPL deposits/withdrawals, the pinned Percolator SBF, and the production matcher. The trace matches this PR's exact-parent EWMA reproducer: queued mark `1,997,385`, attacker extraction `+792,859`, independent victim `-883,860`. The Meta PR remains an ignored RED until this upstream fix is merged and repinned.

aeyakovenko:

PR #370, opened 2026-07-29
[BLOCKER LoF] Preserve pending accrual during terminal recovery

aeyakovenko:
## Impact

An unprivileged caller could use expired terminal recovery on one asset to resolve the market before an authenticated mark was committed on an unrelated exposed asset. In the public two-asset repro, this erased the long's accrued profit and let the opposing short withdraw an extra 10,000,000 collateral atoms.

The same attack remained possible when an old-price crank ran before the honest mark in the terminal slot. A target-only same-slot crank updated the engine raw target without advancing effective price, and terminal recovery still discarded 2,399,000 atoms of the long's value.

## Root cause

PR #368 adds an expired-close fast path that calls the engine auto-crank with no observations. That route could reach global resolution without applying the committed-accrual guard used by the explicit resolve path in PR #255.

## Fix

- Apply the shared committed-accrual preflight before expired-close auto-recovery.
- Treat authenticated wrapper/engine target divergence as pending accrual.
- Also detect the same-slot target-only state where raw target has advanced but effective price has not.
- Return `EngineStale` and rely on SVM rollback until a bounded public crank commits the mark; retrying terminal recovery then succeeds.

## TDD

`v16_attack_expired_recovery_cannot_discard_unrelated_pending_mark` uses only public instructions and real SPL balances. It covers:

- ordinary pending authenticated mark: 10,000,000-atom attempted transfer;
- same-slot late mark after an old-price crank: 2,399,000-atom attempted transfer;
- a successful target-only crank that must still not authorize terminal resolution;
- bounded next-slot catch-up followed by successful terminal recovery.

The test fails against the exact stacked parent and passes with this commit.

## Verification

- focused public exploit test
- `cargo test --all-targets -- --test-threads=1`: 573 passed
- `git diff --check`

This draft is stacked on #368, which is stacked on #255.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8714,17 +8714,27 @@ pub mod processor {
         while asset_index < configured_slots {
             let asset = group.markets[asset_index].engine.asset;
             let exposed = asset.oi_eff_long_q.get() != 0 || asset.oi_eff_short_q.get() != 0;
-            if exposed && asset.slot_last.get() < resolved_slot {
+            if exposed {
                 let profile = read_oracle_profile_from_view(group, cfg, asset_index)?;
-                if committed_market_resolve_accrual_for_profile_view(
-                    &profile,
-                    group,
-                    asset_index,
-                    resolved_slot,
-                    include_pending_mark,
-                )?
-                .is_some()
-                {
+                // A mark can land after the engine already advanced this asset in the same slot.
+                // Compare the authenticated wrapper target with the engine's committed raw target;
+                // effective-price equality or slot equality alone cannot distinguish that order.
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_expired_partial_close_matrix_resolves_and_preserves_idle_exit`. PR370 is stacked on the PR368 terminal-recovery repair; the integrated route reaches Resolved through bounded public continuations and preserves the unrelated funded owner's exact exit. The executable disposition remains in PR #135; no separate production fix is required.

PR #371, opened 2026-07-29
[BLOCKER DoS] Keep winner-first Recovery forfeits live

aeyakovenko:
## Impact

A normally initialized two-asset market could strand the bankrupt side of a secondary-asset shutdown permanently when the winning counterparty forfeited its Recovery leg first. The same sequence could leave provider backing permanently liened. The remaining loser had no successful public continuation while an unrelated fresh base asset kept the market live, so SVM rollback preserved the stuck state.

## Public reproduction

The LiteSVM regression uses only public program instructions and valid account state:

1. Initialize a production-risk two-asset market and three funded portfolios.
2. Fund asset 1's backing domain and trade one lot on asset 1.
3. Move the authenticated mark within configured bounds, materialize the winner's PnL, then open public cross-margin risk on asset 0 so the winner carries a real asset-1 provider-backing lien.
4. Continue bounded authenticated marks and cranks until the asset-1 long is bankrupt.
5. Shut down asset 1 through the normal admin lifecycle route.
6. Have the winning short owner call `ForfeitRecoveryLeg` first, then have the losing long owner do the same.

Against parent `36fa587e` pinned to engine `143e68c`, the loser's call returned `Custom(23)` (`RecoveryRequired`) at 77,366 CU. Rollback erased its attempted declaration/ledger progress, and the winner's exit left 22.6k atoms of provider backing liened. Retry, risk reduction, force close, and auto crank provided no successful exit; keeping asset 0 fresh also prevented global resolution from masking the bug.

## Root cause

The first owner could detach after materializing positive PnL and its matching source claim. The old forfeit path deleted that leg's social-loss weight without waiving the materialized claim or releasing the valid source lien. That contradicted the dead-leg rule: positive PnL must be valued at zero unless source backing is consumed. With no recipient weight left, the bankrupt counterparty could not settle its residual and every local exit reverted.

Engine PR aeyakovenko/percolator#166 fixes the accounting by burning the selected leg's materialized positive PnL/source claim, releasing its valid source liens without touching unrelated domains, and allowing claim-free terminal explicit loss. This wrapper PR contains the public TDD regression and pins engine commit `a9a8588483ac6ecccd5d94bcbc80e116cf5487a8`.

## Validation

- Parent-red public LiteSVM reproduction: loser forfeit fails with `Custom(23)` at 77,366 CU and 22.6k provider atoms remain liened.
- Fixed focused regression: all three portfolios close, all OI/loss weights clear, and the backing provider withdraws the full 100k principal.
- Wrapper full suite: 566/566 LiteSVM tests plus all unit/example targets pass.
- Engine full suite: 132/132 tests pass.
- Engine targeted Kani contracts: `kernel_terminal_explicit_loss_allowed` 211/211 and `kernel_bresidual_step` 147/147 pass.
- `cargo fmt --check` and `git diff --check` pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=143e68c4917ed0400a27b952f036a5677047cd84#143e68c4917ed0400a27b952f036a5677047cd84"
+source = "git+https://github.com/aeyakovenko/percolator?rev=a9a8588483ac6ecccd5d94bcbc80e116cf5487a8#a9a8588483ac6ecccd5d94bcbc80e116cf5487a8"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "143e68c4917ed0400a27b952f036a5677047cd84" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "a9a8588483ac6ecccd5d94bcbc80e116cf5487a8" }
 anchor-lang-v2 = { git = "https://github.com/solana-foundation/anchor", rev = "e3cf760826fa0a60f247635ea0572e59861ec9b5", package = "anchor-lang-v2", default-features = false, features = ["alloc", "guardrails"], optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
 # Host tests use the same zero-copy account/view API as the SBF/program target.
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Confirmed this still applies to the current `percolator-meta` pins with a clean-room, real-SBF LiteSVM reproduction: https://github.com/aeyakovenko/percolator-meta/pull/395 (`348f867`).

Meta pins wrapper `557f15b63c7a5f9e058a2c384ca8728829dc2b63` and engine `e3ea7251ed75f0111ba3cf5a4b87961f6c7cca63`; that engine line does not contain `a9a8588`. After the winner forfeits first, the loser forfeit and retry both return `Custom(23)`, all bounded local owner/keeper alternatives reject and roll back byte-for-byte, honest asset-0 cranking keeps the market live at slot 1,000, stale resolution rejects, and exactly 22,600 provider atoms remain valid-liened. No initialized market/portfolio state is injected; only external SPL fixtures are installed. This is a real persistent public DoS for Meta, not only an upstream unit-harness result.

aeyakovenko:

PR #372, opened 2026-07-29
[BLOCKER LoF] Preserve prior same-domain claim across Recovery forfeit

aeyakovenko:
## Impact

A normally initialized, provider-backed market can erase or strand an independent user's settled profit from an earlier position episode.

The public LiteSVM reproduction uses only wrapper instructions:

1. The victim earns exactly 48,000 atoms on asset 1 and fully closes that position.
2. The victim reopens the same asset and side at the current authenticated mark.
3. The asset enters normal Recovery and the new counterparty forfeits first.
4. Conversion, unilateral reduction, and abandoned-asset pair close cannot remove the victim's leg. Asset 0 remains honestly refreshed, so global resolution does not rescue it.
5. On exact parent engine `a9a8588483ac6ecccd5d94bcbc80e116cf5487a8`, the victim's unilateral forfeit burns the earlier claim.

The strengthened case then earns episode-two PnL and increases risk until the source lien crosses the saved historical floor. The first fix aborted that public forfeit with `LockActive`, leaving the victim persistently stuck.

No market or portfolio bytes are injected. Every account, deposit, backing top-up, trade, mark update, shutdown, failed alternative, and forfeit is constructed through public SBF instructions.

## Fix and TDD

This PR pins engine #167 at `d3133a93cf2cf219b69e0072cb4079d24a07f4f6`.

The engine snapshots the claim bound present when a leg episode attaches. A Recovery forfeit waives only the current episode's suffix. If a later valid lien overlaps the historical floor, the same transaction realizes the older claim into capital before detaching, preventing both `LockActive` and a provider-withdrawal race.

The strengthened test proves:

- the attacker removes the only opposite leg first;
- the victim exits in one bounded public forfeit;
- exactly 48,000 historical atoms become senior capital while episode-two PnL is waived;
- the backing provider can immediately withdraw every unconsumed backing atom; and
- the victim still withdraws all 53,000 remaining atoms through canonical SPL custody.

This PR is intentionally stacked on wrapper #371 / engine #166.

## Verification

- Exact-parent public LiteSVM: RED
- Fixed-head focused Recovery suite: 3/3 passed
- `cargo test --all-targets -- --test-threads=1`: 568 LiteSVM tests passed
- Full CU/max-shape suite included in those 568 tests
- `cargo check --tests`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=a9a8588483ac6ecccd5d94bcbc80e116cf5487a8#a9a8588483ac6ecccd5d94bcbc80e116cf5487a8"
+source = "git+https://github.com/aeyakovenko/percolator?rev=58c085d6c521a98788ee123809c34980d54958e5#58c085d6c521a98788ee123809c34980d54958e5"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=58c085d6c521a98788ee123809c34980d54958e5#58c085d6c521a98788ee123809c34980d54958e5"
+source = "git+https://github.com/aeyakovenko/percolator?rev=d3133a93cf2cf219b69e0072cb4079d24a07f4f6#d3133a93cf2cf219b69e0072cb4079d24a07f4f6"
 dependencies = [
  "bytemuck",
 ]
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Independent Meta clean-room confirmation: https://github.com/aeyakovenko/percolator-meta/pull/396 (`07a92707e75e9f64ac65f62a45ea060a1b68003c`).

Using the same public LiteSVM/SBF transaction sequence and canonical SPL withdrawals:

- exact #371-only parent SBF (`50e2c1d`): attacker supplies 1,100,000 atoms and recovers 1,148,000; victim recovers 1,000,000 instead of 1,048,000;
- #372 head SBF (`5403e61`): attacker recovers exactly 1,100,000 and victim recovers all 1,048,000.

The current Meta pin reaches the #371 B-stale deadlock first, so Meta #396 is intentionally stacked on Meta #395 and demonstrates why both upstream fixes must be repinned together.

aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_prior_claim_forfeit_prerequisite_matrix_preserves_withdrawable_value`. PR372 is stacked on the PR371 Recovery-forfeit repair; on the pinned predecessor its exact two-episode public lifecycle preserves the prior 48,000-atom claim and pays all 1,048,000 atoms through owner-signed CloseResolved. The executable disposition remains in PR #135; no separate production fix is required.

PR #373, opened 2026-07-29
[BLOCKER LoF] Prove retained Recovery haircuts preserve prior claims

aeyakovenko:
## Impact

This adds a clean-room public-interface LiteSVM reproduction for a retained, honestly signed `ForfeitRecoveryLeg` transaction that becomes destructive after an underfunded counterparty forfeits first and moves the market's `B` target.

On the exact stacked parent, the victim's older, fully closed, provider-backed same-domain claim falls from 48,000 atoms to 39,576. The 8,424-atom loss is caused entirely by public transaction ordering; the victim signed while its `B` snapshot was current.

This is `[BLOCKER LoF]`, not signer self-harm: an unprivileged relayer retains unchanged signed bytes and lands them within the valid blockhash window after another public transaction changes the haircut state.

## Coverage

The regression exercises the real SBF program and canonical custody path without state injection:

- creates and fully closes the first profitable position episode
- verifies a real 48,000-atom source-attributed claim
- reopens the same asset and side and accrues 59,424 atoms
- signs the victim forfeit before any `B` loss exists
- lets an underfunded counterparty forfeit first and book an 8,424-atom haircut
- proves conversion, unilateral reduction, and pair-close alternatives are unavailable
- submits the retained transaction through an unprivileged relayer
- keeps the unrelated base market honestly live
- resolves the conservative bankruptcy lock through the existing bounded authority
- verifies owner-signed `CloseResolved` pays exactly 1,048,000 SPL atoms

## Fix dependency

Pins engine commit `0928823d3cb883c762042e7348677944044ef429` from aeyakovenko/percolator#168. The engine nets current-episode `K`, `F`, and `B` before forfeiting the current episode's positive suffix, so those losses cannot reach below the attach-time claim floor.

No wrapper production logic or authority surface is changed.

## Validation

- Exact-parent RED on engine `d3133a93cf2cf219b69e0072cb4079d24a07f4f6`: `39,576 != 48,000`
- Fixed focused GREEN on engine `0928823d3cb883c762042e7348677944044ef429`
- SBF build with toolchain `v1.52`
- 6 wrapper unit tests pass
- all 569 LiteSVM/BPF tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=d3133a93cf2cf219b69e0072cb4079d24a07f4f6#d3133a93cf2cf219b69e0072cb4079d24a07f4f6"
+source = "git+https://github.com/aeyakovenko/percolator?rev=0928823d3cb883c762042e7348677944044ef429#0928823d3cb883c762042e7348677944044ef429"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "d3133a93cf2cf219b69e0072cb4079d24a07f4f6" }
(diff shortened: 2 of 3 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_retained_recovery_haircut_prerequisite_matrix_keeps_prior_claim_floor`. PR373 is stacked on PR372; with the exact retained-before-haircut schedule the pinned predecessor keeps 99,000 PnL, never drops below the earlier 48,000-atom claim, and pays 1,099,000 atoms through CloseResolved. The executable disposition remains in PR #135; no separate production fix is required.

PR #374, opened 2026-07-29
[BLOCKER LoF] Reconcile expired backing in Recovery forfeits

aeyakovenko:
## Impact

A victim can retain an honestly signed `ForfeitRecoveryLeg` transaction before a backing bucket expires and an unprivileged relayer can land it afterward. On the exact parent (`32faddebec5909c71dab414a6a0da44e54319257`), the transition capitalized a historical Recovery claim with lapsed external backing: 48,000 atoms of an independent provider's 100,000-atom principal became withdrawable by the victim.

## Root cause

The wrapper authenticated maturity-sensitive actions with `Clock`, but Recovery forfeit called an engine transition that neither received that execution slot nor reconciled source-bucket expiry. Retained transactions could therefore cross the expiry boundary while the engine still used a cached pre-expiry slot.

## Change

- Pin the engine to aeyakovenko/percolator#169 (`0af8bf28ab9013ea3bd55dc7f56ede6feaa9b1b5`).
- Pass the authenticated execution slot into the Recovery forfeit transition.
- Reconcile only the selected source domain atomically: expired backing is impaired, never capitalized.
- Preserve the user's historical 48,000-atom impaired claim, detach the dead leg, and return exactly the user's own 5,000-atom deposit.

The first reject-only guard was intentionally superseded. Its exact-parent RED returned `Custom(19)` after expiry and permanently stranded the local Recovery exit whenever another healthy market prevented global resolution. The final transition preserves both fund safety and liveness.

This adds no custody path, admin authority, DAO authority, or ability to move user/provider funds.

## Validation

- Exact-parent LoF RED: retained transaction consumed 48,000 atoms of expired provider principal.
- Exact reject-only RED: final safe-exit test fails with `Custom(19)` and leaves the Recovery leg stranded.
- Fixed public LiteSVM/SBF regression: retained transaction lands after expiry, consumes 0 provider principal, detaches the leg, preserves the 48,000-atom impaired claim, and returns exactly the 5,000-atom user deposit.
- Claim-free post-expiry liveness control.
- Wrapper suite: 6 unit + 571 LiteSVM/BPF tests, 0 failures.
- Engine suite: 136 tests; Kani closure proof: 326 checks, 0 failures.
- Production SBF rebuilt with the exact engine pin.
- `cargo fmt --all -- --check` and `git diff --check`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=0928823d3cb883c762042e7348677944044ef429#0928823d3cb883c762042e7348677944044ef429"
+source = "git+https://github.com/aeyakovenko/percolator?rev=0af8bf28ab9013ea3bd55dc7f56ede6feaa9b1b5#0af8bf28ab9013ea3bd55dc7f56ede6feaa9b1b5"
 dependencies = [
  "bytemuck",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,12 +50,12 @@ pinocchio = "0.6"
 arrayref = "0.3"
 num-derive = "0.4"
 num-traits = "0.2"
-percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "0928823d3cb883c762042e7348677944044ef429" }
+percolator = { git = "https://github.com/aeyakovenko/percolator", rev = "0af8bf28ab9013ea3bd55dc7f56ede6feaa9b1b5" }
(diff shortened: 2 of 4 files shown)


aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_retained_recovery_expiry_prerequisite_matrix_avoids_provider_capitalization`. PR374 is stacked after the prior-claim fixes; its exact retained-after-expiry transaction lands on the pinned predecessor but leaves only 5,000 user capital, 68,000 attributed PnL, and zero consumed provider backing, so the claimed 48,000-atom principal capitalization does not occur. The executable disposition remains in PR #135; no separate production fix is required.

PR #375, opened 2026-07-29
[PRIVILEGED LoF] Prevent asset-admin seizure of funded roles

aeyakovenko:
## Impact

A separately delegated per-asset cold admin could replace an independent funded backing authority, live insurance operator, or terminal insurance authority without the incumbent signing. The replacement could then withdraw the incumbent provider/funder's tokens. This is a bounded privileged-role LoF, not a permissionless attack and not an initialization footgun.

## Fix

- Permit asset-admin recovery while the affected backing/insurance role is empty.
- Require a funded incumbent role holder to self-rotate.
- Preserve existing market-authority cascade behavior, oracle/admin rotation, and all public withdrawal paths.

## Public-interface TDD

Against exact parent `36fa587e1424a8c7fdde2c701c10938188a51876`, all three public LiteSVM probes fail and transfer 500 tokens to the replacement:

- funded backing-provider takeover and withdrawal
- funded live-insurance-operator takeover and withdrawal
- funded terminal-insurance-authority takeover and withdrawal after normal resolution

At this head, all three takeovers reject while incumbent withdrawals and provider-approved self-rotation succeed. No account bytes are injected or mutated by the tests.

## Verification

- `cargo fmt --check`
- `cargo test --lib --quiet` (6/6)
- `cargo test --test v16_cu --quiet` (568/568)
- engine pin unchanged (`143e68c...`)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -8739,6 +8739,43 @@ pub mod processor {
         state::write_wrapper_config(&mut data, &cfg_after)
     }
 
+    fn asset_role_holds_funded_value_view(
+        group: &state::MarketViewMutV16<'_>,
+        asset_index: usize,
+        kind: u8,
+    ) -> Result<bool, ProgramError> {
+        let long_domain = asset_index
+            .checked_mul(2)
+            .ok_or(PercolatorError::EngineArithmeticOverflow)?;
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #376, opened 2026-07-29
[BLOCKER DoS] Let asset-0 backing provider complete recovery

aeyakovenko:
## Root cause

A real insurance-consuming bankruptcy sets the durable market bankruptcy hlock. After asset 0 is shut down and all positions and loss state are cleared, that global lock still rejects the recorded provider's backing withdrawal. Asset-0 restart requires both backing buckets to be empty, while the existing shutdown-drain exception deliberately excludes asset 0. The only remaining cleanup path was global resolution, so an otherwise reusable live market became terminally stuck.

## Fix

- Share the empty, mature Recovery predicate without granting asset-0 shutdown-drain authority.
- Let only the existing local backing provider pass the global health lock for asset 0 after Recovery is empty and mature.
- Return `false` from the exception so market governance does not acquire the provider's withdrawal authority.
- Keep insurance withdrawals on the original health gate.

No new authority is introduced and the durable bankruptcy lock is not cleared. Engine withdrawal checks still require fresh, unencumbered backing.

## Regression coverage

The LiteSVM regression reaches the state entirely through public instructions: deposits, trade, authenticated mark movement, permissionless cranks, insurance consumption, shutdown, owner forfeiture, and permissionless side finalization. It verifies that:

- the provider cannot withdraw while the asset is live;
- the provider cannot withdraw before the recovery delay;
- market governance cannot substitute for the external provider;
- insurance remains locked; and
- the recorded provider can withdraw after empty, mature Recovery.

The regression fails on the unmodified wrapper binary with `Custom(21)` at the final provider withdrawal and passes on this branch.

## Validation

- `CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 cargo test --lib`
- `CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 cargo test --test v16_cu` (566 passed)

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4438,14 +4438,13 @@ pub mod processor {
         Ok(())
     }
 
-    fn shutdown_asset_matured_at_slot_view(
+    fn recovery_asset_matured_at_slot_view(
         cfg: &WrapperConfigV16,
         group: &state::MarketViewMutV16<'_>,
         asset_index: usize,
         now_slot: u64,
     ) -> ProgramResult {
         if group.header.mode != 0
-            || asset_index == 0
             || cfg.force_close_delay_slots == 0
             || asset_index >= group.header.config.max_market_slots.get() as usize
             || asset_index >= group.markets.len()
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Meta current-pin clean-room confirmation: https://github.com/aeyakovenko/percolator-meta/pull/386 at commit `50963a4ef1e52e301a737c7cdb21fb28bb2c28a1`.

This strengthens the original reproduction with an actual independent 100-atom backing deposit in the opposite source bucket before the public bankruptcy. After the hlock, asset-0 shutdown, owner forfeits, both public side finalizations, and the Recovery delay, the bucket still reports exactly zero consumed principal and at least the full 100 atoms fresh/unliened. Governance cannot substitute, but the recorded provider's exact 100-atom withdrawal fails on Meta's pinned wrapper with `Custom(21)`.

The PR #376 predicate is appropriately narrow for this case: recorded provider only, asset 0 only, empty and mature Recovery, and it does not unlock insurance or give governance a drain capability.

aeyakovenko:

PR #283, opened 2026-07-20
[BLOCKER LoF] Prevent dust backing from erasing terminal payouts

aeyakovenko:
## Public exploit

A normally initialized EWMA market with a separate honest oracle is vulnerable through both public `TradeNoCpi` and one-leg `BatchTradeNoCpi`:

- victim and directional counterparty each deposit 20,000,000,000 quote atoms;
- an attacker-controlled dust pair opens and closes one raw position quantum around an honest bounded mark round trip;
- the attacker coalition deposits 20,000,002,000 and withdraws 20,000,001,999, paying exactly one atom;
- the victim withdraws only 11,557,223,273, losing 8,442,776,727;
- 8,442,776,728 remains in the canonical vault after repeated permissionless `CloseResolved` and `ClaimResolvedPayoutTopup`, with no remaining live claim.

The no-dust control returns both parties' full deposits and leaves the vault empty. The attacker never signs an oracle update or market resolution.

## Root cause and fix

Terminal source realization used inverse-rate source face consumption as terminal forfeiture. At a tiny credit rate, one effective source atom could remove nearly the entire winner face from the payout denominator.

This PR keeps the exact-parent RED LiteSVM test unchanged and pins engine PR #145. The engine now removes exactly the effective source payment `X` from original terminal face `F`, leaving an exact receipt for `F - X`.

## TDD evidence

Exact parent `da64be639168b496833dd4c701607a94fcb06b6c`, vulnerable engine `143e68c4917ed0400a27b952f036a5677047cd84`:

```text
control: attacker=20000002000 victim=20000000000 vault=0
attack:  attacker=20000001999 victim=11557223273 vault=8442776728
```

Fixed engine `3db81b0b0484754dffb8a9c10e8d86076e969c8d`:

```text
control: attacker=20000002000 victim=20000000000 vault=0
attack:  attacker=20000001999 victim=20000000000 vault=1
```

Both single and batch paths reproduce RED and pass GREEN.

## Verification

- engine normal and fuzz/Kani test matrices pass;
- final wrapper stack built with `cargo build-sbf --tools-version v1.52`;
- 6 unit tests and 567 LiteSVM/CU tests pass against the exact final SBF;
- the terminal one-unit source-backing payout regression passes.

Stacked on #376.

PR #377, opened 2026-07-29
[BLOCKER DoS] Keep fresh insurance withdrawable after asset restart

aeyakovenko:
## Finding

After a real asset-0 bankruptcy is fully settled and its recovered backing is drained, `RestartAssetOracle` starts a fresh asset generation and resets the asset-local insurance-spent counters. The market-wide `bankruptcy_hlock_active` bit intentionally remains durable, however, so every subsequent live `WithdrawInsuranceAsset` still failed with `EngineLockActive` (`Custom(21)`).

That permanently prevented TWAP from consuming fresh post-restart protocol insurance and could strand new external insurance deposits in an otherwise healthy live generation.

## Fix

The durable lock is not cleared. Only the per-asset `WithdrawInsuranceAsset` path may treat it as historical, and only when the exact target asset:

- is `Active`;
- has zero long and short generation-spent counters; and
- has no local OI, stored/stale accounts, pending obligations, B/loss state, or pending loss barriers.

Backing withdrawals retain the unconditional h-lock gate. Threshold stress, loss staleness, recovery reason, target/effective lag, per-domain budgets, and global/source reservation capacity remain unchanged. No authority, recipient, amount, or custody-transfer surface is added.

## TDD evidence

Exact-parent RED on `867ca977b907272e63538b3ea18a204da0d46355`:

- real bankruptcy and backing recovery completed;
- asset restart completed;
- fresh domain insurance top-up completed;
- the one-atom live withdrawal failed with `Custom(21)`.

Fixed GREEN against the real LiteSVM/SBF path:

- `v16_bpf_asset0_bankruptcy_recovery_lets_only_backing_provider_drain`
- `v16_attack_live_insurance_withdraw_rejects_while_stressed_or_hlocked`

The second regression proves pending local obligations and fresh balanced exposure keep the h-lock protective, while threshold and loss-stale flags remain unconditional blockers.

## Verification

Built the final stacked SBF against engine `5d608a8653a4d2535b3ff7d3a011ccc8f4accc23` with `cargo build-sbf --tools-version v1.52`.

- 6 unit tests passed
- 567 LiteSVM/CU tests passed
- SBF SHA-256: `619773eb0931e23682383ff0ce544e99dd0d6bb312704140963caeb20d2809c9`

Stacked on #379, #283, and #376.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,7 +677,15 @@ There are two different insurance withdrawal surfaces:
 - live `WithdrawInsuranceAsset`, which is a per-asset operator path bounded by that asset's funded
   long+short insurance budget
 
-Live insurance withdrawal is intentionally stricter. It is expected to be allowed only when the live market is flat or loss-current, target/effective-lag-free, stress-free, h-lock-free, and has non-negative senior residual. In other words, live insurance can be withdrawn from an empty or fully healthy market, but not while the insurance fund is still protecting unresolved loss or bankruptcy work.
+Live insurance withdrawal is intentionally stricter. It is expected to be allowed only when the
+live market is flat or loss-current, target/effective-lag-free, stress-free, and has non-negative
+senior residual. A bankruptcy h-lock remains an unconditional blocker for backing withdrawals and
+for an insurance asset with any local position, stale account, pending obligation, loss state,
+generation spend, or loss barrier. The per-asset insurance operator may bypass only a historical
+market-wide h-lock after that target asset has restarted Active with an empty local generation.
+Per-domain budgets and global/source reservation capacity still prevent that path from consuming
+insurance assigned to another asset. Threshold stress, loss staleness, and recovery reason remain
+unconditional blockers.
 
 ### Product intuition
 
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -4515,6 +4515,7 @@ pub mod processor {
         group: &state::MarketViewMutV16<'_>,
         domain: usize,
         allow_asset0_local_backing_recovery: bool,
+        allow_empty_asset_historical_bankruptcy: bool,
     ) -> Result<bool, ProgramError> {
         let asset_index = domain / 2;
         if shutdown_asset_empty_and_matured_now_view(cfg, group, asset_index).is_ok() {
@@ -4531,7 +4532,9 @@ pub mod processor {
             }
         }
         reject_permissionless_resolve_matured_live_view(cfg, group)?;
-        if group.header.bankruptcy_hlock_active != 0
+        let historical_bankruptcy_isolated = allow_empty_asset_historical_bankruptcy
+            && asset_is_active_without_local_insurance_risk_view(group, asset_index);
+        if (group.header.bankruptcy_hlock_active != 0 && !historical_bankruptcy_isolated)
             || group.header.threshold_stress_active != 0
             || group.header.loss_stale_active != 0
             || group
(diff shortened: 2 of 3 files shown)


aeyakovenko (commented):
Clean-room adjacent blocker on this exact head (`541ea56419081fe63120db45563e7bf3e47938bc`, engine `5d608a8653a4d2535b3ff7d3a011ccc8f4accc23`): the durable historical bankruptcy lock also permanently prevents restoring any moving oracle after `RestartAssetOracle`.

A real LiteSVM/SBF lifecycle probe performs a partial asset-0 bankruptcy, completes all loss/backing cleanup, restarts the asset, and completes the fresh 50/50 buyback. At the failure point the restarted asset is `Active`, asset-local position/loss state and OI are zero, `pnl_pos_tot`, stale/B-stale/negative counts are zero, and recovery/threshold/loss-stale flags are clear; only `bankruptcy_hlock_active` remains true.

- `PushAuthMark` fails with `Custom(8)` because restart deliberately installs a `MANUAL` profile.
- `ConfigureEwmaMark` then fails with `Custom(21)` because `require_asset_active_for_oracle_reconfiguration_view` unconditionally gates on the historical h-lock.
- `ConfigureAuthMark` and `ConfigureHybridOracle` use the same gate, and the engine has no h-lock clear transition.

That leaves the fresh generation permanently frozen at its restart price. The fix should not clear the durable bit. It should use a target-asset historical-bankruptcy exception analogous to this PR's withdrawal exception, while retaining the existing Active/local-empty/global-clean checks and the existing oracle authority. I found no existing issue or PR for this separate restart-oracle path.

aeyakovenko (commented):
A second clean-room adjacent blocker shows why keeping backing withdrawals under the unconditional durable h-lock is unsafe unless post-restart deposits are also rejected.

Against current wrapper `557f15b63c7a5f9e058a2c384ca8728829dc2b63` (which contains this PR), the real LiteSVM/SBF lifecycle completes a public bankruptcy, all old-generation backing/loss cleanup, governed asset-0 restart, and verifies zero fresh OI. The same recorded external backing provider then:

1. deposits one fresh principal atom into domain 0 successfully;
2. verifies the fresh bucket contains exactly that unused atom; and
3. immediately withdraws the same atom while the fresh asset is live.

Step 3 fails with `Custom(21)`. `TopUpBackingBucket` accepts the Active generation without checking `bankruptcy_hlock_active`; `WithdrawBackingBucket` calls `live_domain_withdraw_health_or_shutdown_view(..., true, false)`, so the historical exception is disabled. The provider has no owner-controlled continuation. Its only fallback is a new governance shutdown plus recovery delay, which terminates the fresh market.

This is new-generation user principal, not old bankruptcy backing. The durable bit should remain set, but an Active, locally empty fresh generation needs the same narrowly scoped historical exception for its existing backing authority, or the deposit path must reject before transferring tokens. No market-authority withdrawal capability should be introduced.

aeyakovenko:
Adjacent same-asset blocker found in the Meta real-SBF lifecycle: https://github.com/aeyakovenko/percolator-meta/pull/369 (`db7ae5d2d2976718e7c2581e3c7c719735e455b4`).

This is **not** the fresh post-restart case covered here. A public partial bankruptcy consumes 2 of 29 owner-protected atoms; after bounded cleanup, zero OI, `asset_has_position_or_loss_state == false`, zero remaining trader capital, and 27 protected atoms remain. The owner-signed atomic TWAP/Subledger exit for its 13-atom loss-adjusted claim still fails with `Custom(21)` because the old generation correctly retains nonzero spent counters. Backing remains unconditionally h-locked, and this PR's insurance exception requires zero spent.

The existing governed `ResolveMarket` control releases the claim, so the external pool owner otherwise depends on futarchy terminating the live market. Any extension should preserve the durable lock and require exact local emptiness, but cap the transfer through existing domain/provider-ledger withdrawal capacity. It must not add an admin recipient, reusable signer, or authority path.

aeyakovenko:
Closed during PR135 consolidation as nonqualifying under the public LoF/persistent-DoS acceptance rule. Evidence: `v16_program_asset0_recovery_matrix_preserves_provider_withdraw_and_restart_progress`. The public Recovery lifecycle settles and withdraws provider principal and remaining domain insurance, restarts once into a fresh market generation, accepts and exactly withdraws fresh domain insurance with zero inherited spend, then completes a fresh bilateral trade and returns both users' capital. The executable disposition remains in PR #135; no separate production fix is required.

PR #379, opened 2026-07-29
[REAL DoS] Preserve source capacity in combined engine pin (restacked)

aeyakovenko:
## Impact

The prior combined pin could silently omit the engine's reserved-source-capacity repair. A publicly admitted position could then consume every source slot and leave normal crank/trade/recovery continuations unable to allocate the source domain they require.

## Fix

Pin the wrapper to engine commit `5d608a8653a4d2535b3ff7d3a011ccc8f4accc23`, which contains the source-capacity repair, terminal source-payout monotonicity, and empty-asset reuse fix. This PR changes dependency metadata only and adds no authority or fund-moving surface.

This is the current-stack replacement for merged PR #378, whose former base was rewritten during dependency restacking.

## Dependencies

- aeyakovenko/percolator#142
- aeyakovenko/percolator#145
- aeyakovenko/percolator#123
- #283

## Verification

Built the exact SBF with `cargo build-sbf --tools-version v1.52` and ran the wrapper suite against it:

- 6 unit tests passed
- 567 LiteSVM/CU tests passed
- terminal one-unit source-backing payout passed
- post-restart insurance withdrawal passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "percolator"
 version = "0.1.1"
-source = "git+https://github.com/aeyakovenko/percolator?rev=3db81b0b0484754dffb8a9c10e8d86076e969c8d#3db81b0b0484754dffb8a9c10e8d86076e969c8d"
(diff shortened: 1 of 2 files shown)


aeyakovenko:

PR #380, opened 2026-07-29
[BLOCKER LoF] Preserve funding across prospective EWMA updates

aeyakovenko:
## Summary

Prevent a fee-supported EWMA trade from retroactively erasing funding that accrued before the trade landed.

The wrapper now persists an active funding mark plus the oldest prospective mark boundary inside the existing 512-byte per-asset wrapper allocation. Permissionless crank funding uses the active checkpoint until the engine asset reaches that boundary, then advances the checkpoint. Trades remain live at every reported price; this changes only internal funding/mark accounting.

## Public exploit

Exact-parent RED commit: `03018bf84a01279189fe3a226a9a8123dbe862d2`.

The LiteSVM reproducer uses only public instructions and production SBF, with no account/state injection:

1. Two independent users hold opposite positions while an attacker controls a second balanced pair.
2. An honest EWMA premium exists for 101 elapsed slots.
3. Immediately before the permissionless catch-up crank, the attacker risk-reduces the balanced pair at a valid nonzero reported price. The fill pays the same 406-atom fee and reaches the same final mark and effective index as the control ordering.
4. Before the fix, that trade advances `mark_ewma_last_slot`, so the subsequent crank returns zero funding for the elapsed interval.
5. At full SPL resolution, the attacker-controlled accounts receive 1,020 more collateral atoms and the independent victim receives 1,020 fewer.

Pre-fix endpoint results:

```text
control=(200017957, 99990717, 99990920, 406, 1499952, 1010100, 1020000000000000000)
attack =(200009897, 99999797, 99989900, 406, 1499952, 1010100, 0)
```

The economic delta scales with position size and elapsed funding; it is not a dust-only assertion.

## Fix

- Extend `AssetOracleProfileV16` from 400 to 424 bytes within its unchanged 512-byte reserved wrapper slot.
- Persist `funding_mark_e6`, `funding_mark_pending_e6`, and `funding_mark_pending_slot`.
- Record trade, authenticated-mark, and hybrid-oracle mark changes as prospective transitions when the engine asset is behind.
- Compute historical funding from the active checkpoint and a counterfactual bounded index that a later mark cannot rewrite.
- Advance the checkpoint after each bounded asset accrual, including the auto-crank continuation.
- Preserve the new profile-only fields in the host `read_market`/`write_market` round trip.

`TradeCpi` delegates to the same single-trade execution path, and `BatchTradeCpi` delegates to the same batch execution path, so the production update logic is shared across all four trade entrypoints. The public exploit directly covers `TradeNoCpi` and `BatchTradeNoCpi`; existing real-matcher tests cover both CPI dispatch paths.

## Coverage

- Endpoint trade-before-crank versus crank-before-trade, single and batch.
- Mid-segment prospective boundary with the exact same public calls in opposite order, single and batch.
- Independent victim SPL payout and attacker-controlled aggregate SPL payout equality.
- Same fee, final mark, final effective index, and funding accumulator.
- Strict vault/equity conservation after bounded account settlement.
- Host serializer checkpoint persistence.

## Compatibility

The account allocation and dynamic market-slot size do not change; the new fields consume 24 previously reserved zero bytes. For a legacy profile already in the vulnerable state (`mark_ewma_last_slot > asset.slot_last`) at upgrade time, the prior mark is not recoverable from persisted state. Such a profile retains the old zero-funding behavior only until bounded cranks catch the engine up, after which checkpoints initialize normally.

Engine pin: `143e68c4917ed0400a27b952f036a5677047cd84` (unchanged).

## Verification

- `cargo test --test v16_kani`: harness target compiles
- `cargo check --all-targets`
- Max-shape CU: 14-leg refresh 657,537; 14-leg liquidation 1,127,276 (below 1.4M)
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -2322,7 +2322,16 @@ pub mod state {
         check_header(data, KIND_MARKET)?;
         write_wrapper_config_to_bytes(data, config)?;
         if config.oracle_mode != ORACLE_MODE_MANUAL {
-            let base_profile = asset_oracle_profile_from_config(config);
+            let existing_profile = read_asset_oracle_profile(data, 0)?;
+            let mut base_profile = asset_oracle_profile_from_config(config);
+            base_profile.insurance_authority = existing_profile.insurance_authority;
+            base_profile.insurance_operator = existing_profile.insurance_operator;
+            base_profile.backing_bucket_authority = existing_profile.backing_bucket_authority;
+            base_profile.oracle_authority = existing_profile.oracle_authority;
+            base_profile.asset_admin = existing_profile.asset_admin;
+            base_profile.funding_mark_e6 = existing_profile.funding_mark_e6;
+            base_profile.funding_mark_pending_e6 = existing_profile.funding_mark_pending_e6;
+            base_profile.funding_mark_pending_slot = existing_profile.funding_mark_pending_slot;
             write_asset_oracle_profile(data, 0, &base_profile)?;
         }
         write_market_wire(data, group)?;
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -47,7 +47,7 @@ pub mod constants {
 
     pub const HEADER_LEN: usize = 16;
     pub const WRAPPER_CONFIG_LEN: usize = 448;
-    pub const ASSET_ORACLE_PROFILE_LEN: usize = 400;
+    pub const ASSET_ORACLE_PROFILE_LEN: usize = 424;
     pub const ASSET_ORACLE_WRAPPER_LEN: usize = 512;
     pub const MARKET_GROUP_LEN: usize = size_of::<MarketGroupV16HeaderAccount>();
     pub const MARKET_ASSET_SLOT_LEN: usize = size_of::<Market<[u8; ASSET_ORACLE_WRAPPER_LEN]>>();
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Request changes: the two-checkpoint fix still loses funding when the first honest crank spans a pending mark boundary.

I extended the PR's public LiteSVM test with one fresh case: publish the honest EWMA discount at slot 1000, let the attacker-owned short plus a separate balanced stamping pair risk-reduce at slot 1050 (`exec_price=995_000`), omit the slot-1050 crank, and let the first correct permissionless crank catch up at slot 1101. The control cranks at 1050 before the same trade. There is no state injection; both `TradeNoCpi` and `BatchTradeNoCpi` reproduce it and full SPL resolution completes.

Fix-head `9a9bc2f` results, tuple = `(independent_long, controlled_stamper_long, controlled_stamper_short, controlled_primary_short, movement_fee, final_mark, final_index, funding)`:

```text
Single control=(99990600, 99995299, 100004303, 100009400, 398, 990102, 990102, -498000000000000000)
Single attack =(99990102, 99999701,  99999701, 100009898, 598, 990102, 990102, 0)
Batch  control=(99990600, 99995299, 100004303, 100009400, 398, 990102, 990102, -498000000000000000)
Batch  attack =(99990102, 99999701,  99999701, 100009898, 598, 990102, 990102, 0)
```

The independent user loses 498 atoms; attacker-controlled portfolios gain 298 net; the extra 200-atom movement fee receives the remainder. Final mark and effective index are identical. This passes the strict public LoF gate.

Root cause: `permissionless_funding_rate_e9_view` applies `funding_mark_e6` to the entire bounded segment even when `funding_mark_pending_slot` lies inside it, then activates the pending mark only after the segment commits. Crank cadence therefore changes settled funding. The accrual segment needs to stop at the oldest pending boundary (or use an equivalent exact premium integral), activate it, and continue through bounded public cranks. Please add the delayed-midpoint single and batch cases before this is pinned.

aeyakovenko:
Meta current-pin clean-room confirmation: https://github.com/aeyakovenko/percolator-meta/pull/387 at commit `54a3579b92235174d50f10f8f89699de6c4b9b85`.

The real pinned SBF reproduces the endpoint case from fresh system-allocated accounts and completes all four `CloseResolved` payouts to SPL balances. Both orderings collect 406 atoms and end at mark 1,499,952 / effective price 1,010,100. Catch-up-before-trade records `f_short_num=1,020,000,000,000,000,000`; trade-before-catch-up records zero. The attacker-controlled aggregate gains exactly 1,020 atoms and the independent victim loses exactly 1,020.

This is distinct from the funding-cadence flooring report: the elapsed interval is not fragmented; only the two public transactions are reordered.

aeyakovenko:

PR #393, opened 2026-08-01
[REAL LoF] Order matcher config mutations

### LINKED ISSUE #392: [REAL LoF] Presigned matcher retry restores revoked LP risk
## Classification

REAL LoF through public instructions on a normally initialized live market. No market admin, oracle authority, insurance authority, backing authority, or LP signature is available to the attacker after revocation.

## Root cause

`SetMatcherConfig` (tag 68) carries only `enabled`; it has no expected current matcher mutation sequence. Relayers can hold multiple fee variants signed under one valid blockhash. If variant A enables a matcher and the LP later disables it, withheld variant B sees the same signer/account tuple and restores the revoked matcher without a fresh LP signature.

`TradeCpi` intentionally does not require the LP signer once a matcher tuple is enabled. A public taker can therefore reopen risk against the LP immediately before an already-known independent oracle update, then flatten through the restored matcher after the update.

This is a same-portfolio-incarnation ABA bug. PR #304 covers the separate close/reinitialize incarnation boundary and does not order repeated mutations within one live incarnation.

## Public LiteSVM RED

The Meta clean-room regression uses real pinned Percolator and controller SBFs, System allocation plus public `InitMarket`/`InitPortfolio`, real SPL collateral movement, an authenticated external matcher fixture, constrained market-controller handoff, an independent oracle, and permissionless cranks.

1. LP signs two fee variants of matcher enable at the same blockhash.
2. Variant A lands; LP revokes the matcher.
3. A control `TradeCpi` rejects while disabled.
4. Withheld variant B restores the matcher.
5. The attacker opens long without the LP signer before an already-known honest upward mark.
6. The independent oracle publishes; permissionless cranks crystallize LP loss.
7. The attacker flattens through the restored matcher, converts released PnL, and withdraws SPL collateral.

Exact vulnerable result:

```text
attacker deposit:       1,000,000 atoms
attacker withdrawal:    1,000,250 atoms
LP crystallized loss:         250 atoms
```

Pinned vulnerable SBF SHA-256: `74ce6b85576fcbcb7596fdc7a774ff92661853d50ca9eb9475fcba7b1b7f40a4`.

Current upstream `main` at `36fa587e1424a8c7fdde2c701c10938188a51876` has the same unsequenced handler.

## Fix

- Reuse the existing 104-byte portfolio matcher tail; do not move any portfolio offsets.
- Encode a monotonic 63-bit mutation sequence in bits 1..=63 of the existing final word and retain bit 0 as `enabled`.
- Decode legacy tail words 0/1 as sequence 0, preserving deployed account compatibility.
- Add `expected_sequence: u64` to tag 68; reject the old unbound payload.
- Require exact equality and increment on every successful enable or disable.
- Treat an absent legacy matcher tail as sequence 0 before the existing realloc migration.

No signer, authority, PDA, admin, custody, token-transfer, or withdrawal surface is added.

## Fixed proof

- Same public Meta exploit: GREEN; stale retry rejected and LP account byte-identical.
- Focused upstream matcher-config LiteSVM compatibility batch: 10/10 GREEN.
- Fresh post-revoke enable at the current sequence succeeds, proving liveness.
- Final fixed SBF SHA-256: `bfdbd77c26aa5aefa978a5f31c9a6d99ef607431f4252d3ef0ce0e55d2d962bc`.

aeyakovenko:
## Classification

REAL LoF through public `SetMatcherConfig`, `TradeCpi`, oracle, crank, conversion, and withdrawal paths. Fixes #392.

This PR is limited to same-portfolio-incarnation mutation ordering. PR #304 covers the separate close/reinitialize incarnation boundary.

## Public RED

A clean-room Meta LiteSVM test against pinned Percolator SBF SHA-256 `74ce6b85576fcbcb7596fdc7a774ff92661853d50ca9eb9475fcba7b1b7f40a4` proves:

1. an LP signs two fee variants of matcher enable at one blockhash;
2. variant A lands and the LP revokes the matcher;
3. the withheld variant restores it without a fresh LP signature;
4. a public attacker opens and flattens against the LP around an already-known independent oracle update;
5. ordinary cranks crystallize 250 atoms of LP loss;
6. the attacker converts and withdraws 1,000,250 atoms after depositing 1,000,000.

The market is initialized through public instructions, handed to the constrained controller, and uses real SPL collateral movement plus an authenticated external matcher fixture. No initialized Percolator state is injected.

## Fix

- Reuse the existing 104-byte portfolio matcher tail; no account offsets move.
- Pack a monotonic 63-bit sequence into the existing final word while retaining bit 0 as `enabled`.
- Decode deployed legacy words 0/1 and absent legacy tails as sequence 0.
- Add `expected_sequence` to tag 68 and reject old unbound payloads.
- Check exact equality and increment on every successful enable or disable.
- Preserve the sequence in ordinary matcher-config storage helpers.
- Document the new client flow.

This adds no signer, authority, PDA, admin, custody, token-transfer, recipient, or withdrawal surface.

## Verification

- Exact-parent public Meta exploit: RED with attacker withdrawal `1,000,250`, deposit `1,000,000`, LP crystallized loss `250`.
- Fixed public Meta exploit: GREEN; stale retry rejects and LP state is byte-identical.
- Upstream matcher-config LiteSVM compatibility batch: 10/10 GREEN.
- Regression proves fresh enable at the current post-revoke sequence remains live.
- Legacy-tail realloc and failed-context rollback tests remain GREEN.
- `rustfmt --check`: clean.
- `git diff --check`: clean.
- Fixed SBF SHA-256: `bfdbd77c26aa5aefa978a5f31c9a6d99ef607431f4252d3ef0ce0e55d2d962bc`.

Commit: `48ccf9a`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,9 @@ Authority fields are split by scope:
 - **`AssetOracleProfileV16.asset_admin`**: per-asset cold key that rotates that asset's scoped authorities
 - **`AssetOracleProfileV16.insurance_authority` / `insurance_operator` / `backing_bucket_authority` / `oracle_authority`**: per-asset operational authorities
 
-Matcher requests do not use a persisted market nonce. The wrapper invokes the matcher and requires the response to echo the request id, LP identity, asset index, oracle price, and requested size/sign constraints.
+Matcher requests do not use a persisted market nonce. The wrapper invokes the matcher and requires
+the response to echo the request id, LP identity, asset index, oracle price, and requested size/sign
+constraints. Matcher-config mutations use a separate portfolio-local sequence.
 
 ### Vault token account (market collateral)
 - SPL Token account holding collateral for this market
@@ -355,6 +357,12 @@ Unsigned LP matcher fills require an enabled matcher config stored directly on t
 `SetMatcherConfig` (tag 68) is signed by the LP owner and writes:
 
 `matcher_program, matcher_context, matcher_delegate, enabled`
+
+Its payload also carries `expected_sequence`. The caller reads the current portfolio matcher
+sequence, signs that exact value, and each successful enable or disable increments it. A legacy
+portfolio with no matcher tail, or a legacy tail whose enabled word is 0/1, starts at sequence 0.
+This ordering prevents a retained older signed transaction from restoring a matcher after the LP
+revokes or replaces it.
 
 During `TradeCpi` / `BatchTradeCpi`, Percolator reads this LP-account config and requires the
 instruction's matcher program, matcher context, and matcher delegate PDA to match it byte-for-byte.
@@ -448,7 +456,8 @@ This section describes intent and operational ordering, not argument-by-argument
     legs (the matcher's return-data cap).
 - **SetMatcherConfig** (tag 68)
   - LP-owner-signed opt-in/out for unsigned LP matcher fills. This writes the matcher config tail
-    on the LP portfolio: matcher program, matcher context, matcher delegate, and enabled flag.
+    on the LP portfolio: matcher program, matcher context, matcher delegate, enabled flag, and
+    monotonic mutation sequence. The payload must carry the current `expected_sequence`.
 
 ### Oracle / mark management
 - External-oracle markets authenticate configured oracle account(s) in the oracle configuration/crank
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Superseded by consolidated wrapper PR #135 and engine dependency aeyakovenko/percolator#195. PR135's frozen finding-blind benchmark maps this finding to INV-010 and an integrated generic public-route oracle; the production fix and current-pin certification are retained in that submission.

PR #410, opened 2026-09-03
[PRIVILEGED LoF] Escheat abandoned shutdown reserves

aeyakovenko:
## Impact

A market authority could shut down a permissionlessly created asset, wait for the public cleanup delay, then use the shutdown fallback in `WithdrawInsuranceAsset`, `WithdrawBackingBucket`, and `WithdrawBackingBucketEarnings` to send another provider's attributed reserves to an admin-owned SPL account.

The current-main public LiteSVM trace deposits 300 insurance atoms and 500 backing atoms from a distinct provider. Before this fix, the normal shutdown path transfers all 800 atoms to marketauth. No program-owned account bytes are modified out of band.

## Fix

Treat marketauth as a shutdown-progress submitter, not a payout authority. Provider-signed cleanup still pays the provider. Marketauth-only cleanup debits the abandoned source bucket and recredits the same amount to asset-0 insurance without moving SPL tokens. Provider ledger identity remains bound to the provider.

## TDD evidence

- Pre-fix: `v16_attack_shutdown_cleanup_cannot_pay_foreign_reserves_to_marketauth` fails with admin balance `800` instead of `0`.
- Post-fix: the same test passes, including provider reclaim-first controls and exact vault/base-insurance reconciliation.
- The existing end-to-end shutdown/force-close/retire/reuse test now requires vault-neutral escheat instead of trusting the admin to redeposit the funds.

## Verification

- `cargo test --lib`: 8/8
- INV-005 module: 43/43
- INV-064 module: 10/10
- Targeted backing-earnings, shutdown lifecycle, and three SPL rollback tests: green
- Full `v16_cu`: 957/962. The five failures reproduce stale merged-main invariant metadata/expectations unrelated to these handlers: RestartAssetOracle replay roster, three missing caller-input fields / field count, shifted Kani inventory lines, and duplicate Recovery-hint semantics.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10508,19 +10508,7 @@ pub mod processor {
         }
 
         let domain_usize = domain as usize;
-        let (bump, amount_u64) = verify_domain_withdrawal_preflight(
-            program_id,
-            market_ai,
-            authority,
-            dest_token,
-            vault_token,
-            vault_authority_ai,
-            domain_usize,
-            amount,
-            expected_market_id,
-            false,
-            DOMAIN_WITHDRAW_AUTH_BACKING,
-        )?;
+        let escheat_to_base_insurance;
 
         {
             let mut market_data = market_ai.try_borrow_mut_data()?;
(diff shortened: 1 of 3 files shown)


aeyakovenko:

PR #411, opened 2026-09-03
[PRIVILEGED LoF] Bind retained single-CPI taker fee consent

aeyakovenko:
## Impact
A retained `TradeCpi` signed by the taker with `fee_bps = 0` could land after the asset-0 insurance authority raised `trade_fee_base_bps`, provided the LP had independently allowed the new fee. On the pre-fix SBF, a one-unit public trade debited the victim and LP by 50,000 quote atoms each, credited 100,000 atoms to insurance, and the privileged fee authority withdrew all 100,000 through `WithdrawInsuranceAsset`. No program-owned state was injected.

This is a privileged LoF: a compromised or malicious fee authority can extract value from an already-signed taker transaction beyond its retained base-fee consent.

## Fix
Reject single-CPI execution when the current base fee exceeds the taker-signed `fee_bps`. The LP matcher cap and protocol maximum remain independent. A newly signed transaction at the current fee still executes.

## TDD evidence
- Parent `c74209f3`: the production-SBF LiteSVM regression fails because the stale zero-cap trade lands.
- Head: stale trade rejects with exact rollback; a fresh 500 bps trade succeeds and the authorized fee is withdrawable.

## Verification
- exact INV-036 regression: pass
- deterministic PR313 CPI fee-consent route test: pass
- PR313 stateful fuzz, 4 cases: pass
- `cargo test --lib`: 8/8 pass

A broader INV-036 relink was attempted but the host filesystem ran out of space during linking; it did not produce a test failure.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9242,8 +9242,9 @@ pub mod processor {
         if stale_matured {
             return Err(PercolatorError::OracleStale.into());
         }
-        let fee_floor_pre = core::cmp::max(fee_bps, cfg_pre.trade_fee_base_bps);
-        if fee_floor_pre > max_trading_fee_bps {
+        // The taker signs `fee_bps` before the matcher runs. A later policy update cannot raise
+        // the base fee above that retained consent; the unsigned LP is bounded separately below.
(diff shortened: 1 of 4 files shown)


aeyakovenko:

PR #412, opened 2026-09-03
[BLOCKER LoF] Bind matcher grants to position episodes

aeyakovenko:
## Classification

`[BLOCKER LoF]`: a public transaction trace lets an unprivileged relayer revive matcher authority that the LP's later position mutation automatically revoked. The revived grant creates fresh unsigned LP exposure; an honest authenticated mark then transfers and permits withdrawal of exactly 10,000 quote atoms from the LP to the attacker.

No program-owned bytes are injected or rewritten out of band. No market, asset, oracle, or insurance authority is compromised.

## Public exploit trace

1. LP signs and retains `SetMatcherConfig` while its matcher is synchronized.
2. LP and attacker owner-sign `TradeNoCpi` open and close transactions. Both portfolios return flat; the wrapper advances the LP position epoch and automatically disables its matcher grant.
3. The retained config still has the current matcher sequence, so the parent handler accepts it and copies the newer position epoch into stale consent.
4. The attacker invokes `TradeCpi` without a fresh LP signature and creates LP exposure.
5. An authenticated mark moves from 100 to 90.
6. Both accounts crank, take a bounded public exit, convert PnL, and withdraw real SPL funds.
7. The trace classifies `victim_loss_atoms = 10_000` and `unauthorized_gain_atoms = 10_000`.

This contradicts the prior INV-012 closure: explicit enable/disable supersession was covered, but automatic position-mutation revocation and the independent matcher-control sequence were not composed in one landing-order history.

## Fix

Tag 68 now signs and checks `position_epoch` in addition to `portfolio_id` and `expected_sequence`. A retained grant therefore cannot cross any automatic revocation that advances the LP position episode. Matcher-control sequence independence is preserved, so unrelated portfolio request lanes are not serialized.

The prior 28-byte tag-68 schema rejects rather than being reinterpreted. The production decoder body was extracted once and is shared by deployed decode and the tractable Kani proof adapter.

## Red/green commits

- Parent: `d5e2ec6fa727a1049a62851c3be19c638815b4e2`
- RED public exploit: `aae89a60`
- Fix and certification: `da1f45e5`

## Verification

Exact rebuilt default-feature SBF: 1,259,208 bytes, SHA-256 `7b8f3035f880876ab7388dfcdbef73bef5dc7ca6c43dd2013f10574ce62b854a`.

- Exact-parent RED: fails with `victim_loss=10000, attacker_gain=10000`.
- Stateful INV-012: 2/2 pass.
- INV-010 matcher control/trade landing permutations: 6/6 pass.
- CU/public INV-012: 17/17 pass.
- Decoder mutation/canonicalization INV-022: 6/6 pass.
- Tag-68 enabled/disabled CU bound: pass.
- Full-width production decoder Kani: 0/817 failures in 11.2s.
- `cargo fmt --all` and `git diff --check`: pass.


aeyakovenko:
Closing as fixed/covered on main. The retained matcher-grant position-episode fix is in commit bb13083e18d20f34195e40bd7411e32cedf373d2 (`fix(invariants): bind matcher grants to position episodes`), with later invariant coverage on main including a657d195b3bd982e5800daddbfe235f22acc19a6.

PR #413, opened 2026-09-03
[BLOCKER LoF] Crystallize flat fees before first open

aeyakovenko:
## Impact

A flat portfolio could age a maintenance-fee liability while retaining its gross capital, then use that gross capital to pass initial margin on its first open. Once the deferred liability became current, public liquidation could consume pre-existing insurance.

The public LiteSVM trace uses production-like 5% IM/MM and authenticated AuthMark updates. Two attacker-controlled accounts deposit 2,200,000 atoms, the stale flat account opens 20,000,000 atoms of notional, and the coalition later withdraws 2,657,999 atoms while independent insurance falls from 1,000,000 to 542,000: a 457,999-atom extraction. No program-owned bytes are injected.

## Fix

Use the existing value-debit maintenance collector before all single and batch trade routes, including empty-to-nonempty first opens. Fee-net undercollateralized opens reject atomically; after adding the exact missing capital, the same first open lands within the CU limit and credits the debt to insurance. The obsolete flat-account exception is removed, reducing production source by 19 lines.

## TDD evidence

- Exact vulnerable base: `d5e2ec6fa727a1049a62851c3be19c638815b4e2`
- RED commit: `349e1801e0536d9105e6774c9197ef3a85401deb`
- RED SBF SHA-256: `230b6db1278dbff258c84f9a3df78c7d9decc6f8653fe06c46fa5ea9834afb20`
- Fixed commit: `1c52f510`
- Fixed SBF SHA-256: `0032c667d904341a4e648303049ef9c3adeb6f0a0be3937e0a6b16608e3e7099`
- Regression: `v16_attack_flat_fee_debt_cannot_back_first_open_or_spend_independent_insurance`
- Maximum 28-source maintenance sync: 569,464 CU
- Dependency-affected PR223 public and stateful tests pass against the fixed SBF; the adapted public test also passes against the exact parent SBF.

The full stateful inventory is still running. One observed failure is the pre-existing matcher-position-episode counterexample tracked by #412; it reproduces unchanged on the exact parent SBF.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -6802,25 +6802,6 @@ pub mod processor {
         collect_maintenance_fee_to_slot_before_value_debit_view(cfg, group, portfolio, now_slot)
     }
 
-    fn collect_maintenance_fee_before_trade_view(
-        cfg: &WrapperConfigV16,
-        group: &mut state::MarketViewMutV16<'_>,
-        portfolio: &mut percolator::PortfolioV16ViewMut<'_>,
-    ) -> Result<u128, ProgramError> {
-        let active_bitmap = portfolio
-            .header
-            .active_bitmap
-            .map(percolator::V16PodU64::get);
-        if percolator::active_bitmap_is_empty(active_bitmap) {
-            // Opening a first leg does not debit an existing exposure. Leave flat-account fee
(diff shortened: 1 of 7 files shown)


aeyakovenko:

PR #414, opened 2026-09-03
[PRIVILEGED LoF] Bind matcher grants to asset generations

aeyakovenko:
## Classification

Privileged LoF. A malicious or compromised asset administrator can replace an empty asset generation and reuse an independent LP matcher grant that was authorized only before that replacement. No dishonest oracle signature or program-owned byte injection is used.

## Exact RED

- Exact main base: `d5e2ec6fa727a1049a62851c3be19c638815b4e2`
- RED commit: `9000e192`
- Public LiteSVM trace: two funded portfolios, LP authorizes an authenticated matcher, admin retires and reactivates asset 1, attacker opens through the stale unsigned LP grant, authenticated mark moves from 100 to 105, attacker closes and converts PnL.
- Parent result: attacker capital `1,000,000 -> 1,250,000`; independent LP capital `1,000,000 -> 750,000`.

## Fix

`SetMatcherConfig` snapshots the engine monotonic `next_market_id` frontier in a zero-copy portfolio tail. `TradeCpi` and `BatchTradeCpi` accept only nonzero asset generations below that frontier before invoking the matcher. Appended, retired-slot replacement, and Recovery-restart generations therefore require fresh LP consent. Existing generations remain usable after unrelated permissionless appends, and the stable matcher PDA/external ABI is unchanged. Legacy grants without the new tail fail closed.

## Verification

- INV-012 focused LiteSVM suite: 20 tests green after correcting the restart harness setup; stale single, mixed old/new batch, exact rollback, fresh reauthorization, restart, and unrelated-append liveness covered.
- Stateful INV-002 replay fuzz and full asset-operation matrix green; no discoveries.
- Full-width Kani frontier predicate: 0/8 failed, 4/4 covers satisfied.
- 14-leg BatchTradeCpi: 876,223 CU.
- 10 MiB market plus matcher tail: 881,877 CU.
- INV-019 identity census, INV-055 fresh-generation admission, INV-087 persisted-field roster, and 9/9 library tests green.
- All test targets compile when including the pre-existing `examples/dump_layout.rs` stale-field typo correction; that unrelated typo is intentionally excluded from this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,9 @@ Authority fields are split by scope:
 - **`AssetOracleProfileV16.asset_admin`**: per-asset cold key that rotates that asset's scoped authorities
 - **`AssetOracleProfileV16.insurance_authority` / `insurance_operator` / `backing_bucket_authority` / `oracle_authority`**: per-asset operational authorities
 
-Matcher requests do not use a persisted market nonce. The wrapper invokes the matcher and requires the response to echo the request id, LP identity, asset index, oracle price, and requested size/sign constraints.
+Matcher requests do not encode a persisted market nonce. The LP's wrapper-owned capability record
+snapshots the engine's monotonic `next_market_id` frontier, while the wrapper requires each response
+to echo the request id, LP identity, asset index, oracle price, and requested size/sign constraints.
 
 ### Vault token account (market collateral)
 - SPL Token account holding collateral for this market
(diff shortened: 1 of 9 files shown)


aeyakovenko:
Closing as fixed/covered on main. The matcher-grant asset-generation/frontier fix is in commit 8f62a5c59a096c8ce678fa8684efe22fcbc1625e (`fix(invariants): bind matcher grants to asset frontier`), with the row 414 invariant evidence recorded on main.

PR #415, opened 2026-09-03
[BLOCKER LoF] Bind retained insurance withdrawals to stock sequence

aeyakovenko:
## Finding

A fee-bumped, signature-distinct retry of one retained `WithdrawInsuranceAsset` payload could execute again after the insurance fund was replenished. The public LiteSVM trace uses independent provider and operator roles: the provider deposits 100,000 atoms, the operator withdrawal lands once, the provider replenishes 100,000 atoms, and an unprivileged relayer lands the stale transaction variant. On the exact parent, the stale variant transfers the newly supplied 100,000 atoms to the operator. No program-owned bytes are injected or edited.

## Fix

- Add `intent_id` to the withdrawal wire schema.
- Serialize insurance deposits and withdrawals through the existing per-asset `insurance_top_up` watermark; no persisted layout or engine change is required.
- Require exact-next (`current + 1`) stock sequencing. This prevents either independent authority from poisoning the shared lane with an arbitrary jump such as `u64::MAX`.
- Check freshness before accounting and consume the watermark only after validation. SVM rollback protects the pre-CPI write if token transfer fails.
- Add the route to the generalized retained-intent matrix and schema/caller-input rosters.

## TDD evidence

- Exact parent `d5e2ec6f`: RED with `stale retained withdrawal consumed 100000 atoms of replenished provider principal`.
- Head: the same public-SBF trace rejects the stale variant with exact market, vault, destination, and supply rollback; a fresh intent remains live.
- Follow-up parent `5acda781`: a hostile operator withdrawal with `intent_id = u64::MAX` lands at 31,814 CU; head rejects it exactly, then accepts an exact-next operator withdrawal and the independent provider's next top-up.
- The 12-family retained retry matrix passes 8 generated cases with no independent discoveries.
- Host decoder: 5/5; deployed decoder: 6/6.
- Insurance-withdrawal policy/auth/isolation/rollback filters: 19/19.
- Kani decoder proof: 0/1659 failed.
- Full-width sequence proof: 0/106 failed, all 4 covers satisfied.
- Maximum shape: 10 MiB market, 11,564 domains, terminal ledger withdrawal = 36,536 CU.
- `cargo test --tests --no-run`: passes. Engine pin remains `495a5590c97055bd71c6f94d849ff0298f243145`.

## Baseline contradictions

Current `main` already has an unrelated failing INV-008 source roster: `RestartAssetOracle` exists in `SupersededIntentKind::ALL` but its TSV row says `incarnation-episode`. `cargo test --all-targets --no-run` also reaches the pre-existing stale `examples/dump_layout.rs` field name. This PR does not mix either cleanup into the security fix.


aeyakovenko:

PR #416, opened 2026-09-03
[PRIVILEGED LoF] Prevent cold-admin takeover of funded oracle

aeyakovenko:
## Impact

A compromised per-asset cold admin could replace a distinct configured AuthMark oracle after independently funded users opened live exposure. The replacement then published a false but movement-cap-compliant mark and resolved the market, transferring 1,000,000 quote atoms from the victim to the attacker. The configured oracle key never signed the takeover or attack.

This is a privileged non-oracle-admin LoF under scripts/loop.md, not an initialization footgun or a dishonest-oracle assumption.

## Public-interface TDD

The RED commit \`e38b6155\` runs only public instructions against exact parent \`d5e2ec6f\` in LiteSVM with the real SBF:

1. Configure an honest AuthMark oracle and cold asset admin while empty.
2. Two independently funded users open equal matched exposure at 1,000,000.
3. The cold admin replaces the incumbent oracle without its signature.
4. The replacement publishes 1,100,000, exactly within the configured 1,000 bps per-slot cap.
5. Public cranks and permissionless stale resolution commit that mark.
6. Resolved close pays the attacker 101,000,000 and the victim 99,000,000 from equal 100,000,000 deposits.

No program-owned bytes are injected or mutated by the test.

## Fix

Treat live asset position/loss state as funded value for the oracle role. Empty-asset admin setup remains allowed. Once user entitlement depends on the oracle, unilateral cold-admin takeover rejects; the incumbent oracle can still self-rotate with the incoming key co-signing.

## Verification

- Exact parent RED: attacker 101,000,000; victim 99,000,000
- Fixed head GREEN: exploit rejects with exact market rollback
- Incumbent funded-oracle self-rotation: GREEN
- Empty-asset admin rotation: GREEN
- Full INV-005 authority suite: 43/43 GREEN against fixed SBF
- Host unit tests: 8/8 GREEN
- \`cargo build-sbf --tools-version v1.52\`
- Fixed SBF SHA-256: \`b6b8672a30c6ff54d051827f32428b406958fec74910787843314cdb52e7b7b8\`
- Engine pin unchanged: \`495a5590c97055bd71c6f94d849ff0298f243145\`

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -11878,7 +11878,11 @@ pub mod processor {
                 backing_has_value(&slot.engine.backing_long)
                     || backing_has_value(&slot.engine.backing_short)
             }
-            ASSET_AUTH_ADMIN | ASSET_AUTH_ORACLE => false,
+            // The oracle owns no token stock directly, but once this asset has exposure it controls
+            // the mark used to settle that stock. Treat that indirect entitlement as funded so a
+            // cold asset admin cannot replace the incumbent oracle without its consent.
+            ASSET_AUTH_ORACLE => asset_local_has_position_or_loss_state_view(group, asset_index),
+            ASSET_AUTH_ADMIN => false,
             _ => return Err(PercolatorError::InvalidInstruction.into()),
         })
     }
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Superseded by #441. The replacement branch is based on current main and ports only the live-exposure oracle-funded authority fix plus the public LiteSVM regression; keeping this stale branch open risks merging older invariant/test state.

aeyakovenko:
Closing as superseded by #441.

PR #418, opened 2026-09-03
[BLOCKER DoS] Retire native wSOL terminal insurance

aeyakovenko:
## Classification

**BLOCKER DoS**: a valid, live wSOL-quoted market can reach a fully terminal state with nonzero unbudgeted insurance. The only public retirement instruction then attempts `Burn` on the native mint. SPL Token rejects that operation with `NativeNotSupported`, so `CloseSlab` rolls back forever and the market cannot retire.

## Public-interface reproduction

The regression uses only public instructions and valid Solana/SPL accounts:

1. Initialize a market whose quote mint is canonical native wSOL.
2. Fund two portfolios and create paid EWMA movement, producing protocol insurance.
3. Enter DrainOnly, close exposure, convert and withdraw both users, and close both portfolios.
4. Add a small external surplus to distinguish classified protocol stock from ambient vault value.
5. Resolve the market and call `CloseSlab`.

On exact `main` (`d5e2ec6f`), the final instruction fails with `Custom(7)` / SPL `NativeNotSupported`. No program-owned bytes are injected, and SVM rollback is preserved.

## Fix

For canonical native wSOL only:

- transfer externally added surplus to the authorized admin token destination;
- close the remaining native vault to Solana's canonical incinerator;
- validate native-mint state and the canonical writable incinerator account.

Non-native quote mints retain the existing burn path.

## Verification

- RED against exact-main SBF: focused public regression fails at terminal `CloseSlab`.
- GREEN against fixed SBF: regression passes at 25,247 CU.
- `cargo test --test v16_cu close_slab -- --nocapture`: 15/15 pass.
- Includes wrong-sink and read-only-sink rollback checks.

Invariant coverage: INV-018, INV-021, INV-070, INV-073, INV-077, INV-080.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -11019,7 +11019,14 @@ pub mod processor {
         expect_owner(market_ai, program_id)?;
         verify_token_program(token_program)?;
 
-        let (cfg_pre, retired_unbudgeted_insurance, bump, vault_balance, secondary_close) = {
+        let (
+            cfg_pre,
+            retired_unbudgeted_insurance,
+            bump,
+            vault_balance,
+            primary_vault_is_native,
+            secondary_close,
+        ) = {
             let mut market_data = market_ai.try_borrow_mut_data()?;
             let (mut cfg, mut group) = state::market_view_mut(&mut market_data)?;
             expect_live_authority(&cfg.marketauth, admin_dest.key)?;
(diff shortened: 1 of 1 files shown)


aeyakovenko:
Closing as fixed/covered on main. The native terminal booked-residue / wSOL retirement fix is in commit 236b4f851d03ba76ac220a479aa4ec5f2c5dafe5 (`fix: retire native booked residue without burn`), with later row 418 invariant coverage on main.

PR #421, opened 2026-09-04
[BLOCKER DoS] Make terminal insurance payout permissionless

aeyakovenko:
## Finding

A permissionless asset creator controls that asset's insurance authority. On exact `main`, it can top up a one-atom insurance budget, let every user claim settle and every portfolio close, then withhold its signature forever. The market remains Resolved with only that atom outstanding, so `CloseSlab` cannot release 2,133,896,480 lamports. The market authority cannot withdraw the atom or rotate the funded authority, and advancing to slot 1,000,000 does not create another path.

This is a public-interface persistent DoS under one honest cranker. It uses only valid instructions and account construction; no program-owned bytes are injected.

## Fix

In Resolved mode only, after `materialized_portfolio_count == 0` and `c_tot == 0`, treat the configured insurance authority as the payout identity rather than requiring its signature. Any caller can deliver the exact remaining insurance budget only to a token account owned by that configured authority.

Live withdrawals remain signer-gated. Authority, generation, epoch, domain-budget, vault, token-owner, and optional-ledger checks are unchanged.

## TDD evidence

Exact-parent RED commit: `2d936d9148ff43abd349668a3486993b1f96db61`

- Users are paid `[1000, 1000]`.
- Honest terminal withdrawal rejects `Custom(8)`.
- Admin authority rotation rejects `Custom(8)`.
- `CloseSlab` rejects `Custom(21)` immediately and at slot 1,000,000.
- Signing with the withholding key releases the atom and then permits close.

Fix commit: `ab075b735762ad1c4301e2f8a86f3a3fb56951df`

- Unsigned canonical terminal payout succeeds at 33,834 CU.
- Wrong destination and wrong authority reject with byte-exact rollback.
- Live unsigned withdrawal still rejects `Custom(6)` with exact rollback.
- Authority-signed compatibility path succeeds.
- 10 MiB / 11,564-domain terminal payout succeeds at 34,365 CU.
- INV-017 signer/alias matrix: 40/40.
- INV-027 protected principal: 11/11.
- `cargo fmt --all -- --check` and `git diff --check`: clean.

Engine pin remains `495a5590c97055bd71c6f94d849ff0298f243145`.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -10820,7 +10820,6 @@ pub mod processor {
         let vault_authority_ai = account(accounts, 4)?;
         let token_program = account(accounts, 5)?;
         let ledger_ai = accounts.get(6);
-        expect_signer(operator)?;
         expect_writable(market_ai)?;
         expect_writable(dest_token)?;
         expect_writable(vault_token)?;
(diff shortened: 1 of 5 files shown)


aeyakovenko:

PR #425, opened 2026-09-04
[BLOCKER LoF] Preserve oracle cap carry across interleaved trades

aeyakovenko:
## Impact

A permissionless attacker can interleave public risk-reducing wash trades with the mark-accrual schedule and consume each interval without preserving the canonical fractional price-movement carry. Even with an honest crank submitted after every attacker trade, the old path leaves the effective mark at 100 instead of 101 and transfers 100 quote atoms from an independently owned, unsigned LP to the attacker.

The public LiteSVM trace uses only deposits, CPI/no-CPI trades, authenticated AuthMark updates, permissionless cranks, conversion, and withdrawal. It performs no program-owned state injection and ends in exact SPL balances:

- canonical: victim 1,000,100; attacker 999,900
- exploit ordering: victim 1,000,000; attacker 1,000,000
- both paths conserve the aggregate 4,000,000 atoms

## Fix

Use the same canonical accrual-path builder as PermissionlessCrank before a position-changing operation. Commit its target/checkpoint and fractional price-movement carry only when the entire bounded path has zero effective-price movement; price-moving intervals still require the ordinary crank route.

This removes the wrapper-only shortcut that advanced the engine slot while dropping oracle-profile carry.

## Regression

`v16_attack_risk_reducing_trade_cannot_erase_canonical_price_movement_remainder` is parent RED / fixed GREEN and compares terminal SPL entitlements, effective price, carry, custody, and CU across the two orderings.

- parent RED reproduced on `d5e2ec6f`
- fixed GREEN on `0fbeb696`
- peak affected trade: 290,503 CU fixed, below the 345,000 cap
- focused funding/order, route, boundary, library, and callsite-roster checks pass

Invariant gap: INV-024/041/045/052/071/086 history composition. Per-call conservation held, but operation partitioning could discard a canonical arithmetic carry and redistribute terminal entitlement.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -15468,11 +15468,11 @@ pub mod processor {
         Ok(target)
     }
 
-    fn zero_move_funding_segment_for_profile_view(
-        profile: &state::AssetOracleProfileV16,
+    fn zero_move_funding_path_for_profile_view(
+        profile: &mut state::AssetOracleProfileV16,
         group: &mut state::MarketViewMutV16<'_>,
         asset_index: usize,
-    ) -> Result<Option<(u64, i128)>, V16Error> {
+    ) -> Result<Option<(u64, u64, Vec<AccrualStepV16>)>, V16Error> {
         if !oracle_v16::profile_is_price_managed(profile)
             || asset_index >= group.header.config.max_market_slots.get() as usize
             || asset_index >= group.markets.len()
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as fixed/covered on main. The canonical accrual carry fix is in commit 2669bf1ba26f64e5b80c5a52a0bea4e55f3af3b5 (`fix: preserve canonical accrual carry across interleaved trades`), with INV-045 coverage on main.

PR #426, opened 2026-09-04
[BLOCKER LoF] Require current Hybrid evidence before health refresh

aeyakovenko:
## Finding

`PermissionlessCrank` could refresh a stale victim certificate and then liquidate from a cached Hybrid mark without carrying any external observation. Replaying the already-consumed prior-slot report had the same result. A newer authenticated rescue report could already be available, but caller-controlled omission selected the adverse cached state.

Exact-main public LiteSVM trace (no program-owned byte mutation):

- victim and independent maker open 50 units at 1,000,000;
- an authenticated Pyth report moves the effective price to 997,600 in slot 2;
- a fresh authenticated 1,000,000 rescue report exists in slot 3;
- two no-tail cranks, or two cranks replaying the slot-2 report, refresh then liquidate the victim;
- victim position falls from 50,000,000 to 49,716,479 and insurance receives 142 atoms;
- after authenticated catch-up and owner-signed close, the victim withdraws 2,599,177 versus 2,600,000 in the fresh-report control: exact terminal loss 823 atoms;
- harmful route max CU is 274,660.

The bounded selector needs more than one transaction for refresh then liquidation. Without a per-step freshness gate, the first stale call consumes the account action while omitting the external state that would make the victim healthy.

## Fix

Before any stale/liquidatable full-account refresh, require every active normal-hours Hybrid profile to have advanced from an authenticated report in the current slot. Prior-slot replay and omission reject with exact market/portfolio/vault rollback. Once `hybrid_soft_stale_slots` matures, canonical EWMA after-hours fallback remains available, so unavailable feeds do not permanently block progress.

A newly advancing report is still governed by configured oracle age/confidence policy; this change closes omission and replay across the multi-call account-action sequence.

## TDD evidence

- Test-only commits: `2e4de70c`, `08aedad2`
- Fix: `0b3d5ae0`
- Parent SBF SHA-256: `230b6db1278dbff258c84f9a3df78c7d9decc6f8653fe06c46fa5ea9834afb20` (RED for omission and prior-slot replay)
- Fixed SBF SHA-256: `7cef1f7cc1565fd7ae62a4ae88f823d85653559237228925a6cad46e4ba4c4fa` (GREEN; both reject with exact rollback and preserve 2,600,000)

Focused controls:

- INV-056 terminal RED/GREEN regression passes; fixed max CU 172,110.
- Unavailable-Pyth after-hours terminal fallback passes; max CU 174,210.
- Full 14-leg composite-oracle liquidation passes; max CU 1,248,200 under 1.4M.
- Engine pin remains `495a5590c97055bd71c6f94d849ff0298f243145`.

Invariant gaps: INV-020 authenticated oracle provenance, INV-056 discovery-only hints/full refresh, INV-072 order-robust crankability, and INV-080 exact rollback.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -14006,7 +14006,7 @@ pub mod processor {
             // before either route can certify the account, even when the engine selects another
             // leg for the first bounded step.
             if (summary.stale || summary.liquidatable) && !summary.b_stale {
-                reject_missing_pending_liquidation_observations_view(
+                reject_incomplete_account_health_observations_view(
                     &cfg,
                     &group,
                     &portfolio,
@@ -14301,7 +14301,7 @@ pub mod processor {
         Ok(())
     }
 
-    fn reject_missing_pending_liquidation_observations_view(
+    fn reject_incomplete_account_health_observations_view(
         cfg: &WrapperConfigV16,
         group: &state::MarketViewMutV16<'_>,
         portfolio: &percolator::PortfolioV16ViewMut<'_>,
@@ -14333,7 +14333,7 @@ pub mod processor {
                 }
                 seen_assets[seen_asset_count] = leg.asset_index;
                 seen_asset_count += 1;
-                reject_missing_observation_that_changes_accrual_view(
+                reject_incomplete_asset_health_observation_view(
                     cfg,
                     group,
                     leg.asset_index as usize,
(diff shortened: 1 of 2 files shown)


aeyakovenko:
Closing as fixed/covered on main. The current Hybrid evidence fix is in commit 7d0d2539b0fa7a5c6b86f6ae300bbf47cafb40fc (`fix: require current hybrid evidence for health refresh`), with INV-020 coverage on main.

PR #428, opened 2026-09-08
[BLOCKER LoF] Consume insurance withdrawal epoch on retained debit

aeyakovenko:
## Summary
- fixes retained WithdrawInsuranceAsset replay against replenished insurance stock
- consumes the authorizing authority epoch on successful live/resolved insurance withdrawals
- adds deterministic public-route row 415 regression under INV-008

## Public-route bug
Two retained WithdrawInsuranceAsset transactions can be signed before either lands. The first withdraws available insurance. A later public top-up replenishes stock. Before this fix, the second retained transaction could withdraw the replenished stock with the same old authority epoch, causing the same retained economic authorization to spend twice.

## Validation
- Parent SBF: new row 415 test fails; retained retry succeeds and destination receives a second 1,000-token withdrawal
- Fixed SBF: new row 415 test passes across asset 0/1 and long/short replenishment cases
- Existing INV-064 insurance policy selector: 10 passed
- cargo fmt --all -- --check: passed
- git diff --check: passed

No engine or schema changes.

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -645,7 +645,8 @@ pub mod state {
         /// backing watermark and preserves both old lanes without changing persisted layout.
         pub backing_fee: u64,
         /// Strict per-asset authority incarnation. Every market/asset authority handoff that can
-        /// affect this asset binds the exact value and increments it atomically.
+        /// affect this asset binds the exact value and increments it atomically. Successful
+        /// insurance withdrawals also consume their authorizing epoch to prevent retained replay.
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as fixed/covered on main. The retained insurance debit epoch fix is in commit 47268c9acd27d6daca738e6a8ff61e49bf950b83 (`fix: consume terminal insurance debit epoch`), with later Row 428 coverage on main including 47b75632.

PR #432, opened 2026-09-09
[PRIVILEGED LoF] Enforce retained single-CPI taker fee terms

aeyakovenko:
## Summary

This isolates the row411 retained single-CPI fee-consent fix:

- reject retained single-CPI execution when the current market base fee exceeds the taker's signed fee terms
- keep the existing LP cap check intact
- add an INV-014 LiteSVM public-route regression that proves rejection before matcher CPI/rollback, then successful execution after the policy is back inside both taker and LP caps

## Verification

```text
env CARGO_TARGET_DIR=/dev/shm/row411-fix-verify cargo test -q --test v16_cu inv_014_delayed_policy_and_policy_epoch_safety::retained_single_cpi_taker_fee::v16_retained_single_cpi_preserves_taker_fee_terms_and_lp_cap -- --exact --nocapture
```

Result:

```text
INV-014: 10 worlds, 10 exact rejections, 10 retained exits, 20 SPL withdrawals; peak CU=162147
1 passed
```

Also passed:

```text
cargo fmt --all -- --check
git diff --check
```

### DIFF
--- a/src/v16_program.rs
+++ b/src/v16_program.rs
@@ -9242,6 +9242,10 @@ pub mod processor {
         if stale_matured {
             return Err(PercolatorError::OracleStale.into());
         }
+        // The taker's signed base-fee allowance is independent of the LP's standing cap.
+        if cfg_pre.trade_fee_base_bps > fee_bps {
+            return Err(PercolatorError::InvalidInstruction.into());
+        }
         let fee_floor_pre = core::cmp::max(fee_bps, cfg_pre.trade_fee_base_bps);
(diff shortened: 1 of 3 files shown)


aeyakovenko:
Closing as fixed/covered on main. The retained single-CPI taker base-fee consent fix is in commit 7f319c6be68114f09079ff39c76233cbe8b490fb (`fix: enforce retained single-CPI taker base-fee consent`), with INV-014 coverage on main.

PR #435, opened 2026-09-13
[REAL LoF] Add mixed-role resolved debt attribution regression

aeyakovenko:
## Summary
- pins engine fix aeyakovenko/percolator#201 / 0b861efbdc5b613c300344dc1d73cd5c2622464f
- adds public-route LiteSVM INV-039 regression for mixed creditor/debtor close ordering
- records row419 coverage/reopening evidence without promoting generic invariant closure

## Bug shape
A publicly constructed resolved mixed-role sequence could underpay a later terminal claimant after the original source claim and creditor legs exited. The red run on the old engine pin paid owner 2 310000 where the independent entitlement oracle expected 340000.

## Validation
- rebuilt default-feature SBF
- exact mixed-role selector
- both INV-079 metadata gates
- cargo fmt --all -- --check
- git diff --check / git show --format= --check HEAD
- surrounding INV-039 run passed 20 cases

Depends on aeyakovenko/percolator#201 before merge.


aeyakovenko:
Closing as superseded by #440. #440 is the focused replacement branch: it pins the current engine fix commit and carries the public INV-039/INV-076 LiteSVM regressions without the broad PR135/invariant-coverage diff.

